PR #19, opened 2020-08-10
CMake changes for custom paths and windows

robe2:
Allow for compiling under windows and custom include paths for jsonc, proj, lwgeom includes.  For some reason doesn't find it in my paths without explicit specifying.
The win32 part is to get around the netcdb.h is missing when compiling under windows.

so my cmake call looks like:

```
	cmake  -G "MSYS Makefiles" \
		-DLWGEOM_INCLUDE_DIR:PATH=${LIBLWGEOM_PATH}/include -DPROJ_INCLUDE_DIR:PATH=${PROJ_PATH}/include \
		-DJSONC_INCLUDE_DIR:PATH=${JSON_PATH}/include \
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,13 @@ execute_process(COMMAND ${PGCONFIG} --includedir --includedir-server OUTPUT_VARI
 include_directories(${PostgreSQL_ACTUAL_INCLUDE_DIR})
 include_directories("include")
 
+if(WIN32)
+    include_directories(${PostgreSQL_ACTUAL_INCLUDE_DIR}/port/win32)
+    if(MSVC)
+        include_directories(${PostgreSQL_ACTUAL_INCLUDE_DIR}/port/win32_msvc/)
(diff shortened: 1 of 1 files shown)


cvvergara:
I am intrigued with the following:
```
	else()
		MESSAGE(STATUS "LWGEOM not found - compiling without PostGIS support")
		set(WITH_POSTGIS OFF)
	endif ()
endif ()

include("test/scripts/test.cmake")

if (WITH_POSTGIS)
		include("point/point.cmake")
endif ()
```

does it need postgis or not?
```
set(WITH_POSTGIS OFF)
```

estebanzimanyi:
Good point Vicky. At the beginning we though of provided temporal types with or without PostGIS support but this is no longer the case. I will correct this.

robe2:
This one was incomplete.  I created another pull request against develop that uses built-in cmake findPostgreSQL and some windows patches that are used in pgRouting.  Closing this one out to minimize on confusion.

PR #27, opened 2020-09-14
Useful cmake files from pgrouting

### LINKED ISSUE #26: Some useful cmake files
**Locate the issue if applicable**
As a first step to make the CMakeLists build for more OS, add some cmake files that are know to work on other projects specifically on pgRouting.

cvvergara:
Closes #26 
cmake Files from pgRouting repository
pgRouting uses them on the build process, I just brought the ones that I feel will be useful (some might not be but they can be removed later if they are not used)
Yeah, many commits for 7 files, but they have all the history associated with them.

Next steps is to start using them.


dkastl:
What about squash/rebase the PR to have a cleaner Git log?
It would be enough to say "Adds Cmake files from pgRouting" rather than having lots of commits from the original Git history.

cvvergara:
I don't know I would need to check that the names are not lost on the squah

dkastl:
Well, on the other hand, do you really think that so many people have significantly contributed to these Cmake files? Or have they been part of the commit because other files were part of the commit? I'm not sure either.

cvvergara:
Reduced the number of commits, we lost the history,
So thanks to Steve, Anton, Daniel and some others that participated on this files :-)

dkastl:
... or lookup the file history in the pgRouting project, where everything is still present.

cmake/mdb/Version.cmake
```diff
@@ -0,0 +1,34 @@
...
+else()
+    set(PGROUTING_DOC_LINK "https://docs.pgrouting.org/${PGROUTING_LIB_VERSION}/en")
+endif()
+
+
+if (PGROUTING_DEBUG)
+    message(STATUS "PGROUTING_VERSION: ${PGROUTING_VERSION}")
+    message(STATUS "PGROUTING_VERSION_MAJOR: ${PGROUTING_VERSION_MAJOR}")
+    message(STATUS "PGROUTING_VERSION_MINOR: ${PGROUTING_VERSION_MINOR}")
+    message(STATUS "PGROUTING_VERSION_PATCH: ${PGROUTING_VERSION_PATCH}")
```
estebanzimanyi, line 24: Please change all PGROUTING_ by MOBILITYDB_

estebanzimanyi (changes requested):

estebanzimanyi (approved):

cvvergara:
Because I used perl I didnt format to the current .editor config

PR #29, opened 2020-09-18
Optimize lifting functions

estebanzimanyi:


mschoema (changes requested):
Could you initialize lower_eq and upper_eq in function tfunc_tsequence_base_discont1?

mschoema (approved):

PR #31, opened 2020-09-21
Reduce code duplication

estebanzimanyi:


point/include/tpoint_analytics.h
```diff
@@ -0,0 +1,33 @@
...
+#include "tpoint.h"
+
+/*****************************************************************************/
+
+extern Datum tpoint_to_geo(PG_FUNCTION_ARGS);
+extern Datum geo_to_tpoint(PG_FUNCTION_ARGS);
+
+/*****************************************************************************/
+
+#endif
```
mschoema, line 33: Fix analytics.h to correspond to the code in analytics.c

mschoema (approved):

PR #46, opened 2020-11-18
Specify the name of the License in the file headers

mschoema:


mschoema:
Since the SQL files in test/* are not code files but only automated tests, I am not certain that we should add the license headers there, but I could be wrong.
The documentation files are provided under [another license](https://github.com/MobilityDB/MobilityDB#licenses).

mahmsakr:
https://opensource.guide/legal/
"When you make a creative work (such as writing, graphics, or code), that work is under exclusive copyright by default."

So we need to add copyright and license everywhere.
For test/*, because it is code, it should also be PostgreSQL License.
For doc/*, you are right, the copyright is the same, but the license is creative commons

mschoema:
I added license headers too the sql test files.

For the documentation, the license is specified in the [documentation itself](https://docs.mobilitydb.com/MobilityDB/develop/) (after building it from the source files):

> This is the manual for MobilityDB v1.0. The MobilityDB Manual is licensed under a Creative Commons Attribution-Share Alike 3.0 License 3. Feel free to use this material any way you like, but we ask that you attribute credit to the MobilityDB Project and wherever possible, a link back to MobilityDB.

On the [Creative Commons License webpage](https://creativecommons.org/choose/results-one?license_code=by-sa&amp;jurisdiction=&amp;version=3.0&amp;lang=en), this is the only described way to refer to the license. If you believe we need to add something more to the doc/* files, we could do that, but I don't know what and where we need to add it.

mahmsakr:
What concerns the documentation license, it sounds good to me that the license appears in the compiled document.

mschoema:
Could you approve the changes and merge the pull request if you think this is sufficient?

mahmsakr:
It seems that your editor automatically trims excess spaces, causing the diff to detect fake additions/deletions. See for example the diff of  point/test/queries/64_tpoint_aggfuncs.test.sql  in this PR.
It is not a big deal. I could merge the PR like this. Ok for you ?

mschoema:
These white spaces where still there from before we added the editorconfig file. Since I saved all the files again with the new config, the spaces disappeared. I saw the issue this caused no the tests, so I modified the expected files also in the last commit 'Fix headers in sql test files' and all the tests pass, so I think it's ok.

Do you still see a nonzero diff in the last version?

mahmsakr (approved):

PR #58, opened 2021-02-27
Add getX, getY, getZ and derivative functions wrt Issue # 56

estebanzimanyi:


src/tnumber_mathfuncs.c
```diff
@@ -502,4 +502,101 @@ tnumber_degrees(PG_FUNCTION_ARGS)
...
+  /* Stepwise interpolation */
+  if (! MOBDB_FLAGS_GET_LINEAR(seq->flags))
+  {
+    Datum derivative = Float8GetDatum(0.0);
+    for (int i = 0; i < seq->count; i++)
+    {
+      TInstant *inst = tsequence_inst_n(seq, i);
+      instants[i] = tinstant_make(derivative, inst->t, FLOAT8OID);
+    }
+  }
```
mschoema, line 529: Should we allow derivatives of stepwise functions?
Theoretically, the derivative of a stepwise function is 0 everywhere except at the places of discontinuity, where the derivative is infinite.

estebanzimanyi: This is exactly what it is done, put a 0.0 converted into a Datum in the timespan of the temporal value.

mschoema: Yes, for the majority of the movement this is correct, but at all the intermediate instants themselves, the derivative is supposed to be infinite. I am not sure setting it all to 0 is the best idea.

estebanzimanyi: You are fully right ! Mahmoud also has the same viewpoint. I will simply disallow the derivative and the speed functions for temporal values with linear interpolation ! Many thanks for your insight !

mschoema (approved):

PR #84, opened 2021-07-18
Add action test postgres versions

cvvergara:
Proposed changes:
The action will test on:
- postgres: 10,11,12,13
- ubuntu 20 (latest) & 18
- postgis: 2.5
see: https://github.com/cvvergara/MobilityDB/runs/3098174915?check_suite_focus=true

### DIFF
--- a/.github/workflows/pgversion.yml
+++ b/.github/workflows/pgversion.yml
@@ -50,6 +50,7 @@ jobs:
             liblwgeom-dev \
             libproj-dev \
             libjson-c-dev \
+            libprotobuf-c-dev \
             postgresql-${{ matrix.psql }} \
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
> Proposed changes:
> The action will test on:
> 
> * postgres: 10,11,12,13
> * ubuntu 20 (latest) & 18
> * postgis: 2.5
>   see: https://github.com/cvvergara/MobilityDB/runs/3098174915?check_suite_focus=true

Many thanks Vicky for your continuous support ! In the last PR we have added asMVTGeom support for temporal points in a similar way as PostGIS does for geometries. This requires the library protobuf-c as PostGIS also requires it.

cvvergara:
Yup, I noticed that you were doing changes.
Will update the PR

PR #97, opened 2021-07-22
Sql organization

cvvergara:
This PR organizes the `.sql` files and build processing
```
sql
├── general
├── npoint
└── point
```
Does not use the C to determine what functionality is or is not is to be included, instead that is determined on the build. 
To make it work, some changes on the sql had to be made.

For example, for one word changes:
```
#if MOBDB_PGSQL_VERSION >= 110000
  RESTRICT = periodsel, JOIN = scalarlejoinsel
#else
  RESTRICT = periodsel, JOIN = scalarltjoinsel
#endif
```
To
```
  RESTRICT = periodsel, JOIN = @JOIN_LE@
```
For several line changes, on the endif the condition of the if was added: 
```
#if MOBDB_PGSQL_VERSION < 110000
CREATE FUNCTION period_gist_decompress(internal)
  RETURNS internal
  AS 'MODULE_PATHNAME'
  LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
#endif //MOBDB_PGSQL_VERSION < 110000
```
Please approve, so that I can do my own merge.


cvvergara:
This is not generated anymore (no need)
https://coveralls.io/builds/41554152/source?filename=build%2Fmobilitydb--1.0.sql.c

estebanzimanyi (approved):

estebanzimanyi:
Wonderful work again Vicky ! Thanks !

PR #103, opened 2021-07-25
Cleanup build

cvvergara:
Before working on the src code,
the build needs to be cleaned up and adjusted so it can find postgres in other environments besides unix.


cvvergara:
The postgres version number was not been captured correctly
```
POSTGRESQL_VERSION_NUMBER=120700
-- POSTGRESQL_VERSION_STRING=PostgreSQL 12.7 (Ubuntu 12.7-0ubuntu0.20.04.1)
-- Found POSTGRESQL: /usr/include/postgresql/12/server (found version "12.7") 
Found compatible PostgreSQL 12.7 (Ubuntu 12.7-0ubuntu0.20.04.1)
-- PG_FULL_VERSION=120000
```

cvvergara:
MobilityDb is not using protobuf in its code and it is not building postgis so there is no need to checj for protobuf

cvvergara:
Part of understanding lwgeom
Finidng lwgeom.h using postgis 2.4 (only available on postgres10 vs 2.5)

I get this interesting situation on 2.4:
```
ERROR:  could not load library "/usr/lib/postgresql/10/lib/libMobilityDB.so": /usr/lib/postgresql/10/lib/libMobilityDB.so: undefined symbol: lwgeom_hash
```
As we can see here:
https://github.com/postgis/postgis/search?q=lwgeom_hash
lwgeom_hash is not defined on lwgeom.h
I double checked:
```
grep lwgeom_hash /usr/include/liblwgeom.h 
```
does not give a result. Therefore it is looking for it on /usr/lib/postgresql/12/lib/postgis2-4.so and its not in that version
```
https://github.com/postgis/postgis/blob/stable-2.4/postgis/lwgeom_btree.c
```
Therefore that is the reason it can not be used with postgis 2.4

cvvergara:
Understanding with Postgis 3
The error is different, (note that on first error when creating the extension it stops, there might me more)
```
ERROR:  could not load library "/usr/lib/postgresql/11/lib/libMobilityDB.so": /usr/lib/postgresql/11/lib/libMobilityDB.so: undefined symbol: getSRSbySRID
```
But interesting:
```
git grep getSRSbySRID ..
../point/include/postgis.h:extern char *getSRSbySRID(FunctionCallInfo fcinfo, int32_t srid, bool short_crs);
../point/src/tpoint_out.c:      srs = getSRSbySRID(fcinfo, srid, true);
../point/src/tpoint_out.c:      srs = getSRSbySRID(fcinfo, srid, false);
```
On Postgres 3 that one is defined in: 
https://github.com/postgis/postgis/blob/stable-3.1/libpgcommon/lwgeom_cache.c
on Postgres 2.5 is defined in
https://github.com/postgis/postgis/blob/stable-2.5/postgis/lwgeom_export.c

In any case It is not the topic of this PR, yet. but those comments are a reminder of what to look when trying to make it work with PostGIS 3

cvvergara:
Need to find PostGIS to make sure that postgis-2.5 is found and that it is used

estebanzimanyi (dismissed):

cvvergara:
@estebanzimanyi 
There is no need to review while it is work in progress, I haven't clicked the button of "Ready for review", it still a draft

cvvergara:
@estebanzimanyi 
From what I understand in this code:
https://github.com/MobilityDB/MobilityDB/blob/develop/CMakeLists.txt#L149-L169
When lwgeom is not found, then there is a fatal error and nothing is done.
So its always building "with postgis", never without it.

cvvergara:
There is a FindGSL but its avaliable on cmake 3.2 will see what happens when using that one
https://cmake.org/cmake/help/latest/module/FindGSL.html

estebanzimanyi:
>
> @estebanzimanyi <https://github.com/estebanzimanyi>
> From what I understand in this code:
>
> https://github.com/MobilityDB/MobilityDB/blob/develop/CMakeLists.txt#L149-L169
> When lwgeom is not found, then there is a fatal error and nothing is done.
> So its always building "with postgis", never without it.
>
Indeed MobilityDB is always built with postgis.

BTW these lines come from Regina
https://github.com/MobilityDB/MobilityDB/pull/19/files

cvvergara:
About the lines, that is using the FindPostgresql, I created a custom one
https://github.com/cvvergara/MobilityDB/blob/cleanup-build/cmake/FindPOSTGRESQL.cmake
So now everything is based on the custom one, that gets everything that its needed in mobitlityDB
To avoid confision its FindPOSTGRESQL instead of FindPostgresql

cvvergara:
Basically the description of the changes of this PR are on the titles of the different commits.

cvvergara:
@estebanzimanyi Now you can review :-)

estebanzimanyi (approved):

PR #113, opened 2021-08-02
Macos action

cvvergara:
Creating a macos action that will be manually triggered
Because currently only postgis 2.5 is required and macos installs postgis 3
The idea of having this action is to have it ready for when postgis 3 is used.

Do not merge, as this PR currently has many commits that will be squashed to only one commit that solves this issue when it is ready to be merged. Which will be announced on a comment and it will be ready for review.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,45 @@
+name: Build for macOS
+
+# manually triggered workflow
+# - macOS test takes too much time
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: install deps
+        run: |
+          brew install postgresql postgis proj liblwgeom
+          # installed with postgis just making it splicit:
+          brew install json-c
+
+      - name: Configure
+        run: |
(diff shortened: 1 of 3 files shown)


cvvergara:
liblwgeom is keg-only and headers are not installed in macos.
```
==> Summary
🍺  /usr/local/Cellar/liblwgeom/2.5.4_1: 13 files, 1.4MB
==> No outdated dependents to upgrade!
==> Checking for dependents of upgraded formulae...
==> Reinstalling 1 broken dependent from source:
liblwgeom
Warning: liblwgeom has been deprecated because it liblwgeom headers are not installed anymore, use librttopo instead!
==> Caveats
==> liblwgeom
liblwgeom is keg-only, which means it was not symlinked into /usr/local,
because conflicts with PostGIS, which also installs liblwgeom.dylib.
```

Will start the cleanup because can't go further as there are no headers installed on macos.
until `librttopo` is used mobilityDB wont work on macos

cvvergara:
Done in this PR
- Adjusted CMakeLists.txt for macos
- Using QGIS's FindPROJ which finds also for macos, and adjusting it for MobilityDB requirements
- Adding the macos actions that can only be triggered manually, as it currently fails due to the lack of headers of lwgeom

estebanzimanyi (approved):

PR #129, opened 2021-08-04
Fixing the test.sh file

mbakli:
Three changes are needed in the test/scripts/test.sh file. 
1- To make MobilityDB used by the test files, we need to set the shared_preload_libraries configuration parameter to Postgis-2.5.so.
2- To load the prerequisites tables, we need to get the load.sql.xz file from the source of the project. The change is in the following three files: test/general/CMakeLists.txt, test/point/CMakeLists.txt, test/npoint/CMakeLists.txt.
3- To start Postgres, the parameter run_ctl needs to be replaced with $PGCTL to get the complete command parameters and to print a message in the log file that is used to indicate whether PostgreSQL has started or not.


estebanzimanyi:
We need to revert back many directory changes, e.g., return to the previous version
´´´
add_test(
  NAME load_npoint_tables
  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test
  COMMAND ${PROJECT_SOURCE_DIR}/test/scripts/test.sh run_passfail ${CMAKE_BINARY_DIR} load_npoint_tables "npoint/data/load.sql.xz"
)
´´´
instead of the current one
´´´
add_test(
  NAME load_npoint_tables
  WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/test/scripts
  COMMAND test.sh run_passfail load_npoint_tables "data/load.sql.xz"
)
´´´

Similarly the previous version
´´´
add_test(
  NAME ${TESTNAME}
  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test
  COMMAND ${PROJECT_SOURCE_DIR}/test/scripts/test.sh run_compare ${CMAKE_BINARY_DIR} ${TESTNAME} ${file}
  )
´´´
instead of the new one
´´´
add_test(
  NAME ${TESTNAME}
  WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/test/scripts
  COMMAND test.sh run_compare ${TESTNAME} ${file}
  )
´´´

With these changes and several orders the .o, .gcno, and and .gcda are generated in the same directories and results of the comparison tests are also done.

Nevertheless there is still an issue that now we use cmake instead of gcc for producing SQL definitions depending on the PostgreSQL version. Therefore, now the string @extschema@ is replaced by an empty string in the following kind of definitions
´´´
CREATE FUNCTION contains(geometry, tgeompoint)
  RETURNS boolean
  AS 'SELECT $1 OPERATOR(@extschema@.@>) $2 AND @extschema@._contains($1,$2)'
  LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
´´´
And then the resulting definition is errorneous in SQL. I am trying to find a solution for this problem.

cvvergara:
Will have a look.

cvvergara:
I am looking at commits one by one, since the https://github.com/MobilityDB/MobilityDB/commit/a61c8862c49a21b87373d5f597ec0ac57de96b75
Some are doing more than what is intended. Like this one https://github.com/MobilityDB/MobilityDB/commit/3ee2c52b117158165206b5584748e3914442088a
I remember I mentioned something about it.

And because all PR come from one branch it makes it difficult to see the original commit
Normally I start working
```
git fetch upstream
git checkout upstream/develop
git switch -c branch-name-that-refers-to-what-the-work-will-be
```
if the branch gets outdated, instead of using the buttong on the PR:
```
git fetch upstream
git rebase upstream/develop
```
and everything will be placed on top of the PR

estebanzimanyi:
> We need to revert back many directory changes, e.g., return to the previous version
> 
> ```
> add_test(
>   NAME load_npoint_tables
>   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test
>   COMMAND ${PROJECT_SOURCE_DIR}/test/scripts/test.sh run_passfail ${CMAKE_BINARY_DIR} load_npoint_tables "npoint/data/load.sql.xz"
> )
> ```
> 
> instead of the current one
> 
> ```
> add_test(
>   NAME load_npoint_tables
>   WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/test/scripts
>   COMMAND test.sh run_passfail load_npoint_tables "data/load.sql.xz"
> )
> ```
> 
> Similarly the previous version
> 
> ```
> add_test(
>   NAME ${TESTNAME}
>   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test
>   COMMAND ${PROJECT_SOURCE_DIR}/test/scripts/test.sh run_compare ${CMAKE_BINARY_DIR} ${TESTNAME} ${file}
>   )
> ```
> 
> instead of the new one
> 
> ```
> add_test(
>   NAME ${TESTNAME}
>   WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/test/scripts
>   COMMAND test.sh run_compare ${TESTNAME} ${file}
>   )
> ```
> 
> With these changes and several orders the .o, .gcno, and and .gcda are generated in the same directories and results of the comparison tests are also done.
> 
> Another problem is the name of MobilityDB library that should not contain the minor version. Thus in the main `CMakeLists.txt` it necessary to change
> 
> ```
> set(MOBILITYDB_EXTENSION_FILE "${MOBILITYDB_LOWERCASE_NAME}--${MOBILITYDB_VERSION}.sql")
> ```
> 
> to
> 
> ```
> set(MOBILITYDB_EXTENSION_FILE "${MOBILITYDB_LOWERCASE_NAME}--${MOBILITYDB_LIB_VERSION}.sql")
> ```
> 
> There is also an issue that now we use cmake instead of gcc for producing SQL definitions depending on the PostgreSQL version. Therefore, now the string @extschema@ is replaced by an empty string in the following kind of definitions
> 
> ```
> CREATE FUNCTION contains(geometry, tgeompoint)
>   RETURNS boolean
>   AS 'SELECT $1 OPERATOR(@extschema@.@>) $2 AND @extschema@._contains($1,$2)'
>   LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
> ```
> 
> Therefore, the resulting definition is erroneous in SQL. The input strings '@extschema@' of the SQL files should be kept as such in the output files. To solve this problem it is necessary to add to the `sql/CMakeLists.txt`
> 
> ```
> # Keep @extschema@ as such
> set(extschema @extschema@)
> ```
> 
> There are still other problems that I am trying to solve but I don't have yet the solution.
> 
> Another issue is that the tests use extensively FIXTURES and these were introduced in cmake version 3.7 while we only require 3.2. See e.g.,
> https://crascit.com/2016/10/18/test-fixtures-with-cmake-ctest/
> 
> Some fixtures for the point and npoint tests were not set up correctly. For the point tests it should be
> 
> ```
> set_tests_properties(${TESTNAME} PROPERTIES
>   FIXTURES_REQUIRED DB ----> DBGEO
>   RESOURCE_LOCK DBLOCK)
> ```
> 
> and for the npoint tests it shoud be
> 
> ```
> set_tests_properties(${TESTNAME} PROPERTIES
>   FIXTURES_REQUIRED DB ----> DBNPOINT
>   RESOURCE_LOCK DBLOCK)
> ```
> 
> @cvvergara Could you please tell us your viewpoint on all these changes? Furthermore, please notice that using cmake instead of gcc produces bigger files since all the comments in the SQL files are kept in the output. Previously, these comments were removed to minimize the size of files when loading the extension.

PR #136, opened 2021-08-09
change 'tiling.pdf' to 'tilling.png' in doc

rchoquet42:


estebanzimanyi:
Could you please try enabling multiple formats as stated in the following link ?
https://github.com/MobilityDB/MobilityDB/pull/137
Please try first without extension and then generate the html, pdf, and ebook generation. If it does not work then please test having multiple graphics in a media object.
Thanks !!!

estebanzimanyi:
Fixed by PR #156

PR #141, opened 2021-08-09
[CI] shell script test process all shell files

cvvergara:
All the files are processed.
There are errors on several files and an issue for each file is going to be opened.

### DIFF
--- a/tools/scripts/test_shell.sh
+++ b/tools/scripts/test_shell.sh
@@ -6,11 +6,17 @@
 DIR=$(git rev-parse --show-toplevel)
 
 pushd "${DIR}" > /dev/null || exit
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
A question that comes to my mind is to whether you want to stop immediately after errors in the first file are found. I would rather report all the files with problems and then exit with code = 1 rather than stopping immediately after finding the first errors. That would reduce the number of cycles in the CI workflow.

estebanzimanyi (approved):

cvvergara:
As you can see here it reports all the files
https://github.com/MobilityDB/MobilityDB/runs/3283597400?check_suite_focus=true#step:4:5

Issue #109, opened 2021-07-29
xzcat program


cvvergara:
Test uses xzcat program, but
1) that is not available on all platgorms as decompression program to xz files
2) Its not looked for on the build

As note for the future:
cmake can do decompression of files https://cmake.org/cmake/help/latest/command/file.html#archive-extract but its available from 3.19 which means that in many operative systems that version might not be available yet.

estebanzimanyi:
Please see the note I put on the closed issue about zstd which is available on 3.15
https://cmake.org/cmake/help/latest/release/3.15.html?highlight=zstd

cvvergara:
Can you give me a link to the issue you mention?

estebanzimanyi:
Last comment in
https://github.com/MobilityDB/MobilityDB/issues/85

cvvergara:
Link to comment:
https://github.com/MobilityDB/MobilityDB/issues/85#issuecomment-888712352

This is what I found about xy files:
https://fileinfo.com/extension/xz
So using something specific that works, say for my computer, it might not work to other systems.
As Regina mentions in this mail:
https://lists.osgeo.org/pipermail/postgis-devel/2021-July/029058.html

> I think using sed and awk in a cmake project is a big NO NO.

Its the same thing for xzcat, the no no, is that they do not exist in other systems.

That comment has a link to cmake 3.15,  if someone is to package with, say bionic, the cmake that they woul be using is 3.10
https://packages.ubuntu.com/bionic/cmake
So if you wants minimum cmake 3.15 then it wouldn't be packaged for bionic (which hasn't reached its EOL).

This is what currently pgRouting has as minimum in order for it to be packaged to several others systems:
https://github.com/pgRouting/pgrouting/blob/main/CMakeLists.txt#L1

I pumped it up in mobilitydb from 3.1 to 3.2
https://github.com/MobilityDB/MobilityDB/blob/master/CMakeLists.txt#L1
https://github.com/MobilityDB/MobilityDB/blob/develop/CMakeLists.txt#L1

Which cmake to be the minimum one, must not be taken lightly, will affect to which systems you are willing to support.

estebanzimanyi:
I think we can simply change xzcat by gzcat. We used xz for its reduced size. For example, compressing the file test/general/data with both xz and gzip shows the difference
```
-rw-r--r-- 1 esteban esteban 2524628 Jul 27 10:46 load.sql.xz
[...]

PR #155, opened 2021-08-14
Add FindXz.cmake, Fix Issue #109

estebanzimanyi:


estebanzimanyi:
Add file FindXz based on
https://github.com/WebKit/webkit/blob/main/Source/cmake/FindXz.cmake
xzcat is in MYS2 
https://packages.msys2.org/package/xz
https://packages.msys2.org/package/mingw-w64-x86_64-xz
and in homebrew 
https://formulae.brew.sh/formula/xz

cvvergara (approved):
As always that coverage variation

PR #156, opened 2021-08-15
Add target doc

estebanzimanyi:


estebanzimanyi:
I think we need to abandon the EPUB format unless there is a better tool than dbtoepub. There are many issues in the generated epub manual. I am looking whether there is an alternative.

cvvergara:
```
doc/images/srf.vsd: *No copyright* UNKNOWN
```

estebanzimanyi:
I have sent an email to the OASIS mailing list for docbook
https://lists.oasis-open.org/archives/docbook-apps/202108/msg00000.html
I have also solved the license problem

cvvergara (approved):
Cool, looks that you are advancing on the documentation

PR #204, opened 2022-04-01
Fix Issue203

estebanzimanyi:


include/general/temporal.h
```diff
@@ -159,10 +166,15 @@ struct tempsubtype_struct
...
- * B:base type passed by value, L: Linear interpolation, C: Continuous base type
+ * Macros for manipulating the 'flags' element with structure xxGTZXLCB, where
+ *   xx: unused bits
+ *   G: coordinates are geodetic
+     T: has T coordinate,
+ *   Z: has Z coordinate
+ *   X: has value or X coordinate
+ *   L: linear interpolation
+ *   C: continuous base type
+ *   B: base type passed by value
```
mschoema, line 177: This code change is not related to the PR, so ideally it should not be present here.

mschoema (approved):

PR #216, opened 2022-05-10
Replace external SQL functions with internal calls

estebanzimanyi:
* Replace input/ouput/send/receive functions with internal calls
* Replace call_functionN with internal calls


estebanzimanyi:
Wrongly targeted master instead of develop

Issue #225, opened 2022-07-16
Version in develop branch has the same number as in master branch.


cvvergara:
**Describe the bug**
Description submitted:

https://lists.osgeo.org/pipermail/mobilitydb-dev/2022-July/000103.html

mschoema:
Version changed to 1.1.0 in PR #223.

PR #226, opened 2022-07-16
Pump up version to 1.1.0

cvvergara:
Related to #225 
Maybe more work needs to be done, but this is a minimal change to have different version number from master.
`mobdb_version.txt` has the version number


estebanzimanyi:
Dear Vicky
Many thanks for pointing this problem out. As you can see in the current PR I am finalizing
https://github.com/MobilityDB/MobilityDB/pull/223
there is a huge restructuring of the MobilityDB codebase in order to build the future MEOS library. For this reason I cannot accept your PR otherwise I will not be able to push my PR. Indeed Github can no longer manage correctly the diffs given the huge difference in the code. I have nevertheless already included your modifications in my PR so they will be available in the forthcoming days as soon as I finish to improve the quality of the code thanks to Codacy
Many thanks for your understanding. 
Esteban

mschoema:
These changes have been added into PR #223

PR #267, opened 2023-01-11
Reduce MEOS API

estebanzimanyi:


mschoema (changes requested):
The compiled binary files of the meos examples have been wrongfully added.
They should be removed

meos/src/general/meos_catalog.c
```diff
@@ -271,7 +342,7 @@ set_basetype(meosType basetype)
 }
 
 /**
- * Ensure that the span base type is supported by MobilityDB
+ * Ensure that the span base type is
```
mschoema, line 345: The change to this comment line should be fixed

mschoema (approved):

Issue #272, opened 2023-01-23
The MFJSON format (in 5.12.1. Input/Output Functions) does not follow the standard version (OGC 19-045r3).


TaehoonK:
**Locate the issue if applicable**
When I checked the user manual ([5.12.1. Input/Output Functions](https://docs.mobilitydb.com/MobilityDB/develop/ch05s12.html#idm3166), MF-JSON related section, such as "_Get the Moving Features JSON representation_"), the Moving Features JSON (shortly MF-JSON) format used the best practice version, not the standard version ([OGC 19-045r3](http://docs.ogc.org/is/19-045r3/19-045r3.html#_mf_json_prism_encoding)).

**Describe the solution you'd like**
If the MobilityDB supports the standard MF-JSON format, the given example should be changed to one of the MF-JSON objects (especially **_[TemporalGeometry](http://docs.ogc.org/is/19-045r3/19-045r3.html#tprimitive)_** or **_[MovingFeature](http://docs.ogc.org/is/19-045r3/19-045r3.html#mfeature)_** object) as below:

Base example (from the [user manual](https://docs.mobilitydb.com/MobilityDB/develop/ch05s12.html#idm3166))
```
{"type": "MovingPoint", "crs":{"type": "name", "properties":{"name": "EPSG:4326"}},
"stBoundedBy":{"bbox":[50.81,4.38,50.81,4.38],
"period":{"begin": "2019-01-01 17:00:00.15+01", "end": "2019-01-01 17:00:00.15+01"}},
"coordinates":[50.81,4.38], "datetimes":"2019-01-01 17:00:00.15+01",
"interpolations":["Discrete"]}
```
Case 1: _**[TemporalGeometry](http://docs.ogc.org/is/19-045r3/19-045r3.html#tprimitive)**_ object in MF-JSON
- "**datetimes**" shall be an array,
- "**type**" of "**crs**" shall be "**Name**" or "**Link**",
- "**interpolations**" should be changed to "**interpolation**"; and "**interpolation**" has an enumeration string, such as "**Discrete**", "**Step**", "**Linear**", "**Quadratic**", and "**Qubic**". 
```
{
	"type": "MovingPoint", 
	"crs":{"type": "Name", "properties":{"name": "EPSG:4326"}},
	"coordinates":[50.81,4.38], 
	"datetimes":["2019-01-01 17:00:00.15+01"],
	"interpolation": "Discrete"
}
```
[...]

PR #280, opened 2023-02-07
 Improve MF-JSON support (Issue #272)

estebanzimanyi:
* Improve MF-JSON support (Issue #272)
* Add abs function for temporal floats
* Improve append functions


meos/src/general/tnumber_mathfuncs.c
```diff
@@ -262,6 +262,124 @@ arithop_tnumber_tnumber(const Temporal *temp1, const Temporal *temp2,
...
+  if (temp->subtype == TINSTANT)
+    result = (Temporal *) tnumberinst_abs((TInstant *) temp);
+  else if (temp->subtype == TSEQUENCE)
+    result = MOBDB_FLAGS_GET_LINEAR(temp->flags) ?
+      (Temporal *) tnumberseq_linear_abs((TSequence *) temp) :
+      (Temporal *) tnumberseq_iter_abs((TSequence *) temp);
+  else /* temp->subtype == TSEQUENCESET */
+    result = (Temporal *) tnumberseqset_abs((TSequenceSet *) temp);
+  return result;
+}
```
mschoema, line 381: I see that there are two functions for the Sequence type, baes on the interpolation of the given sequence.
However, the sequenceset always calls tnumberseq_linear_abs, without looking at the interpolation of its sequences.

I believe tnumberseqset_abs should also check the interpolation of its sequences and accordingly call tnumberseq_linear_abs or tnumberseq_iter_abs.

mobilitydb/sql/general/CMakeLists.txt
```diff
@@ -22,10 +22,10 @@ SET(LOCAL_FILES
...
   036_tnumber_distance
-  037_temporal_similarity
-  038_temporal_aggfuncs
-  040_temporal_waggfuncs
-  042_temporal_gist
+  038_temporal_similarity
+  040_temporal_aggfuncs
+  042_temporal_waggfuncs
+  044_temporal_gist
   044_temporal_spgist
```
mschoema, line 29: Is the identical numbering for temporal_gist and temporal_spgist on purpose or was that a mistake?
I believe it might be better to keep them distinct (ex. 043 and 044).

mschoema (changes requested):

mschoema (approved):

PR #281, opened 2023-02-09
Add stops function

estebanzimanyi:
* Add stops function
* Replace address manipulation functions by MACROS to improve performance
* Fix for new PostgreSQL versions


mschoema (changes requested):
The Stops function does not seem to have the same semantics as the equivalent function in movingpandas.

The movingpandas stops function can be found here:
https://github.com/movingpandas/movingpandas/blob/main/movingpandas/trajectory_stop_detector.py#L59

Our implementation should make sure to produce similar results as the movingpandas function.

mschoema (approved):

PR #287, opened 2023-03-05
Add new functions inspired from Moving Pandas

estebanzimanyi:


meos/src/point/tpoint_analytics.c
```diff
@@ -813,7 +813,7 @@ tsequence_simplify_min_dist(const TSequence *seq, double dist)
         last = true;
     }
   }
-  if (! last)
+  if (k > 1 && ! last)
```
mschoema, line 816: I think this should be:
_if (seq->count > 1 && ! last)_
Indeed, we can have k == 1 if we only maintain the start and end instants (ex. 2 instant sequence where start and end are closer than min_dist

meos/src/point/tpoint_analytics.c
```diff
@@ -907,7 +907,7 @@ tsequence_simplify_min_tdelta(const TSequence *seq, const Interval *mint)
     }
     pfree(duration);
   }
-  if (! last)
+  if (k > 1 && ! last)
```
mschoema, line 910: Same comment as above

CMakeLists.txt
```diff
@@ -159,8 +159,10 @@ if(CMAKE_COMPILER_IS_GNUCC)
   endif()
 endif()
 
+# Set compiler optimizations for release
+set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_DEBUG} -O2")
```
mschoema, line 163: This should probably be:
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -O2")
instead of
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_DEBUG} -O2")

mschoema (approved):

PR #293, opened 2023-03-16
Add macros for accessing sequence (set) elements

estebanzimanyi:


meos/src/general/temporal.c
```diff
@@ -2496,7 +2496,7 @@ temporal_start_instant(const Temporal *temp)
   if (temp->subtype == TINSTANT)
     result = (TInstant *) temp;
   else if (temp->subtype == TSEQUENCE)
-    result = tsequence_inst_n((TSequence *) temp, 0);
+    result = tsequence_inst_n((TSequence *)(temp), 0);
```
mschoema, line 2499: Is the (temp) necessary or can it be omitted

meos/include/meos.h
```diff
@@ -173,10 +173,11 @@ typedef struct
   int32 count;          /**< Number of TInstant elements */
   int32 maxcount;       /**< Maximum number of TInstant elements */
   int16 bboxsize;       /**< Size of the bounding box */
-  Span period;          /**< Time span (24 bytes). All bounding boxes
-                             start with a period so actually it is also
-                             the begining of the bounding box. The extra
-                             bytes needed are added upon creation. */
+  char padding[6];      /**< Not used */
```
mschoema, line 176: Why do we have a padding of 6 here?

meos/include/meos_internal.h
```diff
@@ -287,6 +287,49 @@ extern void stbox_set_box3d(const STBox *box, BOX3D *box3d);
  * Functions for temporal types
  *****************************************************************************/
 
+/* Macros for speeding up access to components of temporal sequences (sets)*/
+
+#ifdef DEBUG_BUILD
+extern const TInstant *tsequence_inst_n(const TSequence *seq, int index);
+extern const TSequence *tsequenceset_seq_n(const TSequenceSet *ss, int index);
```
mschoema, line 294: It seems we do not define _tsequence_offset_ptr_ as extern in the debug build.
Is this on purpose?
If so, maybe set the function to static when defining it.

meos/src/general/tsequence.c
```diff
@@ -759,23 +757,32 @@ tsequence_set_bbox(const TSequence *seq, void *box)
...
+#ifdef DEBUG_BUILD
+/**
+ * @brief Function version of the the macro of the same name for
+ * debugging purposes
+ */
+size_t *
+tsequence_offsets_ptr(const TSequence *seq)
+{
+  return (size_t *)( ((char *) &seq->period) + seq->bboxsize );
+}
```
mschoema, line 769: See previous comment

meos/src/general/tsequenceset.c
```diff
@@ -142,34 +142,30 @@ tsequenceset_set_bbox(const TSequenceSet *ss, void *box)
...
+ * @brief Function version of the the macro of the same name for
  * debugging purposes
  */
 size_t *
 tsequenceset_offsets_ptr(const TSequenceSet *ss)
 {
-  return (size_t *)(((char *)(ss)) + DOUBLE_PAD(sizeof(TSequenceSet)) +
-    DOUBLE_PAD((ss)->bboxsize - sizeof(Span)));
+  return (size_t *)( ((char *) &ss->period) + ss->bboxsize );
 }
```
mschoema, line 152: Same comment as for _tsequence_offset_ptr_

meos/include/meos_internal.h
```diff
@@ -287,6 +287,49 @@ extern void stbox_set_box3d(const STBox *box, BOX3D *box3d);
...
+
+/**
+ * @brief Return the n-th sequence of a temporal sequence set
+ * @note The period component of the bbox is already declared in the struct
+ * @pre The argument @p index is less than the number of sequences in the
+ * sequence set
+ */
+#define tsequenceset_seq_n(ss, index) ( (TSequence *)( \
+  ((char *) &((ss)->period)) + (ss)->bboxsize + \
+  (sizeof(size_t) * (ss)->maxcount) + (tsequenceset_offsets_ptr(ss))[index] ) )
```
mschoema, line 328: Should we write these function using CAPS to denote the fact that they are macros?
Just so we know to avoid writing something like:
`tsequence_inst_n(long_function_returning_a_seq(...), i)`
since this would cause the long function to be evaluated multiple times.

PR #297, opened 2023-03-20
Fix bug with shift function

mschoema:


mschoema:
4 bugs fixed:
- Incorrect upper value in `span_shift`
- Shifting and scaling of sequence bbox was done twice in `tsequence_shift_tscale`
- Shifting and scaling of sequenceset bbox wasn't done in `tsequenceset_shift_tscale`
- Incorrect computation of the bboxes of instant sequences in `tsequenceset_shift_tscale`

Additionally, refractored the code to compute `delta` and `shift` inside `period_shift_tscale`, since this was repeated in the 3 other `*_shift_tscale` functions.

estebanzimanyi (approved):

PR #298, opened 2023-03-20
Integration with PyMEOS

estebanzimanyi:
* Extend trajectory function for tint


meos/src/general/temporal.c
```diff
@@ -1821,7 +1821,9 @@ temporal_interpolation(const Temporal *temp)
   char *result = palloc(sizeof(char) * MOBDB_INTERPOLATION_STR_MAXLEN);
   assert(temptype_subtype(temp->subtype));
   interpType interp = MOBDB_FLAGS_GET_INTERP(temp->flags);
-  if (temp->subtype == TINSTANT || interp == DISCRETE)
+  if (temp->subtype == TINSTANT)
+    strcpy(result, "None");
+  else if (interp == DISCRETE)
```
mschoema, line 1826: I thought we were renaming DISCRETE to NONE.
What is here the difference between "None" and "Discrete"?

meos/src/point/CMakeLists.txt
```diff
@@ -13,12 +13,9 @@ add_library(point OBJECT
...
-
-if(MEOS)
-add_library(general OBJECT
+#if MEOS
   tpoint_meos.c
   tpoint_spatialrels_meos.c
   tpoint_tempspatialrels_meos.c
+#endif /* MEOS */
 )
-endif()
```
mschoema, line 24: same as above

mschoema (approved):

PR #305, opened 2023-04-03
Add quadSpan function

estebanzimanyi:


.github/workflows/windows.yml
mschoema, line 1: What is the purpose of this file, is it supposed to add a new test case for windows?
If so, something is going wrong, as there is no new test added to the list.

meos/src/point/stbox.c
```diff
@@ -1505,15 +1505,15 @@ intersection_stbox_stbox(const STBox *box1, const STBox *box2)
...
 {
   ensure_has_X_stbox(box);
   bool hasz = MOBDB_FLAGS_GET_Z(box->flags);
   bool hast = MOBDB_FLAGS_GET_T(box->flags);
   bool geodetic = MOBDB_FLAGS_GET_GEODETIC(box->flags);
   Span *period = hast ? (Span *) &box->period : NULL;
-  int count = hasz ? 8 : 4;
-  STBox *result = palloc(sizeof(STBox) * count);
+  *count = hasz ? 8 : 4;
+  STBox *result = palloc(sizeof(STBox) * *count);
```
mschoema, line 1516: This probably requires some additional parenthesis around `*count`

mschoema (changes requested):

estebanzimanyi:
________________________________

mschoema (approved):

PR #306, opened 2023-04-05
Add setInterp function

estebanzimanyi:


meos/src/general/temporal.c
```diff
@@ -137,6 +125,51 @@ ensure_continuous_interpolation(int16 flags)
...
+  bool strict)
+{
+  if (! strict || (temp1->subtype == TINSTANT || temp2->subtype == TINSTANT))
+    return;
+  interpType interp1 = MOBDB_FLAGS_GET_INTERP(temp1->flags);
+  interpType interp2 = MOBDB_FLAGS_GET_INTERP(temp2->flags);
+  if (interp1 != interp2)
+    elog(ERROR, "The temporal values must have the same interpolation");
+  return;
+}
```
mschoema, line 171: What is the point of the `strict` parameter here?
If it is only used to tell if one of the parameters may be of instant subtype, we can probably compute it inside the function:
```c
bool strict = !(temp1->subtype == TINSTANT || temp2->subtype == TINSTANT);
```
In the current implementation, calling the function with `strict = false` does nothing.

mobilitydb/sql/point/052_tpoint.in.sql
```diff
@@ -229,10 +229,11 @@ CREATE FUNCTION tgeompoint_inst(tgeompoint)
   RETURNS tgeompoint
   AS 'MODULE_PATHNAME', 'Temporal_to_tinstant'
   LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
-CREATE FUNCTION tgeompoint_seq(tgeompoint, text DEFAULT 'linear')
+-- The function is NOT STRICT since the second argument may be null
+CREATE FUNCTION tgeompoint_seq(tgeompoint)
   RETURNS tgeompoint
   AS 'MODULE_PATHNAME', 'Temporal_to_tsequence'
-  LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+  LANGUAGE C IMMUTABLE PARALLEL SAFE;
```
mschoema, line 236: I think this can stay `STRICT`, since we want `tgeompoint_seq(NULL) = NULL`

mobilitydb/sql/point/052_tpoint.in.sql
```diff
@@ -242,22 +243,23 @@ CREATE FUNCTION tgeogpoint_inst(tgeogpoint)
   RETURNS tgeogpoint
   AS 'MODULE_PATHNAME', 'Temporal_to_tinstant'
   LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
-CREATE FUNCTION tgeogpoint_seq(tgeogpoint, text DEFAULT 'linear')
+-- The function is NOT STRICT since the second argument may be null
+CREATE FUNCTION tgeogpoint_seq(tgeogpoint)
   RETURNS tgeogpoint
   AS 'MODULE_PATHNAME', 'Temporal_to_tsequence'
-  LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+  LANGUAGE C IMMUTABLE PARALLEL SAFE;
```
mschoema, line 250: Same here

PR #349, opened 2023-07-13
Correctly interpolate between geographic points

mschoema:


meos/src/point/tpoint_spatialfuncs.c
estebanzimanyi, line 1: Typo -> Cartesian

estebanzimanyi (approved):

PR #351, opened 2023-07-19
Fix cmake error on PopOS 22.04 - GEOS versions not set

rustprooflabs:
# Details

This PR adjusts the `FindGEOS.cmake` to fix a problem I encountered running `cmake ..` on PopOS 22.04. 

Before the PR running `cmake ..` resulted in an error parsing the GEOS versions.  The version numbers were not printed as expected in the output.

```
-- Found GEOS: /usr/lib/x86_64-linux-gnu/libgeos_c.so ()
-- Found GEOS: /usr/lib/x86_64-linux-gnu/libgeos_c.so ()
-- GEOS_INCLUDE_DIR: /usr/include
-- GEOS_LIBRARY: /usr/lib/x86_64-linux-gnu/libgeos_c.so
-- GEOS_VERSION: 
-- GEOS_VERSION_MAJOR: 
-- GEOS_VERSION_MINOR: 
CMake Error at CMakeLists.txt:140 (math):
  math cannot parse the expression: " * 10 + ": syntax error, unexpected
  exp_TIMES (2).
```

I didn't make any attempt to try to figure out why the value wasn't parsing from `${GEOS_INCLUDE_DIR}/geos_c.h`.  It seems the regex match there should work, but I'm terrible at regex so this was what I was able to make work quickly.  If you don't want to merge this PR in favor of a better approach to making this work I'm okay with that.


rustprooflabs:
I noticed today there was a lot of work after the latest `master` branch on the `dev` branch.  It seems this bug is fixed in `develop` but hasn't been merged to `master` yet.

PR #367, opened 2023-08-15
Fix for mingw64 PG16

### LINKED ISSUE #366: mingw64 building of mobilitydb for PostgreSQL 16, CREATE EXTENSION fails
**Describe the bug**
This seems to be only an issue under mingw64 and with PG 16 specifically.  PG 15 for example is fine.

This is building the 1.1.0alpha package

I tested on debian and that seemed okay.

The extension builds under mingw64, but when I go to do 

```
CREATE EXTENSION mobilitydb;
```

I get an error

```
ERROR:  could not find function "Left_stbox_tnpoint" in file "C:/ming64gcc81/projects/postgresql/rel/pg16w64gcc81/lib/libMobilityDB-1.1.dll" 
```

**Specifications (please complete the following information):**
Use the commands:
```SQL
SELECT version();  
SELECT postgis_full_version();
-- can't run cause can't inatall
SELECT mobilitydb_full_version();
```

Output
```
                                                         version

-------------------------------------------------------------------------------------------------------------------------
--------
 PostgreSQL 16beta3 on x86_64-w64-mingw32, compiled by gcc.exe (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 8.1.0,
 64-bit

 POSTGIS="3.4.0 3.4.0rc4-1-g07a11e15d" [EXTENSION] PGSQL="160" GEOS="3.12.0dev-CAPI-1.18.0" PROJ="8.2.1 NETWORK_ENABLED=O
FF URL_ENDPOINT=https://cdn.proj.org USER_WRITABLE_DIRECTORY=C:\Users\Administrator\AppData\Local/proj DATABASE_PATH=C:\m
ing64gcc81\projects\proj\rel-proj-8.2.1w64gcc81/share/proj/proj.db" LIBXML="2.9.14" LIBJSON="0.12" LIBPROTOBUF="1.2.1" WA
GYU="0.5.0 (Internal)"
(1 row)
```

**Additional context**
At a glance, I'm not seeing anything about how this function is defined that would cause this issue.  Even more surprising why it's fine on PG15.

Last time I had this issue I think was something about how the PGDLLIMPORT something about the ordering.

I'll see if I can nail down the issue.

robe2:
Closes #366
Move meos.h include after tnpoint_boxops.h
otherwise some functions don't get exported under mingw64 pg16.

Not sure why it's only an issue with pg16 and not lower PGs, might be because they reshuffled includes too in PG16

### DIFF
--- a/mobilitydb/src/npoint/tnpoint_posops.c
+++ b/mobilitydb/src/npoint/tnpoint_posops.c
@@ -43,10 +43,10 @@
  * arguments is a stbox.
  */
 
-/* MEOS */
-#include <meos.h>
 /* MobilityDB */
 #include "pg_npoint/tnpoint_boxops.h"
+/* MEOS */
+#include <meos.h>
 
 /*****************************************************************************/
 /* stbox op Temporal */


robe2:
Not needed anymore for some reason.

PR #402, opened 2023-12-04
Fix spgist indices

mschoema:
Changes:
- Add levels correctly when traversing a kdtree
- Always assume inclusive bounds of the span quadtree nodes. Not as optimal as handling the bounds correctly, but this would sometimes cause empty spans to be created, so this was the simplest fix.
- Use the same code for spam overlaps and contains in both quadtree and kdtree. I believe the code for kdtree was trying to only test one dimension at a time, but it was wrong, so I just used the same code as quadtree. This is similar to what spgist does for tbox and stbox.


estebanzimanyi:
Amazing job Maxime !

estebanzimanyi (approved):

PR #412, opened 2023-12-19
Improve MobilityDB code

estebanzimanyi:


meos/include/meos.h
```diff
@@ -1328,7 +1332,7 @@ extern Temporal *temporal_scale_time(const Temporal *temp, const Interval *durat
 extern Temporal *temporal_set_interp(const Temporal *temp, interpType interp);
 extern Temporal *temporal_shift_scale_time(const Temporal *temp, const Interval *shift, const Interval *duration);
 extern Temporal *temporal_shift_time(const Temporal *temp, const Interval *shift);
-extern Temporal *temporal_to_tinstant(const Temporal *temp);
+extern TInstant *temporal_to_tinstant(const Temporal *temp);
 extern Temporal *temporal_to_tsequence(const Temporal *temp, interpType interp);
 extern Temporal *temporal_to_tsequenceset(const Temporal *temp, interpType interp);
```
mschoema, line 1337: Why does `temporal_to_tinstant` return a `TInstant`, while `temporal_to_tsequence` and `temporal_to_tsequenceset` still return a `Temporal`

meos/include/general/meos_catalog.h
```diff
@@ -37,9 +37,12 @@
 /* C */
 #include <stdbool.h>
 /* PostgreSQL */
-#ifndef int16
-typedef signed short int16;
-#endif
+#include <postgres.h>
+// #ifndef int16
+// typedef signed short int16;
+// #endif
```
mschoema, line 43: We cannot `#include <postgres.h>`, since meos_catalog.h gets installed together with meos.h and meos_internal.h when we compile MEOS. Please revert this change

mschoema (changes requested):

meos/src/general/meos.c
```diff
@@ -0,0 +1,467 @@
...
+char *
+meos_get_datestyle(void)
+{
+  char *result = palloc(32);
+  if (!result)
+    return false;
+  sprintf(result, "%s, %s", datestyle_string(DateStyle), 
+    dateorder_string(DateStyle));
+  return result;
+}
```
mschoema, line 385: This should probable raise an error or return `NULL` instead of returning `false` when `palloc` fails.

meos/src/general/meos.c
```diff
@@ -0,0 +1,467 @@
...
+ */
+char *
+meos_get_intervalstyle(void)
+{
+  char *result = palloc(32);
+  if (!result)
+    return false;
+  sprintf(result, "%s", intervalstyle_string(IntervalStyle));
+  return result;
+}
```
mschoema, line 441: Same as above

mschoema (changes requested):

mschoema (approved):

PR #417, opened 2024-01-03
Improve code quality

estebanzimanyi:
* Improve documentation
* Improve spaceTimeSplit
* Improve distance functions
* Improve indexes
* Improve ever/always functions


meos/src/general/lifting.c
```diff
@@ -1551,211 +1543,438 @@ tfunc_temporal_temporal(const Temporal *temp1, const Temporal *temp2,
...
+              (TInstant *) temp1, (TSequence *) temp2, lfinfo) :
+            (Temporal *) tfunc_tinstant_tcontseq(
+              (TInstant *) temp1, (TSequence *) temp2, lfinfo);
+        case TSEQUENCESET:
+          return (Temporal *) tfunc_tinstant_tsequenceset(
+            (TInstant *) temp1, (TSequenceSet *) temp2, lfinfo);
+        default:
+          meos_error(ERROR, MEOS_ERR_INTERNAL_ERROR,
+            "Unexpected temporal subtype: %d", temp2->subtype);
+          return NULL;
```
mschoema, line 1569: We do not need the default in the switch case.
Meos errors should be used for runtime errors, such as wrong parameters, invalid sequence lengths and similar stuff..
The default check here is to test for correctness, which should only be done with the asserts (already present above) in debug mode. At runtime, this default will never be hit, unless there is a bug, which we can detect in debug mode.

This is also true for all other switch cases on subtype.

estebanzimanyi:
Merci !!! Je croyais avoir enlevé cela partout
________________________________
De : Maxime Schoemans ***@***.***>
Envoyé : lundi 8 janvier 2024 14:11
À : MobilityDB/MobilityDB ***@***.***>
Cc : ZIMANYI Esteban ***@***.***>; Author ***@***.***>
Objet : Re: [MobilityDB/MobilityDB] Improve code quality (PR #417)


@mschoema commented on this pull request.

________________________________

In meos/src/general/lifting.c<https://github.com/MobilityDB/MobilityDB/pull/417#discussion_r1444613093>:

> +        default:
+          meos_error(ERROR, MEOS_ERR_INTERNAL_ERROR,
+            "Unexpected temporal subtype: %d", temp2->subtype);
+          return NULL;


We do not need the default in the switch case.
Meos errors should be used for runtime errors, such as wrong parameters, invalid sequence lengths and similar stuff..
The default check here is to test for correctness, which should only be done with the asserts (already present above) in debug mode. At runtime, this default will never be hit, unless there is a bug, which we can detect in debug mode.

This is also true for all other switch cases on subtype.

mobilitydb/src/general/temporal_compops.c
```diff
@@ -47,7 +49,594 @@
...
+ * @ingroup mobilitydb_temporal_comp_ever
+ * @brief Return true if a base value is ever less than a temporal value
+ * @sqlfn ever_lt()
+ * @sqlop @p ?<
+ */
+Datum
+Ever_lt_base_temporal(PG_FUNCTION_ARGS)
+{
+  return EAcomp_base_temporal(fcinfo, &ever_lt_temporal_base);
+}
```
mschoema, line 238: Shouldn't we invert the operator too here?
We're inverting the order of the parameters but not the operator, although it is not commutative.

mschoema: The same happens later too

mobilitydb/src/point/tpoint.c
```diff
@@ -94,6 +94,7 @@ void
 temporalgeom_init()
 {
   lwgeom_set_handlers(palloc, repalloc, pfree, pg_error, pg_notice);
+  return;
```
mschoema, line 97: _Adding empty return to void functions_ could also have been a separate PR to simplify things.

mschoema (changes requested):

meos/src/general/lifting.c
```diff
@@ -1560,13 +1560,9 @@ tfunc_temporal_temporal(const Temporal *temp1, const Temporal *temp2,
               (TInstant *) temp1, (TSequence *) temp2, lfinfo) :
             (Temporal *) tfunc_tinstant_tcontseq(
               (TInstant *) temp1, (TSequence *) temp2, lfinfo);
-        case TSEQUENCESET:
+        default: /* TSEQUENCESET */
```
mschoema, line 1563: Don't use `default` for the last case.
Since our possible cases are fixed, simply put each in a separate `case ...` and don't use a `default` at all for this match case section.

meos/src/general/lifting.c
estebanzimanyi: Solved

meos/src/general/lifting.c
```diff
@@ -1551,211 +1543,438 @@ tfunc_temporal_temporal(const Temporal *temp1, const Temporal *temp2,
...
  */
 static int
-efunc_tinstant_tinstant(const TInstant *inst1, const TInstant *inst2,
+eafunc_tinstant_tinstant(const TInstant *inst1, const TInstant *inst2,
   LiftedFunctionInfo *lfinfo)
 {
+  assert(inst1); assert(inst2); assert(inst1->temptype == inst2->temptype);
   /* The following is ensured by the period bound test in the dispatch
    * function */
   assert(inst1->t == inst2->t);
```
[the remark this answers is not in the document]
estebanzimanyi: As explained in the comment before the assert, the dispatch function eafunc_temporal_temporal has a bounding box test so that when we arrive here we never need to return a -1 value.

mobilitydb/sql/general/001_set.in.sql
```diff
@@ -347,7 +347,7 @@ CREATE FUNCTION asHexWKB(tstzset, endianenconding text DEFAULT '')
  * Constructor functions
  ******************************************************************************/
 
-CREATE FUNCTION set(int[])
```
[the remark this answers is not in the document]
estebanzimanyi: Noted for the future PRs

mobilitydb/src/point/tpoint.c
estebanzimanyi: Noted for future PRs

meos/src/general/lifting.c
estebanzimanyi: Alas, doing that will produce warnings that we would like to avoid. This is the reason of having an assert before the switch

mschoema: The issue comes from 2 aspects:
1. the `subtype` field is a `uint8` instead of the `tempSubType` enum type for performance reasons (since we know we don't need more than 1 byte for the subtype value).
2. `tempSubType` has a 4th value `ANYTEMPSUBTYPE`

My ideas:

For 1: cast `temp->subtype` to `tempSubType` in a match statement: `match ((tempSubType) temp->subtype) ...`. This will let the compiler know that we are matching over an enum.
If we cover all cases we will thus not have any warnings.

For 2, two possible solutions:
1. Add an additional `case ANYTEMPSUBTYPE` instead of `default`, containing a meos_error.
2. Remove `ANYTEMPSUBTYPE` from the enum values and add it as a `#define ANYTEMPSUBTYPE 0` right above the enum definition. This should not break any of the other code in analyze and selectivity using `ANYTEMPSUBTYPE` (I tested the compilation and there were no errors or warnings) and will allow us to switch over only the 3 values in `tempSubType`

I think solution 2 is better with regard to the `match case` solution, but I'd accept solution 1 too.

meos/src/general/lifting.c
```diff
@@ -2076,23 +2321,27 @@ efunc_tsequence_tsequenceset(const TSequence *seq, const TSequenceSet *ss,
...
   const TSequenceSet *ss2, LiftedFunctionInfo *lfinfo)
 {
+  assert(ss1); assert(ss2); assert(ss1->temptype == ss2->temptype);
   int i = 0, j = 0;
   while (i < ss1->count && j < ss2->count)
   {
     const TSequence *seq1 = TSEQUENCESET_SEQ_N(ss1, i);
     const TSequence *seq2 = TSEQUENCESET_SEQ_N(ss2, j);
-    if (efunc_tcontseq_tcontseq(seq1, seq2, lfinfo) == 1)
+    int res = eafunc_tcontseq_tcontseq(seq1, seq2, lfinfo);
```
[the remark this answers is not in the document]
estebanzimanyi: By definition the sequences composing a sequence set are ALWAYS continuous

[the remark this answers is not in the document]
mschoema: Ah yes, I was thinking of stepwise vs linear, but they are both handled in `eafunc_tcontseq_tcontseq`

meos/src/general/lifting.c
```diff
@@ -1766,28 +1985,31 @@ efunc_tdiscseq_tdiscseq(const TSequence *seq1, const TSequence *seq2,
...
+        return 0;
       i++;
     }
     else if (cmp < 0)
       inst1 = TSEQUENCE_INST_N(seq1, ++i);
     else
       inst2 = TSEQUENCE_INST_N(seq2, ++j);
   }
-  return 0;
+  return lfinfo->ever ? 0 : 1;
```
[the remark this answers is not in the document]
estebanzimanyi: Good point ! This case only happens for the two functions eafunc_tdiscseq_tdiscseq and eafunc_tsequenceset_tsequenceset where intersection of the time spans of the two temporal values is needed in addition to the bounding period test done in the dispatch function eafunc_temporal_temporal.
Corrected !

meos/src/general/lifting.c
mschoema: We ended up using `default` for the `TSEQUENCESET` case, as this is similar to the previous `if else` statement syntax. The `assert` makes sure that this default only hits `TSEQUENCESET` cases.

mschoema (approved):

PR #427, opened 2024-01-18
Improve text and mathematical functions

estebanzimanyi:


meos/src/general/set.c
```diff
@@ -1991,6 +2060,81 @@ textset_upper(const Set *s)
...
+Set *
+textcat_text_textset(const text *txt, const Set *s)
+{
+  /* Ensure validity of the arguments */
+  if (! ensure_not_null((void *) s) || ! ensure_not_null((void *) txt) || 
+      ! ensure_set_isof_type(s, T_TEXTSET))
+    return NULL;
+
+  return textcat_textset_text_int(s, txt, INVERT_NO);
+}
```
mschoema, line 2117: This should be `INVERT` instead of `INVERT_NO`

meos/src/general/ttext_textfuncs.c
```diff
@@ -234,6 +187,39 @@ datum_upper(Datum value)
...
+}
+
+/**
+ * @brief Convert the text value to uppercase
+ */
+Datum
+datum_initcap(Datum value)
+{
+  return pg_initcap(DatumGetTextP(value));
+}
```
mschoema, line 221: Do we need 2 functions?
If so, `pg_initcap` should probably return a `text *` instead of a `Datum`.

meos/src/general/ttext_textfuncs_meos.c
```diff
@@ -41,6 +41,21 @@
...
+ * @param[in] txt1,txt2 Values
+ */
+text *
+textcat_text_text(const text *txt1, const text *txt2)
+{
+  /* Ensure validity of the arguments */
+  if (! ensure_not_null((void *) txt1) || ! ensure_not_null((void *) txt2))
+    return NULL;
+  return DatumGetTextP(datum_textcat(PointerGetDatum(txt1), 
+    PointerGetDatum(txt2)));
```
mschoema, line 56: I believe we have a function called `text_catenate`, which uses `text *`. Here we are casting to `Datum` simply for `datum_textcat` to cast back to `text *` and call `text_catenate`

mschoema (changes requested):

mschoema (approved):

PR #431, opened 2024-01-28
Pymeos sync

Diviloper:


meos/include/meos.h
```diff
@@ -1625,13 +1626,17 @@ extern Temporal *teq_float_tfloat(double d, const Temporal *temp);
 extern Temporal *teq_int_tint(int i, const Temporal *temp);
 extern Temporal *teq_point_tpoint(const GSERIALIZED *gs, const Temporal *temp);
 extern Temporal *teq_tbool_bool(const Temporal *temp, bool b);
+extern Temporal *teq_tbool_tbool(const Temporal *temp1, const Temporal *temp2);
```
mschoema, line 1629: Do we need each iteration of `teq_tbool_tbool`, `teq_tfloat_tfloat`, ... if we already have `teq_temporal_temporal`?
Same holds for the other `tne_*` and the other comparison functions.

meos/src/general/type_round.c
```diff
@@ -880,3 +880,4 @@ tnpoint_round(const Temporal *temp, Datum size)
 }
 
 /*****************************************************************************/
+#endif
```
mschoema, line 883: Github is showing a weird stop sign and colouring this line red.
Not sure why this is the case, maybe a special invisible character or line ending?
Maybe EOF character is missing?

Diviloper: I've added a new line and I don't see the stop sign anymore. So it was probably that

meos/include/meos.h
Diviloper: The thing is `teq_temporal_temporal` is not implemented, so I added the other functions (`teq_tbool_tbool` etc.), but forgot to remove `teq_temporal_temporal` from here. The same with `tne_temporal_temporal` and actually all of the rest too (`tlt`, `tle`, etc).

mschoema: Ah yes I see, the functions already exist but were not declared in `meos.h`.
Could you then add the necessay declarations for `*_tbool_tbool`, `*_tfloat_tfloat`, etc for all comparison operators and remove the correspoding `*_temporal_temporal` definitions?
I'll accept the PR when this is done.

mschoema: After second thought, if you could instead keep the `*_temporal_temporal` version of the functions and remove the `*_tbool_tbool`, `*_tfloat_float`, etc, that would be a better solution as it reduces the number of functions and is similar to what is being done in the rest of MEOS.

mschoema: I'm not sure they were ever implemented, but if you look at the current implementation of `teq_tbool_tbool`, for example, you can easily adapt it to `teq_temporal_temporal`:

```c
Temporal *
teq_tbool_tbool(const Temporal *temp1, const Temporal *temp2)
{
  /* Ensure validity of the arguments */
  if (! ensure_not_null((void *) temp1) || ! ensure_not_null((void *) temp2) ||
      ! ensure_same_temporal_type(temp1, temp2) ||
      ! ensure_temporal_isof_type(temp1, T_TBOOL))
    return NULL;
  return tcomp_temporal_temporal(temp1, temp2, &datum2_eq);
}
```

becomes

```c
Temporal *
teq_temporal_temporal(const Temporal *temp1, const Temporal *temp2)
{
  /* Ensure validity of the arguments */
  if (! ensure_not_null((void *) temp1) || ! ensure_not_null((void *) temp2) ||
      ! ensure_same_temporal_type(temp1, temp2))
    return NULL;
  return tcomp_temporal_temporal(temp1, temp2, &datum2_eq);
}
```

and similarly for the other comparison operators.
These functions are in `/meos/src/general/temporal_compops_meos.c`.

Diviloper: Tbh that's easier than I expected it would be.
I will update it for all the temporal comparisons.

Diviloper: It may not be as easy as I thought, due to the tpoint comparisons. Instead of `tcomp_temporal_temporal`, they use `tcomp_tpoint_tpoint`, which checks extra stuff. I added the call to that function when comparing tpoints, but compiling a program against the library fails with a
```shell
/usr/bin/ld: /usr/local/lib/libmeos.so: undefined reference to `tcomp_tpoint_tpoint'
``` 
I've just tried importing the header file where `tcomp_tpoint_tpoint` is defined (`point/tpoint.h`), but had the same problem. I also tried adding the declaration to `meos_internal.h`, but then it's MEOS who fails to compile.

I'll investigate further later, but any suggestion is welcome. I've added the changes in the [`pymeos_sync_tcomp` branch](https://github.com/Diviloper/MobilityDB/tree/pymeos_sync_tcomp) of my fork (see https://github.com/Diviloper/MobilityDB/commit/b8af4f3b4f5c3f8883dfca9d6e30361ba47c7846), in case you want to check.

mschoema: It seems that the function `tcomp_tpoint_tpoint` is indeed defined in `point/tpoint.h`, but it is declared as `static` in `point/tpoint_compops.c`, so it will indeed not be available outside of that file. I think just removing the `static` should fix it.

However, it also seems that the `teq_tpoint_tpoint` and `tne_tpoint_tpoint` functions you removed are being used in MobilityDB code. We thus have to keep them or call another method from MobilityDB.

I would opt for changing the code of `teq_temporal_temporal` to something like:

```c
Temporal *
teq_temporal_temporal(const Temporal *temp1, const Temporal *temp2)
{
  /* Ensure validity of the arguments */
  if (! ensure_not_null((void *) temp1) || ! ensure_not_null((void *) temp2) ||
      ! ensure_same_temporal_type(temp1, temp2))
    return NULL;
  if (tgeo_type(temp1->temptype))
    if (! ensure_same_srid(tpoint_srid(temp1), tpoint_srid(temp2)) ||
        ! ensure_same_dimensionality(temp1->flags, temp2->flags))
      return NULL;
  return tcomp_temporal_temporal(temp1, temp2, &datum2_eq);
}
```

Of course you can put everything in a single if statement if you manage to find a clean syntax.

With this definition, we can fully remove the `tcomp_tpoint_tpoint` function and we can call `teq_temporal_temporal` instead of `teq_tpoint_tpoint` in MobilityDB.

What do you think?

Diviloper: Sounds good to me. I'm working on it now, hopefully I'll have it done soon.

Diviloper: Okay, I think it's fixed. `tcomp_tpoint_tpoint` no longer exists. I initially changed the MobilityDB functions that were calling `teq_tpoint_tpoint` and `tne_tpoint_tpoint` (`Teq_tpoint_tpoint` and `Tne_tpoint_tpoint` respectively) to call the `t**_temporal_temporal` version. (f4d3a56) 
However, I couldn't make this work, and the library was crashing during loading due to a undefined symbol error. Maybe it was just a missing import or something like this, but I ended up just deleting those MobilityDB functions, and using the generic temporal versions.

Diviloper: I just realized that the MobilityDB function were not equivalent. The `Teq_tpoint_tpoint` passed through `teq_temporal_temporal`, while `Teq_temporal_temporal` doesn't, it goes straight to `tcomp_temporal_temporal`. I assume the checks done are not necessary for the rest of temporal types, since being of the same temporal type and not null is checked at the SQL level (is this right?), but they are necessary for the tpoints to check for the dimensionality and srid.
I'm reverting/fixing this now

Diviloper: Okay, it was just "something like this". The `teq_temporal_temporal` and `tne_temporal_temporal` functions were defined in `temporal_compops_meos.c` which is not part of the `general` library that MobilityDB uses. I just moved their definitions to `temporal_compops.c` which is part of it.

meos/src/general/temporal_compops.c
```diff
@@ -725,5 +725,50 @@ tcomp_temporal_temporal(const Temporal *temp1, const Temporal *temp2,
...
+  /* Ensure validity of the arguments */
+  if (! ensure_not_null((void *) temp1) || ! ensure_not_null((void *) temp2) ||
+      ! ensure_same_temporal_type(temp1, temp2))
+    return NULL;
+  if (tgeo_type(temp1->temptype) && (
+          ! ensure_same_srid(tpoint_srid(temp1), tpoint_srid(temp2)) ||
+          ! ensure_same_dimensionality(temp1->flags, temp2->flags)))
+    return NULL;
+  return tcomp_temporal_temporal(temp1, temp2, &datum2_ne);
+}
```
mschoema, line 774: I believe something went wrong with the file ending again

meos/src/general/temporal_compops_meos.c
```diff
@@ -39,6 +39,7 @@
 #include <meos_internal.h>
 #include "general/temporal.h"
 #include "general/type_util.h"
+#include "point/tpoint_spatialfuncs.h"
```
mschoema, line 42: This should probably be moved to temporal_compops.c

meos/include/meos.h
mschoema: > I assume the checks are not necessary for the rest of temporal types, since being of the same temporal type and not null is checked at the SQL level (is this right?), but they are necessary for the tpoints to check for the dimensionality and srid.

Indeed, the functions are `STRICT`, so they automatically return `NULL` when one of the arguments is `NULL`. They can also only be called with the specified parameter types, so both inputs will always have the same type. The same happens for tpoint types, but we indeed still need to check their srid and dimensionality.

mschoema (approved):

PR #443, opened 2024-02-18
Improve SQL API

estebanzimanyi:
* Uniformize the name of SQL functions in CamelCase removing underscores
* Remove subtype suffix (i.e., Inst, Seq, SeqSet) from constructors having a base type as first argument


mschoema:
This PR should be done on develop, not master

PR #445, opened 2024-02-20
Add PJ definition in postgis_ext_defs.in.h

Diviloper:
The previous PR (#441) was not completely ready, as it was missing some pieces to make it usable in PyMEOS.
I've added the definition of `PJ` (copied from the [proj.h header](https://github.com/OSGeo/PROJ/blob/deb1a5720fdfcd50d53c93ffb4011cacb36f2c89/src/proj.h#L221)).
I've checked all versions from 6.1 (the minimum required by MobilityDB) until the last (9.3) and it is defined like this, so I think we're safe.
I've also removed the `#include "proj.h"` since I don't think it's necessary here (although please check :).

### DIFF
--- a/meos/include/postgis_ext_defs.in.h
+++ b/meos/include/postgis_ext_defs.in.h
@@ -408,7 +408,10 @@ extern double lwpoint_get_m(const LWPOINT *point);
 extern int lwgeom_has_z(const LWGEOM *geom);
 extern int lwgeom_has_m(const LWGEOM *geom);
 
-#include "proj.h"
+/* PROJ */
+
+struct PJconsts;
+typedef struct PJconsts PJ;
 
 typedef struct LWPROJ
 {


mschoema:
Awesome, if this works and we don't need to include proj.h, this is indeed a welcome change.
However, do we not need to define `struct PJconsts`, is it enough to just declare it?

Diviloper:
Regarding the import, all the experiments I've tried work without it. That said, there are still a few imports of `proj.h` throughout the code (e.g [`meos.c`](https://github.com/MobilityDB/MobilityDB/blob/c5503620caccf4054bc88a854cc57bd522c5a7e0/meos/src/general/meos.c#L42), [`liblwgeom.h`](https://github.com/MobilityDB/MobilityDB/blob/c5503620caccf4054bc88a854cc57bd522c5a7e0/postgis/liblwgeom/liblwgeom.h#L38), and [`type_srid.c`](https://github.com/MobilityDB/MobilityDB/blob/c5503620caccf4054bc88a854cc57bd522c5a7e0/meos/src/point/type_srid.c#L38)), so I don't really know how much of a difference the change makes in practice.

Regarding the `PJconsts`, I think it is enough like this. That empty definition is extracted from the [`proj.h` header](https://github.com/OSGeo/PROJ/blob/498d409835bb3e3d0f0a2debb3303b1bf89f9ca6/src/proj.h#L224), and it is [stated in the documentation](https://proj.org/en/9.3/development/reference/datatypes.html#c.PJ) that `As a user of the PROJ library you are only exposed to pointers to this object and the contents is hidden behind the public API`.
The actual details are available in the [`proj_internal.h` header](https://github.com/OSGeo/PROJ/blob/498d409835bb3e3d0f0a2debb3303b1bf89f9ca6/src/proj_internal.h#L456).
I've been able to compile PyMEOS CFFI with these changes successfully (without runtime issues in PyMEOS), and normally the compilation is very sensitive to undefined or partially defined symbols.

mschoema (approved):

PR #447, opened 2024-02-23
Hide Postgis and and Postgres functions in MEOS library

mschoema:
* This PR adds CMake parameters to hide the Postgis and Postgres functions used by the MEOS library.
* The only exported Postgis function is `geo_get_srid`, which is a MEOS function wrapping the Postgis `gserialized_get_srid` function.

### DIFF
--- a/meos/include/postgis_ext_defs.in.h
+++ b/meos/include/postgis_ext_defs.in.h
@@ -393,20 +393,9 @@ typedef struct
 }
 LWTIN;
 
-extern LWPOINT *lwpoint_make(int32_t srid, int hasz, int hasm, const POINT4D *p);
+/* Functions */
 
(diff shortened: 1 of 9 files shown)


mschoema:
@Diviloper This PR will probably require some slight changes in PyMEOS.

1. `lwgeom_get_srid` is now `geo_get_srid` and receives a `GSERIALIZED *` instead of a `LWGEOM *`.
2. The other Postgis functions have been remove. I don't think there were of any use in PyMEOS, but let me know if I'm wrong. We can think of wrapping them in MEOS and adding them back (like for `geo_get_srid`) if needed.

estebanzimanyi (approved):

Diviloper:
I've checked and the only changed needed was in `pymeos_cffi`, and it's already been taken care of.

PR #449, opened 2024-02-23
Fix temporal_bbox_slice

estebanzimanyi:


mobilitydb/src/general/temporal_index.c
```diff
@@ -60,7 +61,8 @@ Temporal_gist_compress(PG_FUNCTION_ARGS)
   {
     GISTENTRY *retval = palloc(sizeof(GISTENTRY));
     Span *s = palloc(sizeof(Span));
-    temporal_bbox_slice(entry->key, s);
+    Temporal *temp = temporal_slice(entry->key);
+    temporal_set_bbox(temp, s);
```
mschoema, line 65: I know that `Temporal_gist_compress` should only be called for types whose bounding box is a span, but I would still suggest to use `temporal_set_tstzspan` instead of `temporal_set_bbox` here to avoid potential bugs in the future.
The same holds for other similar coccurences.

Generally, `temporal_set_bbox` should only be called with a `bboxunion` as second parameter, for example when testing for overlap between the boxes of two temporal types.
If the dimension of the second parameter is known (span, tbox or stbox), the corresponding function should be called (`temporal_set_tstzspan`, `temporal_set_tbox`, `temporal_set_stbox`, with the last two having an assert to make sure it is only called on types that have these dimensions).

mschoema (approved):

meos/src/point/tpoint.c
```diff
@@ -67,6 +67,32 @@ geo_copy(const GSERIALIZED *g)
...
+      tpointinst_set_stbox((TInstant *) temp, box);
+      break;
+    case TSEQUENCE:
+      tpointseq_set_stbox((TSequence *) temp, box);
+      break;
+    default: /* TSEQUENCESET */
+      tpointseqset_set_stbox((TSequenceSet *) temp, box);
+  }
+  return;
+}
```
mschoema, line 94: Note that the `tpoint` GiST and SP-GiST functions are also used by the `tnpoint` type.
This means that this function can be called by `tnpoint` values, which currently crashes for `TINSTANT` subtypes, because `tpointinst_set_stbox` cannot handle `tnpoint` values.

This function should thus probably be renamed to `tspatial_set_stbox` to denote the fact that it can be used by all types with an `STBox`. The `tspatial_type` function can then be used in the assert instead of `tgeo_type`. (Same with `tpointinst_set_stbox`, `tpointseq_set_stbox` and `tpointseqset_set_stbox`)

mschoema (approved):

PR #454, opened 2024-03-16
Fix mobilitydb oid cache loading

mschoema:
- Split oid cache loading between type and operator cache
- Only load type cache on `CREATE EXTENSION` call
- Use `get_extension_schema` and `GetSysCacheOid2` to avoid calls to calls to `PushOverrideSearchPath`


estebanzimanyi (approved):
Congratulations for this amazing job !

Issue #464, opened 2024-04-02
Incorrect output of some spatial relationship functions


mschoema:
**Describe the bug**

The aDisjoint and aDwithin functions do not work correctly in the tpoint-geo case.

The aDwithin function is testing for intersection between the trajectory of the tgeompoint and the geometry, which is essentially eDwithin. Instead, it should check for containement of the traj in the geometry.

The aDisjoin function does not take into account linear interpolation of the tpoint and just tests for disjointness at the instants.

**To Reproduce**

```SQL
test=# select adisjoint(tgeompoint '[Point(0 0)@2001-01-01, Point(2 0)@2001-01-02]', geometry 'Point(1 0)');
 adisjoint 
-----------
 t
(1 row)

test=# select adwithin(tgeompoint '[Point(0 0)@2001-01-01, Point(2 0)@2001-01-02]', geometry 'Point(1 0)', 0.5);
 adwithin 
----------
 t
(1 row)
```

**Expected behavior**

```SQL
test=# select adisjoint(tgeompoint '[Point(0 0)@2001-01-01, Point(2 0)@2001-01-02]', geometry 'Point(1 0)');
 adisjoint 
-----------
 f
(1 row)

test=# select adwithin(tgeompoint '[Point(0 0)@2001-01-01, Point(2 0)@2001-01-02]', geometry 'Point(1 0)', 0.5);
 adwithin 
----------
 f
(1 row)
```

**Specifications**

```SQL
test=# select mobilitydb_full_version();
                                            mobilitydb_full_version                                            
---------------------------------------------------------------------------------------------------------------
 MobilityDB 1.1.0, PostgreSQL 16.2, PostGIS 3.4.2, GEOS 3.10.2-CAPI-1.16.0, PROJ 8.2.1, JSON-C 0.17, GSL 2.7.1
(1 row)
```

PR #468, opened 2024-04-04
Fix ever/always spatial relationships

estebanzimanyi:
Fix Issue #464

### DIFF
--- a/meos/src/point/tpoint_spatialrels.c
+++ b/meos/src/point/tpoint_spatialrels.c
@@ -395,10 +395,9 @@ edisjoint_tpoint_geo(const Temporal *temp, const GSERIALIZED *gs)
   /* Ensure validity of the arguments */
   if (! ensure_valid_tpoint_geo(temp, gs) || gserialized_is_empty(gs))
     return -1;
-  GSERIALIZED *traj = tpoint_trajectory(temp);
-  bool result = DatumGetBool(geom_covers(PointerGetDatum(gs),
-    PointerGetDatum(traj)));
-  pfree(traj);
+  datum_func2 func = &geom_covers;
+  int result = spatialrel_tpoint_traj_geo(temp, gs, (Datum) NULL,
+    (varfunc) func, 2, INVERT);
   return result ? 0 : 1;
 }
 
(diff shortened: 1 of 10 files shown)


meos/src/point/tpoint_spatialrels.c
```diff
@@ -722,12 +652,13 @@ adwithin_tpoint_geo(const Temporal *temp, const GSERIALIZED *gs, double dist)
       ! ensure_not_negative_datum(Float8GetDatum(dist), T_FLOAT8))
     return -1;
 
+  GSERIALIZED *traj = tpoint_trajectory(temp);
   GSERIALIZED *buffer = geometry_buffer(gs, dist, "");
-  datum_func2 func = get_intersects_fn_gs(temp->flags, gs->gflags);
-  int result = spatialrel_tpoint_traj_geo(temp, buffer, Float8GetDatum(dist),
-    (varfunc) func, 3, INVERT_NO);
+  bool result = DatumGetBool(geom_covers(PointerGetDatum(buffer),
```
mschoema, line 657: We could maybe call `spatialrel_tpoint_traj_geo` with `temp` and `buffer` here.

meos/src/point/tpoint_spatialrels.c
```diff
@@ -370,113 +380,26 @@ acontains_geo_tpoint(const GSERIALIZED *gs, const Temporal *temp)
...
 {
-  return DatumGetInt32(ea_disjoint_tpoint_geo(temp, gs, EVER));
+  /* Ensure validity of the arguments */
+  if (! ensure_valid_tpoint_geo(temp, gs) || gserialized_is_empty(gs))
+    return -1;
+  GSERIALIZED *traj = tpoint_trajectory(temp);
+  bool result = DatumGetBool(geom_covers(PointerGetDatum(gs),
+    PointerGetDatum(traj)));
+  pfree(traj);
+  return result ? 0 : 1;
```
mschoema, line 402: Same here

meos/src/point/tpoint_spatialrels.c
```diff
@@ -367,116 +377,31 @@ acontains_geo_tpoint(const GSERIALIZED *gs, const Temporal *temp)
...
 {
-  return DatumGetInt32(ea_disjoint_tpoint_geo(temp, gs, EVER));
+  /* Ensure validity of the arguments */
+  if (! ensure_not_geodetic(temp->flags) || gserialized_is_empty(gs) ||
+      ! ensure_valid_tpoint_geo(temp, gs))
+    return -1;
+  datum_func2 func = &geom_covers;
+  int result = spatialrel_tpoint_traj_geo(temp, gs, (Datum) NULL,
+    (varfunc) func, 2, INVERT);
+  return result ? 0 : 1;
```
mschoema, line 404: This will return 1 if result = -1, which is not wanted.
We probably want `return result < 0 ? result : (result > 0 ? 0 : 1);`
Should we make macros for this?

mschoema: Something like:
```C
#define INVERT_RESULT(result) (result < 0 ? -1 : (result > 0) ? 0 : 1)
```

Could be another name, this is just the first idea I had.
We just need to make sure we only call it as `INVERT_RESULT(var)` and never as `INVERT_RESULT(f(param))`, otherwise f could be evaluated twice.

I'm fine with avoiding the macro and keeping the expressions duplicated, but we need to make sure they're all the same.

mschoema:
You can mark this PR as 'ready for review' when all the changes have been implemented

mschoema (approved):

PR #469, opened 2024-04-05
Configure pg_config.h using CMake

mschoema:
This solves some compilation issues on arm and ios.


estebanzimanyi (approved):

mschoema:
@Diviloper Could you check if these fixes work on your ARM machine or if anything is missing?

PR #471, opened 2024-04-06
Use datum_func typedefs when possible

mschoema:


estebanzimanyi (approved):
Absolutely wonderful ! Thanks !

PR #472, opened 2024-04-07
Add arm to macOS workflow

Diviloper:
Run the MobilityDB tests on both intel- and arm-based macOS.
Explicitely  specify macos versions (`macos-13` and `macos-14`) instead of `macos-latest` (which now, or at least soon, is the same as `macos-14`, see [GitHub docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories))

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -29,6 +29,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ macos-13, macos-14 ]
+        include:
+          - os: macos-13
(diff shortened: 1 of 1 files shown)


mschoema:
Amazing!
Looking at previous runs of the action using `macos-latest` it looks like the used image was `macos-12` instead of the expected `macos-14`. Manually setting the version thus does the trick.

mschoema (approved):

PR #486, opened 2024-05-06
Fix tsequenceset_make_gaps for non-exit error handling

Diviloper:
Checks the result of `tsequence_make_exp` before moving on.
It was crashing on PyMEOS when, for example, trying to create a `TBoolSeqSet` from gaps with Linear interpolation.
That error was never happening on MEOS because the default error handler just exits the problem (which is not the case in the PyMEOS error handler)


mschoema (approved):

mschoema:
Nice find.
If you want this bugfix to be backported to the stable-1.1 branch, don't hesitate to open another PR on that branch with the same fix.

PR #492, opened 2024-05-15
Replace sprintf with snprintf

estebanzimanyi:


meos/src/general/pg_types.c
```diff
@@ -284,7 +286,7 @@ int4_out(int32 val)
 #if POSTGRESQL_VERSION_NUMBER >= 130000
   mobdb_ltoa(val, result);
 #else
-  sprintf(result, "%d", val);
+  snprintf(result, 12, "%d", val);
```
mschoema, line 289: Shouldn't we use `MAXINT4LEN` here somehow?

meos/src/general/pg_types.c
```diff
@@ -361,7 +363,7 @@ int8_out(int64 val)
   memcpy(result, buf, len);
 #else
   result = palloc(MAXINT8LEN + 1);
-  sprintf(result, "%ld", val);
+  snprintf(result, sizeof(result), "%ld", val);
```
mschoema, line 366: Cannot use `sizeof(result)`, since `result` is a `char *`

meos/src/general/temporal.c
```diff
@@ -677,11 +677,11 @@ mobilitydb_full_version(void)
...
   proj_vers = pji.version;
 #endif
-  const char* geos_vers = GEOSversion();
-  const char* json_c_vers = json_c_version();
+  const char *geos_vers = GEOSversion();
+  const char *json_c_vers = json_c_version();
 
-  char *result = palloc(sizeof(char) * MOBDB_VERSION_STR_MAX_LEN);
-  int len = snprintf(result, MOBDB_VERSION_STR_MAX_LEN,
+  char *result = palloc(256);
```
mschoema, line 683: This should also use `MOBDB_VERSION_STR_MAXLEN`

mschoema (changes requested):

mobilitydb/src/point/tpoint.c
```diff
@@ -67,8 +67,8 @@
 static void
 pg_error(const char *fmt, va_list ap)
 {
-  char errmsg[PGC_ERRMSG_MAXLEN];
-  vsnprintf (errmsg, sizeof(errmsg), fmt, ap);
+  char errmsg[PGC_ERRMSG_MAXLEN + 1];
+  vsnprintf (errmsg, PGC_ERRMSG_MAXLEN, fmt, ap);
   errmsg[PGC_ERRMSG_MAXLEN]='\0';
```
mschoema, line 72: It would be better to keep `char errmsg[PGC_ERRMSG_MAXLEN];` and just remove the line with `errmsg[PGC_ERRMSG_MAXLEN]='\0';`, since vsnprintf already adds the null terminator

mobilitydb/src/point/tpoint.c
```diff
@@ -80,8 +80,8 @@ pg_error(const char *fmt, va_list ap)
 static void
 pg_notice(const char *fmt, va_list ap)
 {
-  char errmsg[PGC_ERRMSG_MAXLEN];
-  vsnprintf (errmsg, sizeof(errmsg), fmt, ap);
+  char errmsg[PGC_ERRMSG_MAXLEN + 1];
+  vsnprintf (errmsg, PGC_ERRMSG_MAXLEN, fmt, ap);
   errmsg[PGC_ERRMSG_MAXLEN]='\0';
```
mschoema, line 85: same here

mschoema (changes requested):

mschoema (approved):

PR #496, opened 2024-05-17
Add optional interp parameter to tsample

estebanzimanyi:


meos/src/general/temporal_analytics.c
```diff
@@ -587,8 +587,8 @@ tsequenceset_tsample(const TSequenceSet *ss, const Interval *duration,
     ninsts += tsequence_tsample_iter(TSEQUENCESET_SEQ_N(ss, i), lower_bucket,
       upper_bucket, tunits, &instants[ninsts]);
   }
-  return tsequence_make_free(instants, ninsts, true, true, DISCRETE,
-    NORMALIZE_NO);
+  return tsequence_make_free(instants, ninsts, true, true, interp,
+    NORMALIZE);
```
mschoema, line 591: For `tsequenceset_sample` we cannot just create a sequence with the desired interpolation.
It works if we have `interp == DISCRETE`, but for `LINEAR` or `STEP`, we need to take the gaps of the input sequence set into account.

mschoema:
Another thing to decide is: can we arbitrarily chose the output interpolation?
For example, if the input is linear, do we allow the output to be stepwise or does that not make sense?
If would say we allow it and its up to the user to use the right interpolation.

mschoema (approved):

PR #498, opened 2024-05-20
Improve MFJSON output

estebanzimanyi:


mschoema (dismissed):

mschoema:
Can you regenerate the manual after doing a rebase of the last PR accepted on master?
Actually, should the pdf of the manual be pregenerated in this repo?
It seems to me that this is not required as we are pregenerating the doc with github pages anyways.
And as we can see in this PR, github doesn't handle pdf diffs well, so it would be best to remove it.

mschoema (approved):

PR #512, opened 2024-07-08
Improve GSL initialization and global variable management

estebanzimanyi:
* Distinguish global variables and constants
* Improve GSL initialization


meos/src/general/meos.c
```diff
@@ -474,8 +477,60 @@ meos_finalize(void)
   meos_finalize_timezone();
   /* Finalize PROJ */
   proj_context_destroy(_PJ_CONTEXT);
+  /* Finalize GSL */
+  gsl_rng_free(_GENERATION_RNG);
+  gsl_rng_free(_AGGREGATION_RNG);
```
mschoema, line 482: This should maybe go into a `gsl_finalize()` function that also sets `_GSL_INITIALIZED` to `false`.

meos/include/meos.h
```diff
@@ -353,6 +356,9 @@ extern bool meos_set_intervalstyle(char *newval, int extra);
 extern char *meos_get_datestyle(void);
 extern char *meos_get_intervalstyle(void);
 
+extern gsl_rng *gsl_get_generation_rng(void);
+extern gsl_rng *gsl_get_aggregation_rng(void);
```
mschoema, line 360: Should these functions be exposed in `meos.h`?

estebanzimanyi: These functions are used for example in
https://github.com/estebanzimanyi/MobilityDB/blob/global/meos/src/point/tpoint_datagen.c#L258
https://github.com/estebanzimanyi/MobilityDB/blob/global/meos/src/general/skiplist.c#L124

meos/include/meos.h
```diff
@@ -38,6 +38,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
+/* GSL */
+#include <gsl/gsl_rng.h>
+#include <gsl/gsl_randist.h>
```
mschoema, line 43: Could you also remove the includes that were added in `meos.h`?
After that everything looks god so I'll accept and merge the PR.

mschoema (approved):

PR #521, opened 2024-07-17
Add function stboxesFromSegs

estebanzimanyi:


meos/src/point/tpoint_tile.c
```diff
@@ -1029,7 +1029,7 @@ tpoint_space_time_split_init(Temporal *temp, float xsize, float ysize,
   /* Create function state */
   STboxGridState *state = stbox_tile_state_make(temp, &bounds, xsize, ysize,
     zsize, tunits, pt, torigin, border_inc);
-  int ndims = 2 + (hasz ? 1 : 0) + (state->tunits ? 1 : 0);
+  int ndims = 2 + hasz ? 1 : 0 + state->tunits ? 1 : 0;
```
mschoema, line 1032: You might have to rebase with master.
This change should not be here.

meos/include/meos.h
```diff
@@ -1948,7 +1952,7 @@ extern bool tpoint_is_simple(const Temporal *temp);
 extern double tpoint_length(const Temporal *temp);
 extern Temporal *tpoint_speed(const Temporal *temp);
 extern int tpoint_srid(const Temporal *temp);
-extern STBox *tpoint_stboxes(const Temporal *temp, int *count);
+extern STBox *tpoint_stboxes_from_segs(const Temporal *temp, int max_count, int *count);
```
mschoema, line 1955: I thought that the suffix `_from_segs` would be used together with the `segs_per_box` parameter. I don't think it fits well with a function where you specify the max number of boxes. The old name `tpoint_stboxes` seems to be better in this case.

mobilitydb/sql/general/007_spanset.in.sql
```diff
@@ -858,6 +858,29 @@ CREATE FUNCTION round(floatspanset, integer DEFAULT 0)
...
+  AS 'MODULE_PATHNAME', 'Spanset_spans_n'
+  LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION spansN(datespanset, int DEFAULT 0)
+  RETURNS datespan[]
+  AS 'MODULE_PATHNAME', 'Spanset_spans_n'
+  LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE FUNCTION spansN(tstzspanset, int DEFAULT 0)
+  RETURNS tstzspan[]
+  AS 'MODULE_PATHNAME', 'Spanset_spans_n'
+  LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
```
mschoema, line 882: The name `spansN` is too close to `spanN` while doing a widely different thing. We need a different name for this function

mschoema (approved):

PR #532, opened 2024-07-20
Finding leaks

estebanzimanyi:


meos/src/point/tpoint_tile.c
```diff
@@ -1143,6 +1143,7 @@ tpoint_space_time_split(Temporal *temp, float xsize, float ysize, float zsize,
       times[i] = DatumGetTimestampTz(box.period.lower);
     result[i++] = atstbox;
   }
+  pfree(state);
```
mschoema, line 1146: This is not needed, we only exit the loop after freeing the state.

meos/src/point/tpoint_spatialfuncs.c
```diff
@@ -1176,7 +1176,7 @@ geopoint_make(double x, double y, double z, bool hasz, bool geodetic,
     lwpoint_make3dz(srid, x, y, z) : lwpoint_make2d(srid, x, y);
   FLAGS_SET_GEODETIC(point->flags, geodetic);
   GSERIALIZED *result = geo_serialize((LWGEOM *) point);
-  lwpoint_free(point);
+  lwgeom_free(lwpoint_as_lwgeom(point));
```
mschoema, line 1179: Not sure why this is needed instead of just `twpoint_free`

meos/src/general/set.c
```diff
@@ -985,8 +985,17 @@ set_make_free(Datum *values, int count, meosType basetype, bool order)
...
+  {
     result = set_make_exp(values, count, count, basetype, order);
-  pfree(values);
+#if MEOS
+    if (! basetype_byvalue(basetype))
+    {
+      for (int i = 0; i < count; i++)
+        pfree(DatumGetPointer(values[i]));
+    }
+#endif /* MEOS */
```
mschoema, line 996: We should not have an `#if MEOS` here. Either the function always frees the values or it never does, but having a different behaviour in MEOS and MobilityDB will be too confusing.

meos/src/general/span_aggfuncs_meos.c
```diff
@@ -82,7 +82,7 @@ spanset_append_span(SpanSet *ss, const Span *span, bool expand)
   SpanSet *result = spanset_make_exp(spans, ss->count + 1, maxcount,
     NORMALIZE_NO, ORDER);
   pfree(spans);
-  // pfree(ss);
+  pfree(ss);
```
mschoema, line 85: This is related to my comment from #531 and is a good change, but we need to specify in the doc of the function that we potentially free the input spanset, so the user always have to call the function as:
```C
ss = spanset_append_span(ss, ...);
```

Also this means we need to revert the changes made to `examples/intspan_agg.c` and `examples/floatspanset_agg.c`

meos/src/general/span_aggfuncs_meos.c
```diff
@@ -128,7 +128,7 @@ spanset_append_spanset(SpanSet *ss1, const SpanSet *ss2, bool expand)
 
   SpanSet *result = spanset_make_exp(spans, count, maxcount, NORMALIZE_NO,
     ORDER);
-  pfree(spans);
+  pfree(spans); pfree(ss1);
```
mschoema, line 131: Same comment as above

mschoema (changes requested):

meos/src/point/tpoint_tile.c
mschoema: Did you check if this `pfree` is needed or not?
I believe we always free the state before breaking out of the loop, so this new `pfree` should not be needed.

mschoema (approved):

PR #543, opened 2024-07-27
Add output function for (multi)ranges

estebanzimanyi:


meos/src/general/pg_types.c
```diff
@@ -70,6 +72,8 @@
   extern Datum timestamp_out(PG_FUNCTION_ARGS);
   extern Datum timestamptz_out(PG_FUNCTION_ARGS);
   extern Datum interval_out(PG_FUNCTION_ARGS);
+  extern Datum range_out(PG_FUNCTION_ARGS);
+  extern Datum multirange_out(PG_FUNCTION_ARGS);
```
mschoema, line 76: This should be removed too, since it's now in `mobilitydb/src.general/type_util.c`

mschoema (changes requested):

mschoema (approved):

PR #546, opened 2024-07-31
Make some pointers in the parameters constant

### LINKED ISSUE #545: Make non-mutated pointers const
**Current Issue**
Right now, there are many functions which parameters pointers could be const. Example: https://github.com/MobilityDB/MobilityDB/blob/2784b78898ced300a2d7af8536c7434595b00eb4/meos/src/general/temporal_tile.c#L1154

**Why is it relevant**
1. Readability from the code as such, since if you see the `const`, you know that by convention that variable won't be changed.
2. Wrapper improvement. The code needed to create the wrapper in Rust will be cleaner, since in Rust, if the pointer type it's specified as const, the type generated by bindgen (the tool that takes C headers and generates the Rust equivalents), is `const *T`, whereas if `const` it's not specified, it outputs `mut *T`, this means that for functions where I wouldn't need to pass a mutable reference, I still need to. 

**Describe the solution you'd like**
For every pointer type received as a parameter, if it's not modified, add a `const` modifier.

Davichet-e:
## Make some pointers in the parameters constant
Fixes #545.

### Testing
I built MEOS and it completed succesfully

### Unknowns
- Will this break any wrapper?
  - Is there any Regex used in ~PyMEOS~ (confirmed with author) or JMEOS that would make this change a breaking change?

### DIFF
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -1464,8 +1464,8 @@ extern Temporal **tpointarr_round(const Temporal **temp, int count, int maxdd);
  * Modification functions for temporal types
  *****************************************************************************/
 
-extern Temporal *temporal_append_tinstant(const Temporal *temp, const TInstant *inst, double maxdist, const Interval *maxt, bool expand);
-extern Temporal *temporal_append_tsequence(const Temporal *temp, const TSequence *seq, bool expand);
+extern Temporal *temporal_append_tinstant(Temporal *temp, const TInstant *inst, double maxdist, const Interval *maxt, bool expand);
+extern Temporal *temporal_append_tsequence(Temporal *temp, const TSequence *seq, bool expand);
 extern Temporal *temporal_delete_tstzspan(const Temporal *temp, const Span *s, bool connect);
 extern Temporal *temporal_delete_tstzspanset(const Temporal *temp, const SpanSet *ss, bool connect);
 extern Temporal *temporal_delete_timestamptz(const Temporal *temp, TimestampTz t, bool connect);
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -328,7 +328,7 @@ typedef enum
 
 } errorCode;
 
-extern void meos_error(int errlevel, int errcode, char *format, ...);
+extern void meos_error(int errlevel, int errcode, const char *format, ...);
 
 /* Set or read error level */
 
@@ -342,7 +342,7 @@ extern int meos_errno_reset(void);
  *****************************************************************************/
 
 /* Definition of error handler function */
-typedef void (*error_handler_fn)(int, int, char *);
+typedef void (*error_handler_fn)(int, int, const char *);
 
 extern void meos_initialize_timezone(const char *name);
 extern void meos_initialize_error_handler(error_handler_fn err_handler);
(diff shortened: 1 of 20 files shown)


meos/src/general/temporal.c
```diff
@@ -1502,7 +1502,7 @@ temporal_compact(const Temporal *temp)
  * @param[out] count Number of instants or sequences kept
  */
 void
-temporal_restart(Temporal *temp, int count)
+temporal_restart(const Temporal *temp, int count)
```
Davichet-e, line 1505: Just realize this shouldn't be const

Davichet-e: I will wait to fix it in case there are other comments to address

Diviloper:
Regarding PyMEOS, this shouldn't affect the wrapper at all, so no worries in that regard.

Davichet-e:
I just realised I didn't change the declarations in the meos.h file, working on it. Will also add more consts

meos/include/meos.h
```diff
@@ -1464,15 +1464,15 @@ extern Temporal **tpointarr_round(const Temporal **temp, int count, int maxdd);
  * Modification functions for temporal types
  *****************************************************************************/
 
-extern Temporal *temporal_append_tinstant(Temporal *temp, const TInstant *inst, double maxdist, Interval *maxt, bool expand);
-extern Temporal *temporal_append_tsequence(Temporal *temp, const TSequence *seq, bool expand);
+extern Temporal *temporal_append_tinstant(const Temporal *temp, const TInstant *inst, double maxdist, const Interval *maxt, bool expand);
+extern Temporal *temporal_append_tsequence(const Temporal *temp, const TSequence *seq, bool expand);
```
mschoema, line 1468: In these functions the input temp can be free'd

mschoema (changes requested):

Davichet-e:
Forgot to check the examples, will do now

meos/src/general/temporal_modif.c
```diff
@@ -2117,7 +2117,7 @@ temporal_append_tsequence(const Temporal *temp, const TSequence *seq, bool expan
       return result;
     }
     case TSEQUENCE:
-      return (Temporal *) tsequence_append_tsequence((TSequence *) temp, seq,
+      return (Temporal *) tsequence_append_tsequence((const TSequence *) temp, seq,
```
mschoema, line 2120: Here we should actually free `temp` before returning, since we are returning a new pointer.
This function is made to be used like this:
```c
temp = temporal_append_tsequence(temp, ...)
```
So we need to free the input if we return a newly allocated value

mschoema: The same goes for the `TINSTANT` case above actually

Davichet-e:
Checked the examples, no warning were generated when compiling them

Davichet-e:
Sorry for the messy commit message, I forgot if you put a string between ``, it executes the command with that name. It shouldn't matter since I guess we will rebase and will just squash all of the commits into one.

meos/src/general/temporal_modif.c
```diff
@@ -2110,15 +2110,20 @@ temporal_append_tsequence(Temporal *temp, const TSequence *seq, bool expand)
...
       pfree(seq1);
+      pfree(temp);
       return result;
     }
     case TSEQUENCE:
-      return (Temporal *) tsequence_append_tsequence((TSequence *) temp, seq,
+    {
+      Temporal *result = tsequence_append_tsequence((const TSequence *) temp, seq,
         expand);
+      pfree(temp);
```
Davichet-e, line 2124: The tests are now failing, I guess it's because of these added `free`s. Since these changes are not part of the objective of this PR, I'll remove them and create another PR if needed

meos/src/general/temporal_modif.c
Davichet-e: Tests failed when I introduced this proposed change. As it's not part of the scope of this PR, will not include them here but rather in other PR if needed

mschoema (approved):

PR #548, opened 2024-08-01
Improve spanset_spans by taking account size of holes

estebanzimanyi:


mobilitydb/test/general/expected/007_spanset.test.out
```diff
@@ -732,13 +732,13 @@ SELECT spans(intspanset '{[1, 2), [3, 4), [5, 6), [7, 8), [9, 10)}', 1);
 SELECT spans(intspanset '{[1, 2), [3, 4), [5, 6), [7, 8), [9, 10)}', 2);
         spans         
 ----------------------
- {"[1, 6)","[7, 10)"}
+ {"[1, 8)","[9, 10)"}
```
Davichet-e, line 735: If I understand the `spans` function correctly, it returns the spans consisting a `spanset`. In this case, shouldn't return "[1, 2)", "[3, 4)"?

mschoema: Without the second parameter, the function returns all the spans from the spanset. The second paramter is used to limit the amount of spans returned, but the spans should still cover all the values inside the spanset. The return values are thus formed by merging spans until we get to the desired amount. This is used in the `mgist` and `mspgist` indices.

Currently we don't have a functiont that just returns the first N spans of a spanset, but we do have a function `spanN`, which returns the N'th span of the spanset.

meos/src/general/spanset.c
```diff
@@ -1866,26 +1906,25 @@ spanset_spans(const SpanSet *ss, int max_count, int *count)
...
+    /* Sort the holes to fill the original spanset in increasing value/time */
+    spanarr_sort(holes, nfills);
+    SpanSet *tofill = spanset_make_exp(holes, nfills, nfills, NORMALIZE_NO,
+      ORDER_NO);
+    /* Resulting spanset with the holes filed */
+    SpanSet *res = union_spanset_spanset(ss, tofill);
+    assert(res->count == max_count);
+    /* Construct the resulting array of spans */
+    for (int i = 0; i < res->count; i++)
+      memcpy(&result[i], SPANSET_SP_N(res, i), sizeof(Span));
```
mschoema, line 1927: We might be missing a few `pfree` calls here.

mobilitydb/test/general/expected/007_spanset.test.out
Davichet-e: What do you think of creating an additional function `spanset_spans_merge` that has this behaviour of reducing the number of spans of the spanset to the requested number, and having the `spans` function just return the expected, the spans of the spanset.

mschoema: I don't disagree with splitting this into 2 functions in the MEOS API, since the functionality is different.
What do you think @estebanzimanyi

estebanzimanyi: I would agree on having two functions. This also applies to the functions tnumber_tboxes and tpoint_stboxes. I would propose to accept all the ongoing PRs and then I can work on splitting these functions.

mschoema (approved):

Issue #549, opened 2024-08-02
Renaming of tile functions: Breaking changes


estebanzimanyi:
### Description

In MobilityDB and in MEOS the names of the tiling functions have been uniformized. While the splitting functions distinguish between valueSplit, valueTimeSplit, spaceSplit, and spaceTimeSplit, the corresponding tile functions were simply named tileList and tile. The new names of the functions make explicit the dimensions requested for the tiles.

### What is changing

In SQL
* Function tileList for tboxes renamed to valueTimeTiles
* Function tile for tboxes renamed to valueTimeTile
* Function tileList for stboxes renamed to spaceTimeTiles
* Functions tile for stboxes renamed to spaceTimeTile

In MEOS
* Function tbox_tile_list renamed to tbox_value_time_tiles
* Function tbox_tile renamed to tbox_value_time_tile
* Function stbox_tile_list renamed to stbox_space_time_tiles
* Functions stbox_tile renamed to stbox_space_time_tile

### Additional functions

In SQL
* spaceTiles
* spaceTile

In MEOS
* stbox_space_tiles
* stbox_space_tile

PR #550, opened 2024-08-02
Improve naming for tile functions

estebanzimanyi:
* Improve spanset_spans functions by taking into account the size of the holes
* Improve documentation
Implements the changes stated in Issue #549


mschoema:
This PR contains the changes from #548, which should not be here

mschoema (approved):

meos/src/general/span.c
```diff
@@ -1972,7 +1972,7 @@ span_ne(const Span *s1, const Span *s2)
  * @ingroup meos_internal_setspan_comp
  * @brief Return -1, 0, or 1 depending on whether the first span is less than,
  * equal, or greater than the second one
- * @param[in] s1,s2 Sets
+ * @param[in] s1,s2 Spans
```
mschoema, line 1975: I think this might still cause conflicts. Either you change this here, or I accept this PR and you can change this in the other one. The other one still needs a possible commit for some `free` calls

estebanzimanyi: This function is for the qsort algorithm and thus it is NOT supposed to be used as an external function.

mschoema (approved):

PR #551, opened 2024-08-02
Change return type of temporal_*_inc to bool

Davichet-e:
## Change return type of temporal_*_inc to bool
All of the functions that return whether the bounds are inclusive or not are returning bool, except the temporal ones. This PR brings consistency by changing the type to bool to conform to the other functions.

### Why?
- Consistency
- In Rust, even when bool is just a typedef for an int, when generating the CFFI interface, it makes Rust think it will return a fully flexed int instead of a boolean

### Unknowns
Will this break JMEOS or any of the wrappers with static typed languages?


mschoema:
This will not work correctly.
The reason why it returns an `int` is because the function can produce an error.
Thus, 0/1 means false/true, while -1 means Error.
With your current changes, Error and false cannot be distinguished.

Davichet-e:
That's how it works for all of the other _lower_inc and _upper_inc functions. I would argue the same family of functions should have consistent API

mschoema:
Ah indeed, the other functions do this as well.
However, I would argue that this is wrong in the other functions and they should return -1 as well.
I agree that this is not ideal design, but otherwise the error could be missed.
What are your thoughts related to the error handling?
@estebanzimanyi what do you think?

Davichet-e:
Taking into account the current meos codebase, there're many places where we return the result without having a distinct value for the error (see all the comparison functions). Personally, for the accessor functions where there is no processing whatsoever, I don't think having a distinct value to mark the input parameters being null will substantially improve error handling of the codebase. I therefore think we should just return bool.

The other option is never returning bool in any predicate function since pointers can be NULL (more than 50 functions would change)

estebanzimanyi:
I would agree to change the two int -> boolean for now for having a coherent codebase.
Then we can have a thorough reflection on how to manage all the boolean returning functions.

mschoema:
Ok, that works.
Let's maybe create an Issue to reference this so we don't forget.

mschoema (approved):

PR #553, opened 2024-08-03
Modify temporal types structs

Davichet-e:
### Current status
The current way in which MEOS supports inheritance is by having a common set of fields in each child class. Namely, every subclass of `Temporal` (`TInstant`, `TSequence` and `TSequenceSet`), has in the top the struct the same fields of `Temporal`. This way, we are able to cast between pointers of `Temporal` and subtypes easily. 

### Proposed change
The current approach duplicates the information already present in `Temporal` in each subclass, which couples the implementation. Therefore we propose leveraging the offset behaviour of C, where the first field has offset 0, which means a pointer to the new temporal field is also a pointer to the whole object as such. This effectively means that the memory structure is absolutely the same between the 2 approaches.

### Trade-offs
- Advantages:
  - Avoid duplicating data, which effectively decouples the implementation and would allow for easier refactorings since instead of having to change the field in 4 different places, you would only need to change it in 1.
  - Compartmentalization: We make more explicit what fields are exclusive to the subtype and which of the supertype
  - Simplicity: It's easier to reason about the hierarchy and teach about it as now the relations are made explicit. Also since it's the [standard way](https://stackoverflow.com/questions/415452/object-orientation-in-c) to have inheritance/polymorphism in C, it's easier for tenured C programmers to understand.

- Disadvantages:
  - More verbose, as now we have to access first the temporal field (memory wise nothing changed, it's just syntax change)
  - Big change: As it's a building block of the library, many files had to change, which introduces a risk of a typo that could be easily overlooked in the review. Also the wrappers authors would need to make sure that this change doesn't affect them (in theory it shouldn't)

### Testing
I've tested it building both MEOS and MobilityDB, and running some of the examples under `meos/examples`

### Additional information
Additional information regarding this new approach to inheritance can be seen in this video: https://www.youtube.com/watch?v=443UNeGrFoM&t=4275s

### DIFF
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -197,10 +197,7 @@ typedef struct
  */
 typedef struct
 {
-  int32 vl_len_;        /**< Varlena header (do not touch directly!) */
-  uint8 temptype;       /**< Temporal type */
-  uint8 subtype;        /**< Temporal subtype */
-  int16 flags;          /**< Flags */
+  Temporal temporal;
   TimestampTz t;        /**< Timestamp (8 bytes) */
   Datum value;          /**< Base value for types passed by value,
                              first 8 bytes of the base value for values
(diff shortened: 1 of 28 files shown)


mschoema:
I agree that this is the ideal way to go when really working with OOP concepts, for example when splitting parent and child classes in separate header files.
However, in our case, the header of these structs will likely not change soon and, even if they do, they are defined right after each other in 1 file.
I thus don't think that this change will bring much benefit and the additional verbosity is not desired.

Davichet-e:
Fair enough, I agree that the benefit is not big, and if at any point this new architecture is regarded as desirable, we can always revive the PR. Closing it.

PR #555, opened 2024-08-05
Separate boxes and boxesMerge functionality

estebanzimanyi:


meos/include/meos.h
```diff
@@ -652,8 +652,9 @@ extern Span *intspan_shift_scale(const Span *s, int shift, int width, bool hassh
 extern SpanSet *intspanset_shift_scale(const SpanSet *ss, int shift, int width, bool hasshift, bool haswidth);
 extern GSERIALIZED *point_transform(const GSERIALIZED *gs, int32 srid);
 extern GSERIALIZED *point_transform_pipeline(const GSERIALIZED *gs, const char *pipelinestr, int32 srid, bool is_forward);
-extern Span *set_spans(const Set *s, int max_count, int *count);
-extern Span *spanset_spans(const SpanSet *ss, int max_count, int *count);
+extern Span *set_spans(const Set *s);
+extern Span *set_spans_merge(const Set *s, int max_count, int *count);
+extern Span *spanset_spans_merge(const SpanSet *ss, int max_count, int *count);
```
mschoema, line 657: Shouldn't `set_spans` also return a count? And `spanset_spans` seems to be missing.

mschoema (changes requested):

meos/include/meos.h
estebanzimanyi: All functions for which the number of values is known in advance do not return the value, e.g.,
https://github.com/estebanzimanyi/MobilityDB/blob/boxes_merge/meos/src/general/spanset.c#L1890

meos/include/meos_internal.h
```diff
@@ -368,7 +368,8 @@ extern SpanSet *numspanset_shift_scale(const SpanSet *ss, Datum shift, Datum wid
 extern Set *set_compact(const Set *s);
 extern void span_expand(const Span *s1, Span *s2);
 extern SpanSet *spanset_compact(const SpanSet *ss);
-extern Span *spanset_spans(const SpanSet *ss, int max_count, int *count);
+extern Span *spanset_spans(const SpanSet *ss);
+extern Span *spanset_spans_merge(const SpanSet *ss, int max_count, int *count);
```
mschoema, line 372: These should be removed as they are now in `meos.h`

meos/include/meos.h
estebanzimanyi: This is different because for them there is a formula that depends, e.g., on whether the tsequence has 1 or 2 instants and the interpolation.

mschoema: Fair enough

mschoema (approved):

PR #556, opened 2024-08-05
Rename boxesMerge -> boxesN

estebanzimanyi:


estebanzimanyi:
I will send a more complete PR with this approach

Issue #405, opened 2023-12-06
In-memory indexes for the MEOS library


estebanzimanyi:
MEOS is the underlying C library driving MobilityDB.
https://libmeos.org/
MEOS corresponds to the C library GEOS for trajectory management.
https://libgeos.org/

Bindings for MEOS in several programming languages are being developped
* Python  https://github.com/MobilityDB/PyMEOS
* Java: https://github.com/MobilityDB/JMEOS
* Javascript: https://github.com/MobilityDB/MEOS.js
* C#: https://github.com/MobilityDB/MEOS.NET

MEOS currently does not support in-memory indexes. On the other hand GEOS supports STRTree indexes
https://libgeos.org/usage/c_api/

We are looking for contributors that can help us to develop this feature. The solution should follow a similar approach as the one for GEOS extending the STRTree indexes with a time dimension.

estebanzimanyi:
The MEOS in-memory quad-tree and k-d tree indexes requested here are implemented and merged. A generic space-partitioning engine (https://github.com/MobilityDB/MobilityDB/pull/1615) covers integer, big-integer, float, date and timestamptz spans and temporal boxes; it is extended to spatiotemporal boxes (https://github.com/MobilityDB/MobilityDB/pull/1618) and to the temporal pgpointcloud boxes (https://github.com/MobilityDB/MobilityDB/pull/1639), with a nearest-neighbour cursor (https://github.com/MobilityDB/MobilityDB/pull/1634) and multi-entry variants. Both tree kinds and the overlaps/contains/contained-by searches are validated against a brute-force oracle. They reuse the same node geometry as the SP-GiST operator classes, which already cover the PostgreSQL side including the temporal pgpointcloud types. Closing as delivered.

PR #563, opened 2024-08-15
[Issue #405] RTree in memory index

Matematikoi:
Implementation of an RTree in memory index to solve issue [405](https://github.com/MobilityDB/MobilityDB/issues/405).

The original code has been taken from Josh Baker [rtree implementation](https://github.com/tidwall/rtree.c) and strip down to a more simple RTree. 

This first version is only working with intersection but future versions should be able to work with any supported function that is already implemented for GIST. 

The basic unit of inserting and querying is an STBox although the code has been done in a way that it allows an easy transition to other structures likes spans or boxes. 

When querying the RTree returns an ID that identifies the STBox that was inserted.


meos/include/meos.h
```diff
@@ -1226,6 +1246,15 @@ extern TBox *intersection_tbox_tbox(const TBox *box1, const TBox *box2);
 extern STBox *union_stbox_stbox(const STBox *box1, const STBox *box2, bool strict);
 extern STBox *intersection_stbox_stbox(const STBox *box1, const STBox *box2);
 
+/*****************************************************************************
+ * RTree functions
+ *****************************************************************************/
+
+extern RTree * rtree_create (int basetype);
```
Matematikoi, line 1253: Here I don't know what to do with the `meosType`. I consider is important to have this definitions in meos.h and not in meos_internal.h, but meosType is in meos_catalog.h. 

I went with this solution but feel free to recommend something else.

meos/src/point/tpoint_rtree.c
```diff
@@ -0,0 +1,663 @@
...
+    return;
+  }
+  if (node -> count == MAXITEMS) {
+    * split = true;
+    return;
+  }
+  RTreeNode * right;
+  node_split(rtree, node -> nodes[insertion_node], & node -> boxes[insertion_node], & right);
+  node -> boxes[insertion_node] = * node_box_calculate(node -> nodes[insertion_node]);
+  node -> boxes[node -> count] = * node_box_calculate(right);
```
mschoema, line 454: You probaby want `node_box_calculate` to return a stack-allocated `STBox` instead of a pointed to a heap-allocated box. Indeed, currently you are copying its value and you are never freeing the returned box so you probably don't want to allocate it in the first place.

meos/src/point/tpoint_rtree.c
```diff
@@ -0,0 +1,663 @@
...
+ */
+static double
+box_unioned_area(const STBox * box,
+  const STBox * other_box, RTree * rtree) {
+  STBox * union_box = palloc(sizeof(STBox));
+  memcpy(union_box, box, sizeof(STBox));
+  stbox_expand(other_box, union_box);
+
+  double answer = box_area(union_box, rtree);
+  free(union_box);
```
mschoema, line 166: Here you don't need to allocate and free memory, just create a new `STBox` by copy and pass its reference to the other functions.

meos/include/meos.h
mschoema: I think we can probably just include `meos_catalog.h` in `meos.h`, but let me think about it.

mschoema (changes requested):

meos/src/point/tpoint_rtree.c
Matematikoi: done!

meos/src/point/tpoint_rtree.c
Matematikoi: done!

meos/include/meos.h
Matematikoi: My problem with importing `meos_catalog.h` in `meos.h`  is definitions like `tempSubtype` which are done in `meos.h` are used in `meos_catalog.h`.

As far as I understand `meos.h` is the base for the rest of the definitions. So importing `meos_catalog.h` sounds weird. I was thinking in moving the definition of  `meosType` to `meos.h` but that should probably be another PR. WDYT?

mschoema: @estebanzimanyi Any thoughts on this?
We're trying to see how to define a function in `meos.h` which takes a `meosType` as input.

mschoema (approved):

meos/include/meos.h
mschoema: For the moment let's leave it like this and we will see how to improve it later

PR #576, opened 2024-08-27
[Issue #405] In memory index - obscure basetype to end user

Matematikoi:
Previously instead of using meosType an int was being used. This new approach allows us to use meosType.

count for the Rtree has been deprecated since it is not being used at the moment.

Replaces rtree_create() for rtree_stbox_create().


mschoema:
I pushed a commit that avoids the use of the `RTreeMetadata` structure.
The second commit is just there to remove the `#if MEOS`, since it's not needed anymore thanks to the next PR.

The rest of the changes seem good, so if you agree with my changes I'll merge the PR.

mschoema (approved):

meos/include/point/tpoint_rtree.h
```diff
@@ -74,14 +74,23 @@ typedef struct RTreeNode{
...
+/**
+ * Rtree in memory index basic structure.
+ *
+ * It works based on STBox. The spliting criteria is based on the largest axis.
+ * The inserting criteria is based on least enlarging square.
+ *
+ * The get axis function makes it ease to implement with X,Y,Z and time or any
+ * combination that you may want.
+ */
+struct RTree {
```
Matematikoi, line 86: I see. Didn't thought about using this! thanks!

LGTM we can go ahead and merge it!

Issue #609, opened 2024-09-27
Failure on cross-compiling for Windows using mingw


evetion:
**Describe the bug**
Over at https://github.com/JuliaPackaging/Yggdrasil/pull/9453 I try to cross-compile MEOS for different platforms for the Julia programming language. When cross-compiling for Windows (using mingw), this fails:

```
[ 43%] Building C object meos/postgres/port/CMakeFiles/port.dir/pg_bitutils.c.obj
cd /workspace/srcdir/MobilityDB/build/meos/postgres/port && /opt/bin/x86_64-w64-mingw32-libgfortran5-cxx11/x86_64-w64-mingw32-gcc --sysroot=/opt/x86_64-w64-mingw32/x86_64-w64-mingw32/sys-root/ -DGSL_VERSION_STRING=\"2.7.1\" -DMEOS=1 -DMEOS_IS_BIG_ENDIAN=0 -DMOBILITYDB_VERSION_STRING="\"MobilityDB 1.2.0\"" -DNO_FFSL -DNPOINT=0 -DPOSE=0 -DPOSTGIS_PGSQL_VERSION=140 -DPOSTGIS_VERSION_NUMBER=30500 -DPOSTGIS_VERSION_STRING="\"PostGIS 3.5.0\"" -DPOSTGRESQL_VERSION_NUMBER=140200 -DPOSTGRESQL_VERSION_STRING="\"PostgreSQL 14.2\"" -DSYSTEMTZDIR=\"/usr/share/zoneinfo\" -D_USE_MATH_DEFINES @CMakeFiles/port.dir/includes_C.rsp -O3 -DNDEBUG -O2 -fvisibility=hidden   -Wall -Wextra -std=gnu1x -Wunused-parameter -Wno-strict-aliasing -MD -MT meos/postgres/port/CMakeFiles/port.dir/pg_bitutils.c.obj -MF CMakeFiles/port.dir/pg_bitutils.c.obj.d -o CMakeFiles/port.dir/pg_bitutils.c.obj -c /workspace/srcdir/MobilityDB/meos/postgres/port/pg_bitutils.c
/tmp/ccIjAGDk.s: Assembler messages:
/tmp/ccIjAGDk.s:109: Error: incorrect register `%eax' used with `q' suffix
make[2]: *** [meos/postgres/port/CMakeFiles/port.dir/build.make:77: meos/postgres/port/CMakeFiles/port.dir/pg_bitutils.c.obj] Error 1
[...]

PR #610, opened 2024-09-27
Do not compile pg_bitutils.c as it's not needed and fails with mingw

mschoema:


mschoema:
This should fix #609
Not sure what the problem was, but we are not using functions from pg_bitutils, so just don't compile it.

For documentation purposes in case we need pg_bitutils later:
The problem potentially comes from either HAVE_X86_64_POPCNTQ or HAVE__BUILTIN_POPCOUNT.
We hard-code HAVE__BUILTIN_POPCOUNT to 1 in pg_config.h.in instead of checking if it is available, so that might be the issue. Cmake detects that HAVE_X86_64_POPCNTQ should be false, so I don't think it is related to HAVE_X86_64_POPCNTQ, but I can't be sure.

From the Yggdrasil build process with mingw:
```
[13:47:02] -- Looking for __get_cpuid
[13:47:03] -- Looking for __get_cpuid - found
[13:47:03] -- Performing Test HAVE_X86_64_POPCNTQ
[13:47:03] -- Performing Test HAVE_X86_64_POPCNTQ - Failed
```

estebanzimanyi (approved):

PR #611, opened 2024-09-28
Improve code with cppcheck

estebanzimanyi:
* List global variables and constants


meos/src/general/globals.c
```diff
@@ -0,0 +1,87 @@
...
+/*****************************************************************************
+ * MEOS global variables 
+ *****************************************************************************/
+
+/* File /meos/src/general/error.c */
+
+/**
+ * @brief Global variable that keeps the last error number
+ */
+int MEOS_ERR_NO = 0;
```
mschoema, line 55: `MEOS_ERR_NO` is defined as `static` in **error.c**, so this should not be here.

meos/src/general/globals.c
```diff
@@ -0,0 +1,87 @@
...
+
+/*****************************************************************************/
+
+/* File /meos/src/general/meos.c */
+
+/* Global variables for GSL */
+
+bool MEOS_GSL_INITIALIZED = false;
+gsl_rng *MEOS_GENERATION_RNG = NULL;
+gsl_rng *MEOS_AGGREGATION_RNG = NULL;
```
mschoema, line 70: Similarly, if the global variable can be defined as `static` in a single file, this should be preffered I think.
Only export global variables used in multiple source files.

mobilitydb/src/general/globals.c
mschoema, line 1: Same here, I would keep the global variables in meos_catalog.c that can be static.

other_globals.c
mschoema, line 1: We need to mention somewhere that this file is just for documentation purposes and is not being compiled.

mschoema (changes requested):
I only commented on the global variable handling, but the cppcheck changes are ok.

meos/src/point/tpoint_rtree.c
mschoema, line 1: These changes revert the changes made to tpoint_rtree.c in a previous PR

mobilitydb/src/general/meos_catalog.c
```diff
@@ -302,14 +291,25 @@ populate_operoid_cache()
...
+/**
+ * @brief Initialize Oid cache
+ */
+void
+mobilitydb_initialize_cache()
+{
+	/* Cache the info if we don't already have it */
+	if (! MOBILITYDB_CONSTANTS)
+		MOBILITYDB_CONSTANTS = get_mobilitydb_constants();
 }
```
mschoema, line 313: This function should probably be declared `static inline`

mschoema (changes requested):

mschoema (approved):

PR #616, opened 2024-10-04
Enable meos build

estebanzimanyi:


mschoema (commented):
This will not be enough to solve the issue, since the functions are defined in postgis/libpgcommon/, which is not being compiled with MEOS.

mschoema (approved):

PR #634, opened 2024-10-17
Improve MEOS API for PG types

estebanzimanyi:

PR #637, opened 2024-10-17
Bring back the timezone pg functions to MEOS to fix meos_initialize_timezone in stable-1.2

Davichet-e:
Right now `meos_initialize_timezone` does not work in the branch `stable-1.2`. The reason is that the timezone library is not accessible since #447, this make meos throw the following error when trying to use `meos_initialize_timezone`:
```
undefined reference to `meos_initialize_timezone'
```

This is already changed in master since https://github.com/MobilityDB/MobilityDB/pull/634, but it's not fixed. This PR is against stable1.2, so I'm not sure about the consequences of the versioning.

## Unknowns
Maybe I should comment line 8?

### DIFF
--- a/meos/postgres/timezone/CMakeLists.txt
+++ b/meos/postgres/timezone/CMakeLists.txt
@@ -4,5 +4,5 @@ add_library(timezone OBJECT
   pgtz.c
   )
 
-set_property(TARGET timezone PROPERTY C_VISIBILITY_PRESET hidden)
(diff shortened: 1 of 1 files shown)


mschoema:
Ah indeed, unfortunately this was not caught by the simple action we have setup for MEOS.
I'll accept the PR and quickly delete and recreate the Release 1.2.0, because I don't want to create a 1.2.1 just for this.

mschoema (approved):

PR #638, opened 2024-10-18
Fix tprecision for discrete interpolation

estebanzimanyi:


meos/src/general/temporal_analytics.c
```diff
@@ -273,7 +273,9 @@ tsequence_tprecision(const TSequence *seq, const Interval *duration,
       {
         start = end; end = NULL;
       }
-      if (interp != DISCRETE)
+      if (interp == DISCRETE)
+        k = 0;
+      else
```
mschoema, line 278: Does this not change the tests? If not, we should probably add a couple that test this condition.

mschoema (approved):

PR #642, opened 2024-11-08
Fix tsample and tprecision

estebanzimanyi:
Unify the API of both functions


estebanzimanyi:
Will keep the current API but correct tprecision in another PR.

PR #645, opened 2025-01-06
Add cast tbool -> tint

estebanzimanyi:


mschoema (approved):

mschoema:
This might be overkill, but if we want to optimize this function we could just do a `temporal_copy` and then overwrite all `temptype` values in a single loop. This works, because the memory representation of `true` / `false` is the same as `1` / `0` (at least should be, not sure about endianness problems).

PR #648, opened 2025-01-24
Add interpolation parameter to appendInstant

estebanzimanyi:


mobilitydb/test/general/expected/040_temporal_aggfuncs.test.out
```diff
@@ -547,13 +547,82 @@ WITH temp(inst) AS (
...
+  SELECT tint '1@2000-01-01' UNION
+  SELECT tint '2@2000-01-02' UNION
+  SELECT tint '3@2000-01-03' UNION
+  SELECT tint '4@2000-01-04' UNION
+  SELECT tint '5@2000-01-05' )
+SELECT appendInstant(inst ORDER BY inst, 'discrete') FROM temp;
+                                                                          appendinstant                                                                           
+------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ [1@Sat Jan 01 00:00:00 2000 PST, 2@Sun Jan 02 00:00:00 2000 PST, 3@Mon Jan 03 00:00:00 2000 PST, 4@Tue Jan 04 00:00:00 2000 PST, 5@Wed Jan 05 00:00:00 2000 PST]
+(1 row)
```
mschoema, line 560: The `discrete` parameter should be put before the `ORDER BY`. Currently it's seen as part of the order by and the function thus assumes `linear` interpolation by default.

This should be clearly emphasized in the documentation. Maybe with a link to the postgres doc as well:
https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES
https://www.postgresql.org/docs/current/xaggr.html#XAGGR-POLYMORPHIC-AGGREGATES

mschoema (approved):

PR #651, opened 2025-02-02
Improve cmake code for mobilitydb_datagen extension

mschoema:
This is the first in a set of PRs to improve the CMake code of MobilityDB and MEOS.
Major changes in this PR:

- Update cmake_minimum_required from 3.7 to 3.12 (needed for other changes in the PR).
- Rename the `.control` files to `.in.control`.
- Add a `header.in.sql` file for the datagen extension. This file contains the license, a notice that the extension sql file is auto-generated, and the line to prevent manually sourcing the file in psql.
- Improve generation of extension sql file: First configure header, then append all needed sql file after removing their license (the header already contains the license).
- Remove unused files `(control.in`, `test.sql)`


mschoema:
@estebanzimanyi 
I also added the `random_geo.sql` file in the list of sql files for mobilitydb_datagen, since it is needed by the other files.
Was there a reason not to include it in the first place?

estebanzimanyi (approved):

mschoema:
> @estebanzimanyi I also added the `random_geo.sql` file in the list of sql files for mobilitydb_datagen, since it is needed by the other files. Was there a reason not to include it in the first place?

I thought we needed `random_geo.sql`, because `random_int` and `random_float` are defined in there, but they are also defined in `random_temporal.sql`, so the file is actually not needed. Removed it for now.

estebanzimanyi (approved):

PR #665, opened 2025-04-06
Implementation of the trgeo type

estebanzimanyi:
* Initial implementation, will be completed in subsequent PRs


CMakeLists.txt
```diff
@@ -122,6 +129,14 @@ else()
   add_definitions(-DPOSE=0)
 endif()
 
+if(RGEO)
+  message(STATUS "Including rigid geometries")
+  add_definitions(-DRGEO=1)
+else()
+  message(STATUS "Excluding rigid geometries")
+  add_definitions(-DRGEO=0)
+endif()
```
mschoema, line 138: If DRGEO is enabled we need to also enable DPOSE, since it can't work without it. Something more should be added here to make sure we don't have DPOSE disabled when DRGEO is enabled.

mschoema (approved):

PR #667, opened 2025-04-15
Improve interpolation and validity functions

estebanzimanyi:
This PR concludes the main changes induced by the refactoring of MobilityDB code to cope with the new spatial data types that have been introduced recently, namely, `tgeometry`, `tgeography`, `tpose`, `trgeometry`, and `tcbuffer`, which complement the previous spatial data types, namely, `tgeompoint`, `tgeogpoint`, and `tnpoint`.
Subsequent PRs will improve both the implementation and the documentation of this PR.


mschoema (changes requested):

meos/src/cbuffer/cbuffer.c
```diff
@@ -127,16 +127,21 @@ cbuffersegm_locate(const Cbuffer *start, const Cbuffer *end,
...
       return -1.0;
   }
-  long double result;
+  else
+  {
+    /* If the radii are equal return result1 where
+     * - result1 = -1.0 if gs1 == gs2 == gs, or
+     * - result1 in [0,1] if gs1 != gs2 */
+    return result1;
+  }
```
mschoema, line 136: We still need to check that `start->radius == value->radius`, otherwise we should always return `-1`. And what should we return if we have `gs1 == gs2 == gs` and `start->radius == end->radius == value->radius`? This only happens for a constant segment (`start == end`) where `value` is the same as the constant. What do we return in this case for the other types?

mschoema (changes requested):

mschoema:
There is also still a merge conflict in the `meos/src/cbuffer/tcebuffer_spatialfuncs.c` file that needs to be resolved..

meos/src/general/tsequence.c
```diff
@@ -206,6 +206,201 @@ datum_collinear(Datum value1, Datum value2, Datum value3, meosType basetype,
...
+      DatumGetNpointP(value));
+#endif
+// #if POSE || RGEO
+  // if (basetype == T_POSE)
+    // return posesegm_locate(DatumGetPoseP(value1), DatumGetPoseP(value2),
+      // DatumGetPoseP(value));
+// #endif
+  meos_error(ERROR, MEOS_ERR_INTERNAL_TYPE_ERROR,
+    "Unknown locate function for type: %s", meostype_name(basetype));
+  return false;
```
mschoema, line 273: This should return -1 instead of false

mobilitydb/test/general/expected/022_temporal.test.out
```diff
@@ -6717,19 +6717,19 @@ SELECT minusValues(tfloat '[1@2000-01-01, 2@2000-01-02]', 1 - 1e-17);
...
 (1 row)
 
 SELECT minusValues(tfloat '(1@2000-01-01, 2@2000-01-02]', 1 + 1e-12);
- minusvalues 
--------------
- 
+                            minusvalues                             
+--------------------------------------------------------------------
+ {(1@Sat Jan 01 00:00:00 2000 PST, 2@Sun Jan 02 00:00:00 2000 PST]}
 (1 row)
```
mschoema, line 6727: Is it normal that this changed? The second query seems correct, which it wasn't before. But I don't understand why the answer to the first query changed? I think the old answer (`{(1@Sat Jan 01 00:00:00 2000 PST, 2@Sun Jan 02 00:00:00 2000 PST]}`) was correct.

estebanzimanyi: The condition for locating a value in a segment were uniformized in all *segm_locate() functions 
https://github.com/estebanzimanyi/MobilityDB/blob/tcbuffer/meos/src/general/tsequence.c#L232

mschoema (approved):

PR #669, opened 2025-04-27
Improve MEOS examples

estebanzimanyi:
* Replicate PostGIS PROJ cache in MEOS
* Improve functions for spatial types


mschoema (changes requested):
I'm still going through the PR, but it will take a bit more time. These are the things I've already noticed.

mschoema (changes requested):

mschoema (approved):
Requested changes will be included in next PR

PR #674, opened 2025-06-21
Add Ways cache

estebanzimanyi:


mschoema (commented):
This seems good, except for one thing. Why did we rename `tDistance` to `distance`? I'm not sure this is the best idea, as it's a functions that's used a lot and removing the backwards compatibility with 1.2 isn't ideal.

PR #675, opened 2025-07-06
Enable DBSCAN clustering in MEOS

estebanzimanyi:


mschoema:
The PR looks good, but are we sure this is something we want to add to MEOS?
It seems we are just exposing more of liblwgeom through MEOS, which is probably not the best way of going at it.

postgis/liblwgeom/lwunionfind.c
```diff
@@ -154,6 +154,8 @@ UF_get_collapsed_cluster_ids(UNIONFIND* uf, const char* is_in_cluster)
 	for (i = 0; i < uf->N; i++)
 	{
 		uint32_t j = ordered_components[i];
+		// MEOS
+		new_ids[j] = -1;
 		if (!is_in_cluster || is_in_cluster[j])
```
mschoema, line 159: Is this because of a bug? Should this be pushed to PostGIS?

mschoema (approved):

PR #686, opened 2025-07-30
Enable Rtree for all bounding boxes

estebanzimanyi:


meos/src/geo/temporal_rtree.c
```diff
@@ -0,0 +1,862 @@
...
+{
+  for (int i = 0; i < node->count; ++i)
+  {
+    if (overlaps_stbox_stbox(query, &node->boxes[i]))
+    {
+      if (node->isRoot == RTREE_ROOT_NODE)
+        add_answer(node->ids[i], ids, count);
+      else
+        node_search(node->nodes[i], query, ids, count);
+    }
```
mschoema, line 703: This doesn't look right. the id's should be at the leaves, not at the root.

meos/include/temporal/temporal_rtree.h
```diff
@@ -0,0 +1,93 @@
...
+#define MINITEMS ((MAXITEMS) * (MINITEMS_PERCENTAGE) / 100 + 1)
+#define RTREE_ROOT_NODE true
+#define RTREE_INNER_NODE false
+
+/**
+ * @brief Internal representation of an RTree node.
+ */
+typedef struct RTreeNode
+{
+  bool isRoot;
```
mschoema, line 58: Instead of a boolean isRoot, I would use an enum RTreeNodeType, which could be root, inner, or leaf.

meos/include/temporal/temporal_rtree.h
```diff
@@ -0,0 +1,93 @@
...
+  int count;
+  union 
+  {
+    struct RTreeNode *nodes[MAXITEMS];
+    int64 ids[MAXITEMS];
+  };
+  /* STBox is the largest MEOS bounding box */
+  STBox boxes[MAXITEMS];
```
mschoema, line 66: I'm not sure this is the best idea, we're wasting a lot of space. I would maybe put a flexible array member instead: `char boxes[]`. Then accessing box `n` would just be `(BOX_TYPE) node->boxes + n * bboxsize`. This seems cleaner and will use less memory for Span or TBox RTrees.

meos/src/temporal/temporal_rtree.c
mschoema, line 1: This file seems to be a duplicate of `meos/src/geo/temporal_rtree.c`. I don't know which one should be kept. I would maybe even call it `bbox_rtree`, since we use it to store and query different bbox types.

meos/src/geo/tspatial_rtree.c
```diff
@@ -412,7 +413,7 @@ node_split(RTree *rtree, RTreeNode *node, STBox *box, RTreeNode ** right_out)
     double min_dist = rtree->get_axis(&node->boxes[i], largest_axis, false) -
       rtree->get_axis(box, largest_axis, false);
     double max_dist = rtree->get_axis(box, largest_axis, true) -
-      rtree->get_axis(&node->boxes[i], largest_axis, true);
+      rtree->get  _axis(&node->boxes[i], largest_axis, true);
```
mschoema, line 416: Added space is a typo

mschoema (changes requested):
I'm wondering if we should keep this for 1.4. The implementation is in my opinion not complete, and we shouldn't add it half-baked to 1.3.

estebanzimanyi:
Perfect to leave it for 1.4. I will correct the issues in a forthcoming PR.

meos/include/meos.h
```diff
@@ -289,6 +289,19 @@ typedef struct
...
 
+/* RTree functions */
+
+extern RTree *rtree_create_intspan();
+extern RTree *rtree_create_bigintspan();
+extern RTree *rtree_create_floatspan();
+extern RTree *rtree_create_datespan();
+extern RTree *rtree_create_tstzspan();
+extern RTree *rtree_create_tbox();
+extern RTree *rtree_create_stbox();
```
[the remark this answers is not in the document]
estebanzimanyi: In MEOS we always provide wrapper functions that hides Oids or similar from the user, see for example
https://github.com/MobilityDB/MobilityDB/blob/master/meos/src/temporal/spanset_meos.c#L57-L131
I followed the same approach here.

meos/src/geo/temporal_rtree.c
estebanzimanyi: This was already on the master branch. 
https://github.com/MobilityDB/MobilityDB/blob/master/meos/src/geo/tspatial_rtree.c#L574-L587
Could you please specify the problem ?

meos/examples/rtree_example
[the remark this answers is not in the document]
estebanzimanyi: Removed

meos/examples/rtree_example_new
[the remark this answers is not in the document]
estebanzimanyi: Removed

meos/examples/rtree_example_new.c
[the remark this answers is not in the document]
estebanzimanyi: Fixed by removing the old version rtree_example.c and renaming the file to rtree_example.c

meos/include/meos_geo.h
```diff
@@ -596,10 +596,10 @@ extern bool stbox_ne(const STBox *box1, const STBox *box2);
...
 /* RTree functions */
 
-extern RTree *rtree_create_stbox();
-extern void rtree_free(RTree *rtree);
-extern void rtree_insert(RTree *rtree, STBox *box, int64 id);
-extern int *rtree_search(const RTree *rtree,const STBox *query, int *count);
+// extern RTree *rtree_create_stbox();
+// extern void rtree_free(RTree *rtree);
+// extern void rtree_insert(RTree *rtree, STBox *box, int64 id);
+// extern int *rtree_search(const RTree *rtree,const STBox *query, int *count);
```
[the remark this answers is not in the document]
estebanzimanyi: Removed.

meos/include/temporal/temporal_rtree.h
estebanzimanyi: Done

meos/include/temporal/temporal_rtree.h
estebanzimanyi: Changed as suggested and defined a macro `RTREE_NODE_BBOX_N` to access the n-th bounding box of a node

estebanzimanyi:
> I'm wondering if we should keep this for 1.4. The implementation is in my opinion not complete, and we shouldn't add it half-baked to 1.3.

IMHO the last commit solved all issues. The changes of this PR basically does not change the algorithm only enable bounding boxes of varying sizes. For me it still can go in 1.3.

meos/src/temporal/temporal_rtree.c
mschoema: Ok

mschoema:
> IMHO the last commit solved all issues. The changes of this PR basically does not change the algorithm only enable bounding boxes of varying sizes. For me it still can go in 1.3.

Fair, the RTrees are already there anyway. I still think there are things left to do, but this PR can already be added to 1.3-alpha1. Maybe I'll find the time to improve the RTrees further before the actual 1.3 release.

estebanzimanyi:
> > IMHO the last commit solved all issues. The changes of this PR basically does not change the algorithm only enable bounding boxes of varying sizes. For me it still can go in 1.3.
> 
> Fair, the RTrees are already there anyway. I still think there are things left to do, but this PR can already be added to 1.3-alpha1. Maybe I'll find the time to improve the RTrees further before the actual 1.3 release.

Thanks ! This enables to be used for other bindings, in particular DuckDB.

meos/src/geo/temporal_rtree.c
mschoema: Well, an internal node stores pointers to its children (`node->nodes`, which maybe should be renamed for clarity). The id's are stored in the leaf nodes (`node->ids`). Thus, we need something like:

```
if (node->nodetype == RTREE_LEAF_NODE)
    add_answer(node->ids[i], ids, count);
else
    node_search(node->nodes[i], query, ids, count);
```

Here, we always directly return the ids from the root node, which is incorrect.

meos/include/temporal/temporal_rtree.h
mschoema: I think the distinction is needed for the `node_search` function as is discussed above.

estebanzimanyi: I have added support for the three node types in the last commit and the test program works as before. However, it remains to be determined where in the algorithm to set the node to RTREE_NODE_LEAF. Any idea ?

meos/src/geo/temporal_rtree.c
estebanzimanyi: Where in the algorithm the nodetype should be set to `RTREE_LEAF_NODE` to be used here ? See also comment below.

PR #692, opened 2025-08-04
Improve performance problems identified by Issue #679

estebanzimanyi:

PR #694, opened 2025-08-06
Follow unary_union parameter in tpointseqset_linear_trajectory

Diviloper:
Related to #692.
Apply final `geom_unary_union` in `tpointseqset_linear_trajectory` according to `unary_union` parameter.


Diviloper:
@estebanzimanyi this change requires modifying some tests, I started with some, but reach some that use a `temp` thing that I'm not familiar with. Could you take a look?

Issue #679, opened 2025-07-23
Poor query performance on postgis 3.5 builds


noodletired:
**Describe the bug**
Hi there! Just testing newer mobilitydb 1.3 features using public docker images and noticed query performance on `3.5-master` is very poor. For instance a simple query testing if an arbitrary shape `eIntersects` a GIST indexed `tgeompoint` column on `3.4-master`/`latest` takes <500ms, but on `3.5-master` takes >>500'000ms on the same hardware. Have confirmed this behaviour on both postgres 16 and 17 images. 

Assuming this is either related to the newer postgis 3.5 version or instability of `master`, but thought I'd bring it to moblitydb developers' attention. Looking forward to stable 1.3 release. 😄 

**To Reproduce**
1.  `docker run -p 25432:5432 mobilitydb/mobilitydb:17-3.5-master`
2.  Statements to insert 10000 rows of dummy data (sequences of 300 random points)
```SQL
CREATE EXTENSION IF NOT EXISTS mobilitydb;
CREATE EXTENSION IF NOT EXISTS pgcrypto;
CREATE TABLE test_paths (
  id SERIAL PRIMARY KEY,
  path tgeompoint
);
CREATE INDEX idx_test_paths_spatial ON test_paths USING GIST (path);
```

```SQL
DO $$
DECLARE
    i INT;
    j INT;
    geom_str TEXT;
    t TIMESTAMPTZ;
    lon DOUBLE PRECISION;
    lat DOUBLE PRECISION;
BEGIN
    FOR i IN 1..10000 LOOP
        geom_str := '';
        t := clock_timestamp();

        FOR j IN 0..299 LOOP
            lon := random() * 360 - 180;  -- Longitude range
            lat := random() * 180 - 90;   -- Latitude range
            geom_str := geom_str || format(
                'Point(%s %s)@%s',
                lon,
                lat,
                (t + (j * interval '1 second'))::TEXT
            );

            IF j < 299 THEN
                geom_str := geom_str || ', ';
            END IF;
        END LOOP;

        EXECUTE format(
            'INSERT INTO test_paths(path) VALUES ((''[%s]''))',
            geom_str
        );
    END LOOP;
END $$;
```

3.  Query intersect
```SQL
WITH shape AS (
   SELECT ST_MakeEnvelope(-10, 10, -20, 20) AS geom
) SELECT id
FROM test_paths, shape
WHERE eIntersects(path, shape.geom)
```

[...]

PR #695, opened 2025-08-06
Improve performance problems identified by Issue #679

estebanzimanyi:
Continuation of the PR #694 by @Diviloper


estebanzimanyi:
The difference comes when using or not ST_UnaryUnion. It introduces many floating point imprecisions that invalidate many of the equality or containment tests. Now the usage of this function is by default set to false.

PR #697, opened 2025-08-09
Fix datum_eq for geography points

estebanzimanyi:


Diviloper (commented):
Should the `doc/` files be in this PR?
For the rest, looks good to me.

Diviloper (dismissed):
Indeed, although the documentation for temporal circular buffers in Spanish seems to be missing

Diviloper (approved):

PR #715, opened 2025-09-10
Fix WKB input for text values

estebanzimanyi:


mschoema:
I'm currently looking at this, and I have ideas on how to improve the code to remove redundancy even further than just with the macro.

Also, it seems to me that the issue with the text wkb doesn't come from the `text_from_wkb_state` function, but rather from the `text_to_wkb_buf`. Indeed, `text_to_wkb_buf` calls `bytes_to_wkb_buf` to encode the characters, which will swap the order of all bytes (characters) in the string. Thus, we are essentially serializing a string by writing its reverse, which is not a wanted behaviour. The byte-swapping should only happen for multi-byte int/float types (which includes time/date types, which are essentially ints). But since text is made up of individual chars, we do not need to swap anything.

Considering all of this, I think we should close this PR and I'll reopen a new one with the correct fixes and the new refactor. Also, how come the current behaviour doesn't cause any issues in the tests? Is it because they all run on little-endian architecture?

PR #718, opened 2025-09-20
Fix big-endian handling for ttext type and refactor code to remove redundancy

mschoema:
Each individual commit in this PR contains a specific code change and has a detailed comment. I will try to merge this PR using 'rebase and merge', which avoids an undesirable squash when the commits in a PR have each their own scope and comment.

I'm not sure if this is better than squashing all commits of a PR into a single commit with a single message, but let's try it out and see.


mschoema:
Well the commits look nice after the rebase and merge, but github doesn't automatically add a link to the PR in their message. I don't think that's ideal in our case. I will stick with squash and merge in the future but will try to write a clean commit message with each PR.

PR #726, opened 2025-09-30
Extract /postgres from /meos

estebanzimanyi:
In order to prepare the upgrade of the PostgreSQL code embedded in MEOS from version 14 to version 18 it is necessary to move the /postgres directory outside from the /meos directory and make it completely independent of the MEOS library.


meos/CMakeLists.txt
```diff
@@ -89,6 +89,13 @@ add_definitions(-DPOSTGIS_PGSQL_VERSION=${POSTGIS_PGSQL_VERSION})
 # MobilityDB directories
 #--------------------------------
 
+# PostgreSQL
+configure_file(../postgres/pg_config.h.in ../postgres/pg_config.h)
+include_directories(SYSTEM "${CMAKE_SOURCE_DIR}/postgres/timezone")
+include_directories(SYSTEM "${CMAKE_SOURCE_DIR}/postgres")
+include_directories(SYSTEM "${CMAKE_BINARY_DIR}/postgres")
+add_subdirectory("${CMAKE_SOURCE_DIR}/postgres" "postgres")
```
mschoema: You don't need to do `include_directories` for the subdirectories (e.g. `/timezone` here), since the include statements are relative to the top-level `/postgres` folder.
This is also what causes the issue with `DBL_MAX`, since including `<float.h>` then finds `/postgres/utils/float.h` instead of the system `float.h` file.

PR #732, opened 2025-10-21
Fix workflow for macos 14

leilabourouf:
For macOS 14, if the installed PostgreSQL version is older than 17, you must manually build and install PostGIS.
This is because Homebrew’s postgis formula now only supports PostgreSQL 17.
The workflow file includes the necessary build steps to compile PostGIS 3.4.2 from source for PostgreSQL@14.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -45,17 +45,38 @@ jobs:
           mkdir build
           cd build
           cmake -DPOSTGRESQL_BIN="/opt/homebrew/opt/postgresql@${{ matrix.psql }}/bin" -DCMAKE_BUILD_TYPE=Release ..
+
       - name: Build
         run: |
           cd build
           make -j
           sudo make install
+
+      - name: Library symlinks
+        run: |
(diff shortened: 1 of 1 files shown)


mschoema:
I think we can close this PR. I'll open a new on from my current branch on top of which we can build.
I'm currently looking into the latest issue, but this seems to be a problem in the CMakeLists files, not a problem in the workflow. So I'm not sure if I should fix this in a second PR or do everything in one.

If you decide to keep looking into this and find a solution for the CMakeLists, feel free to let me know. The you can either open a new PR with the solutions or send me a patch and I'll commit it (this method still keeps you as the patch author).

Also, I don't believe we need to update the findPostGIS script, since we only want to detect the postgis version installed for the particular postgres installation we are compiling against, and not other postgis installations that might be present in your file system. This becomes important when having multiple installations on the same system, for example when you need to work with multiple versions at the same time.

One last recommendation. I'm not sure which text editor you are using, but you should try to make sure that it always adds a trailing newline at the bottom of the file. Otherwise we get these red stop signs at the bottom of the file, as can be seen in the 'Files changed' tab.

PR #733, opened 2025-10-21
Fix macos workflow and installation

mschoema:
This PR updates the PG versions used in the macos workflow to 17 and 18. macos-13 is also removed as it will get discontinued soon, and macos-15 is added. Since we don't use the default homebrew postgres version anymore (which was 14), we need to explicitly specify the paths.

We also update the CMakeLists to use POSTGRESQL_BIN_DIR instead of hardcoded paths when linking against postgres in macos builds.

A final issue remains as can be seen here: https://github.com/mschoema/MobilityDB/actions/runs/18694452236/job/53308149277

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -55,7 +55,6 @@ jobs:
       - name: test install
         run: |
           export PATH="/opt/homebrew/opt/postgresql@${{ matrix.psql }}/bin:$PATH"
-          initdb -D /opt/homebrew/var/postgresql@${{ matrix.psql }}/
           pg_ctl -D /opt/homebrew/var/postgresql@${{ matrix.psql }}/ start
           createdb ___mobdb___test___
           psql -d ___mobdb___test___ -c "CREATE EXTENSION mobilitydb CASCADE;  SELECT postgis_full_version(); SELECT mobilitydb_full_version();"


mschoema:
@leilabourouf Feel to send patches if you figure out the issue. I will also look in parallel.

leilabourouf:
I think i have something that works, i opened another PR since I am not very familiar on how to send a patch.
the issue came from .so files they need to be .dylib

mschoema:
I see, so we just need to add `SUFFIX ".dylib"` in the CMakeLists file and it works.
Thank you very much for looking into this and fixing the issue.

I also see that in your PR you changed the `initdb` path to `$HOME/pgdata`. Is there a reason for that?
In my current PR I simply removed the `initdb` command as homebrew already initializes a data directory at `/opt/homebrew/var/postgresql@${{ matrix.psql }}/`, so I used that in the `pg_ctl` call.

mschoema:
If you are interested in learning how to create commit patches, I can guide you through it and I will add your commit to this PR and merge it. If so, you can try the following in your local git repo:

```bash
git remote add mschoema mschoema/MobilityDB
git fetch mschoema
git switch -c macos mschoema/macos
```

This will create a local branch called `macos` that starts from the current state of my branch (with the 3 commits in this PR). Then, you can add a new commit with the fix on this new branch. This would thus contain the one-line fix with `SUFFIX ".dylib"`. Once you have created this commit, there are two common options for collaboration:

1. You can push this branch to your github and create a new PR from that branch. The PR would then contain the 4 commits (the 3 in this PR, plus your new fix). This would probably be the easiest solution and is commonly also used if you plan on continuing to work on the PR.
2. You can create a patch from the commit you just made and send it to me. The easiest way would be by attaching the patch file to a message in this PR. This way, I can apply your patch on my branch which will add it to this PR. This method is more commonly used if you want to send a small fix or change to some else's PR while letting them keep in woking on the PR from their branch.

For method 2, the way you create a patch from your commit is the following:

```bash
git format-patch HEAD^1 -v 1
```

This will create a new file called `v1-0001-the-name-of-your-commit.patch` that contains the changes in the last commit made on the current branch. Then you can send this file, which someone else (me) can apply to their branch using `git am`.

Let me know if you're interested in trying either option. If not, I can always just commit your current PR directly, but I think this would be a good thing to try out.

leilabourouf:
Sorry, I forgot about this. We can try out the patch approach if I ever help with another issue. Thanks for the explanation and for merging @mschoema !

PR #740, opened 2025-10-30
Add quad-tree and k-d tree indexes to MEOS

estebanzimanyi:


estebanzimanyi:
Now that have finished the work on the R-tree, maybe you could have a look at this PR #740 where I prepare the infrastructure so that Gabriel can begin working on these indexes. I have just rebased this PR.

estebanzimanyi:
## Architectural note for the MobilityDuck-side follow-up

Capturing this here so the design conversation lands next to the code rather than getting deferred to the future follow-up PR.

The three new headers — `meos/include/temporal/span_index.h`, `meos/include/temporal/tbox_index.h`, `meos/include/geo/stbox_index.h` — currently mix two API audiences:

| Subset | API | Audience |
|---|---|---|
| **MEOS-pure** | `getQuadrant2D/4D/8D`, `overlap2D/contain2D/left2D/...`, `distance_span_nodespan`, `distance_*_nodebox`, `spannode_init/copy`, `spannode_quadtree_next`, `spannode_kdtree_next`, `SpanNode/STboxNode/TboxNode` structs | Any MEOS consumer (MobilityDB, MobilityDuck, future bindings) |
| **PG-glue** | `span_index_leaf_consistent(... StrategyNumber)`, `span_gist_inner_consistent(... StrategyNumber)`, `span_index_recheck(StrategyNumber)`, `span_spgist_get_span(Datum, ...)` | MobilityDB only (depends on `<postgres.h>` + `<access/stratnum.h>`) |

For MobilityDuck to consume the MEOS-pure subset (build a TSPGTREE index that mirrors the existing TRTREE), the headers as currently shaped force a `<postgres.h>` dependency on a non-PG consumer.

### Proposed follow-up PR (after this one merges)

Split the API surface into two layers:

1. **Pure subset** — re-exposed via `meos_internal.h` (or new `meos_internal_index.h`) so it's part of the published vcpkg-installed include set, with no `<postgres.h>` dependency.
2. **PG-glue subset** — stays in `meos/include/{geo,temporal}/*_index.h` exactly as is. These remain MEOS-internal headers consumed only by MobilityDB.

The split in the implementation files (`meos/src/.../*_spgist.c`) follows the same lines — pure functions stay, PG-dependent functions either stay or move to a thin `mobilitydb/src/.../...` shim.

### What this unblocks

With the pure subset published, MobilityDuck gains the building blocks for a quadtree/k-d index alongside its existing R-tree:
- `STboxNode` + `getQuadrant8D` + `spannode_quadtree_next` for the tree structure.
- `distance_stbox_nodebox` / `distance_span_nodespan` for kNN with priority queue (the same kNN traversal that's currently blocked on the R-tree side because `rtree_search_nearest` doesn't exist).

### Why mention it now

If reviewers (cc @mschoema) have an opinion on the eventual public-vs-internal shape of these primitives — naming, where the split should land, whether the structs themselves should ship via `meos_internal.h` — better captured here while the API is being shaped than after merge.

The follow-up PR itself stays out of this PR's scope. It's a several-hour refactor I'd open against post-merge `master`.

estebanzimanyi:
The same quad-tree and k-d tree SP-GiST index addition rebased onto current master is open as #1171, green across the CI matrix. There the MEOS-side index logic is derived from the current kernel — span, tbox, and stbox types — with the PostgreSQL SP-GiST access methods delegating to it and unsupported types reported through `meos_error`. The tbigint and pointcloud index paths ride their respective family branches.

PR #749, opened 2025-11-28
Add EKF-based outlier filtering and AIS cleaning example

marianaGarcez:
- Introduces temporal_ext_kalman_filter to perform Extended Kalman Filter (EKF)–based outlier filtering on TSEQUENCE/TSEQUENCESET temporal floats and tgeompoint values (non‑geodetic, 2D/3D).

- Adapts the tinyEKF implementation into MEOS (mos/include/temporal/tinyekf_meos.h) with a constant‑velocity motion model, process noise q, measurement variance, and Mahalanobis gating threshold.

- Implements the EKF logic in meos/src/temporal/temporal_analytics.c, supporting two modes:
  - drop: remove outlier instances beyond the gate.
  - fill: insert filtered outliers with predictions.

- Adds a MEOS example program meos/examples/ais_ekf_clean.c that reads AIS CSV data, joins per‑MMSI tgeompoint trajectories, applies EKF cleaning, and writes both cleaned trajectories and optionally the removed points to CSV.

- Exposes the EKF filter in the public API (meos/include/meos.h) and wires it into the SQL layer for temporal and tpoint analytics (mobilitydb/sql/temporal/046_temporal_analytics.in.sql, mobilitydb/sql/geo/076_tpoint_analytics.in.sql).


estebanzimanyi:
Superseded by #793 — same content (marianaGarcez's commit authorship preserved) squashed onto current master and with the 076 regression test's expected outputs refreshed to match the documented 'keep mode' behavior + trimmed to integer counts so macOS/ARM64 floating-point differences don't churn the output. Closing this one.

Issue #425, opened 2024-01-15
Localization support for the MEOS library


estebanzimanyi:
MEOS is the underlying C library driving MobilityDB.
https://libmeos.org/

Bindings for MEOS in several programming languages are being developped
* Python  https://github.com/MobilityDB/PyMEOS
* Java: https://github.com/MobilityDB/JMEOS
* Javascript: https://github.com/MobilityDB/MEOS.js
* C#: https://github.com/MobilityDB/MEOS.NET

MEOS currently does not support localization, that is, use the locale features of the underlying operating system to provide locale-specific collation order, number formatting, and other aspects.

We are looking for contributors that can help us to develop this feature.

estebanzimanyi:
Picking this up. PR #813 lands the first concrete step: a narrow, locale-safety fix that closes a real correctness bug today (libc `strtod()` mis-parses `1.5` as `15` under `LC_NUMERIC=de_DE.UTF-8` because it treats `.` as a thousands separator), plus a CI smoke test that pins the contract going forward.

What this **does not** do: full locale-aware text collation. `varstr_cmp()` already had a stub-with-TODO comment for that work; addressing it properly will need PG's `pg_locale_t` / ICU plumbing brought into MEOS standalone, which is a much larger PR.

Strategy: incremental. Land the narrow fix + CI now, document the locale contract in `meos.h`, then iterate on collation in follow-up PRs.

Follow-ups already scoped:
- Wire `varstr_cmp()` to PG's `pg_locale_t` / `strcoll_l` for locale-aware text collation
- Localize remaining `snprintf("%f", …)` error-message calls
- MSVC `_strtod_l` plumbing once #513 unlocks MSVC support

Issue #403, opened 2023-12-05
UTF8 support for the MEOS library


estebanzimanyi:
MEOS is the underlying C library driving MobilityDB.
https://libmeos.org/
Bindings for MEOS in several programming languages are being developped
* Python  https://github.com/MobilityDB/PyMEOS
* Java: https://github.com/MobilityDB/JMEOS
* Javascript: https://github.com/MobilityDB/MEOS.js
* C#: https://github.com/MobilityDB/MEOS.NET

MEOS currently does not support multibyte characters such as UTF8.

We are looking for contributors that can help us to develop this feature.

estebanzimanyi:
Picking this up. PR #814 lands the bootstrap step — same incremental shape as #812 (Windows) and #813 (locale).

**TL;DR**: MEOS's I/O parsers and formatters are *already* UTF-8 byte-preserving. The encoding-agnostic property comes for free because the parsers only match ASCII delimiters (`"`, `,`, `}`, ...) and UTF-8 continuation bytes (>= 0x80) can't collide with them. Multi-script text round-trips through `ttext_in`/`ttext_out`, `textset_in`/`textset_out`, and `text_in`/`text_out` with bytes preserved verbatim — verified locally with German ("Zürich"), Japanese ("日本"), Arabic ("العربية"), Korean ("한국").

**The remaining limitation**: case folding. `text_upper` / `text_lower` / `text_initcap` use PG's ASCII-only `asc_toupper`/`asc_tolower`, so `text_upper("Zürich")` returns `"ZüRICH"` instead of `"ZÜRICH"`. The non-ASCII bytes survive unchanged (no silent corruption); this is a missing feature, not a bug.

**This PR**: documents the contract in `meos.h` (sibling to the Locale contract from #813), adds the same caveat to the `text_upper`/`text_lower`/`text_initcap` doc comments, and lands a CI job that pins both the round-trip behaviour and the documented case-folding limitation. No implementation changes.

**Follow-ups already scoped**:
- Unicode-aware case folding (ICU or libunistring) — expose as a new function so `text_upper`'s ASCII contract stays stable
- Character-count helpers if/when callers need them
- Locale-aware collation under #425 (separate issue)

PR #751, opened 2025-12-16
Vendor PostgreSQL 18 base types into MEOS as a pgtypes library

estebanzimanyi:
Vendors PostgreSQL 18 base types into MEOS as a `pgtypes` library.

Adds `pgtypes/` — PostgreSQL 18 base-type sources (numutils, timestamp, varlena, pg_locale, regexp, palloc, varatt, jsonb) vendored into MEOS so the standalone library can construct, parse and serialise PG-compatible values without linking libpq or the backend. Installed as `libpgtypes.a`.

The previous `postgres/` vendoring and the `temporal/postgres_types.h` wrapper are removed; the error-handling declarations move to a new `meos_error.h`, and the sources include the pgtypes headers directly.

**Rebased onto current `master`** via a per-file 3-way merge so that master's evolution since the original base is preserved (e.g. the `mult`→`mul` rename, the `tpfunc_unary` typedef) — the only net change is the pgtypes migration itself (add `pgtypes/`, remove `postgres/`, the header restructuring, and the include rewiring). Both targets build green; core temporal regression 102/102, geo 45/45.


meos/examples/rtree_example.c
```diff
@@ -1,7 +1,7 @@
 /*****************************************************************************
  *
  * This MobilityDB code is provided under The PostgreSQL License.
- * Copyright (c) 2016-2023, Université libre de Bruxelles and MobilityDB
+ * Copyright (c) 2016-2025, Université libre de Bruxelles and MobilityDB
```
Davichet-e: Should be 2026

Davichet-e (commented):
Impressive work.

I've one concern, we're copying around 400 files I think. Couldn't we only have what we strictly neeed, and create a script to import the rest at build time?

So leave the new folder like:

```
  pgtypes/
  ├── CMakeLists.txt              # builds the library from extracted+patched sources
  ├── MANIFEST.txt                # explicit list of files to extract from PG
  ├── PG_VERSION                  # single line: "REL_18_0" (pinned tag)
  ├── extract.cmake               # CMake script that does the work
  ├── patches/
  │   ├── 0001-meos-memory-contexts.patch
  │   ├── 0002-no-backend-dups-nonlinux.patch
  │   ├── 0003-codacy-static-analysis.patch
  │   └── 0004-platform-portability.patch
  ├── include/                    # MEOS-authored API headers (stay in repo)
  │   ├── pgtypes.h
  │   ├── pg_bool.h
  │   ├── pg_int.h
  │   ├── pg_float.h
  │   ├── pg_numeric.h
  │   ├── pg_text.h
  │   ├── pg_json.h
  │   ├── pg_date.h
  │   ├── pg_time.h
  │   ├── pg_timestamp.h
  │   └── pg_interval.h
  ├── src/                        # MEOS-authored wrapper .c files (stay in repo)
  │   └── pgtypes_meos_api.c      # thin wrappers exposing pg_* prefixed names
  └── pg_config.h.in              # templated by CMake (stays)
```

And the code vendored from PG, get it using a cmake script.

This would dramatically cut the repository's footprint, a real improvement for clones, CI, etc.

meos/examples/rtree_example.c
estebanzimanyi: Fixed in cf476d450 — applied a tree-wide bump on this branch for every `Copyright (c) XXXX-{2023|2025}, Université libre de Bruxelles and MobilityDB` line, including the new files added by this PR. Master gets the same bump via #842. Vendored notices (PostGIS, PostgreSQL Global Development Group, individual authors) are deliberately left at their upstream values — see #842 for the rationale.

estebanzimanyi:
Thanks for raising this — you're seeing something real, and I think the picture is bigger than this PR. Let me share the overall direction so we can decide together.

**What this PR is actually bootstrapping**

The 400 files of vendored PG code aren't there because we needed them for one feature — they're the substrate for a class of capabilities we want to enable in MEOS over time:

- **#403 UTF8 / multibyte support** → needs `mbutils.c`, `pg_wchar.h`, encoding tables.
- **#425 Locale support** (collation, formatting) → needs `pg_locale.c`, `formatting.c`, ICU shims.
- **Temporal text / timestamptz operations that respect collation / timezone** → pulls in regex paths, `pgtz.c`, IANA glue.
- **The general "lift any PG function over any PG type to its temporal version" capability.** Once we have a stable wrapper layer over PG's type system, lifting becomes pure wiring: `tlift(f, instant_value)` calling `f` in a loop. No per-type or per-function porting work for users who ask "I'd like a temporal `f(x)` over PG type `T`".

So the realistic trajectory is **400 files today, ~1000–1500 a few years out**. This isn't transitional vendoring; it's a long-term substrate that grows with the lift surface.

**Why your concern is right, and gets stronger over time**

At 400 files the boundary works on convention; by 1500 it won't. Patches as `*.patch` files compound the same way — each PG-version bump becomes an N-patch re-baselining exercise, and patches review poorly. Everything you're flagging — footprint, audit, accidental edits across the boundary — becomes more acute as the surface grows, not less.

**Where I'd land — and why not the proposal as written**

Four candidate architectures, ordered by long-term fit:

1. **Status quo (scattered in-tree).** Fails at scale.
2. **Isolated in-tree snapshot** under e.g. `meos/postgres-snapshot/` with `MANIFEST.txt` + `PG_VERSION` + `patches/` + `sync.sh`. Captures most of your audit/boundary benefit. Doesn't shrink the repo. Distro-friendly, bindings-friendly, offline-friendly.
3. **Extract-at-build (your proposal).** Cleanest mechanically. My concern is real-world fallout: distros (Debian, Conda-forge, Fedora, Homebrew) sandbox builds with no network access; the MEOS bindings ecosystem (PyMEOS, JMEOS, MEOS.NET, MEOS.js, RustMEOS, GoMEOS) consumes MEOS as a self-contained source bundle. Both groups would need a "pre-extracted tarball" pathway, which adds release-engineering complexity. Doable, but the cost is non-trivial.
4. **Sibling repo `meos-pg-runtime` as an explicit curated fork**, consumed via submodule. This honestly represents what the code is — we're already running our own port of PG's runtime, the patches prove it. Patch history becomes commit history. PG-version bumps become reviewable merges instead of patch-rebase batches. (Precedent in shape, though not detail: pgvector, Citus, TimescaleDB, Neon.)

I'm leaning **(4) as the right long-term answer** with **(2) as a sensible interim step** if (4) is too big a commitment to take on in the same window as everything else in flight. **(3) I'd avoid** unless we explicitly accept the distro / bindings cost — but if you've already been thinking through that fallout and have a story for it, I want to hear it before I rule it out.

**Proposal for this PR**

I'd suggest we merge #751 as-is — the strategic posture isn't decided by where these bytes live today, and #751 is the gate that unlocks the whole pgtypes API surface. Then I'll open a tracking issue for "pgtypes vendored-code architecture" with these options spelled out, and we can decide there with proper room to breathe rather than under PR review. Happy to co-draft that issue with you if you'd like.

Does this framing match what you were getting at? If you've already been thinking specifically of (2) or (3), I'd love to hear which — that's a meaningful signal.

meos/examples/rtree_example.c
estebanzimanyi: Already addressed — the master-merge I pushed earlier today (commit `ae10434f2`) brought in the org-wide `2025` → `2026` copyright bump (PR #842 / commit `0e183c354`). The current file on the branch tip reads `Copyright (c) 2016-2026`.

estebanzimanyi:
@Davichet-e thank you for the review. Two parts to address:

**On the year typo (`rtree_example.c` line 4):** already in. The master-merge I pushed earlier today (commit `ae10434f2`) folded in the org-wide `2025` → `2026` bump from #842 / commit `0e183c354`. The branch tip now reads `Copyright (c) 2016-2026`.

**On the vendor-extraction proposal** — this is a substantive and legitimate architectural improvement. I want to engage with it carefully rather than reject or accept on the spot.

The proposal's strengths:
- ~400 files removed from the repo → meaningfully smaller clones, cleaner GitHub UX, faster shallow checkouts.
- The pinned PG version becomes explicit in `PG_VERSION` and the diffs become legible `patches/*.patch` files instead of a fork-and-edit history.
- Provenance is auditable: anyone can rebuild `pgtypes/` from the pinned tag + the patches.

The trade-offs to weigh:
- **CI / offline builds**: network fetch from PG source on every fresh build. Cacheable, but adds a moving piece (CI cache invalidation, mirror fallback).
- **Patch maintenance**: every time we touch one of these files we instead generate / update a patch. That's how distributions work; it's a real workflow cost vs. the current direct-edit pattern.
- **PG version bumps**: a future PG-19 bump becomes 'pin tag + regenerate patches' instead of 'sync new files'. Probably better, but requires the patches to be small and clean.
- **Implementation cost**: substantial. The skeleton you sketched (`extract.cmake` + `MANIFEST.txt` + `patches/`) is mostly straightforward, but distilling the current pgtypes diff into per-concern patches is the part that takes real time — probably 1-2 weeks of focused work plus a CI iteration loop.

**My proposal for sequencing**:
- Land this PR as-is (with the vendored files), since it's already merge-ready and the PG-18 unblock is itself the goal.
- Open a follow-up tracking issue capturing the extraction proposal — the skeleton, the patch breakdown, the CI implications — so it gets evaluated in its own right rather than blocking this PR.
- I'm happy to file that issue with credit to your sketch and Cc you for review.

Does that sequencing work for you? Or do you see a path to land both at once that I'm missing?

estebanzimanyi:
Rolling back the merge (`ae10434f2`) and TLS port (`1d0c4a393`) I pushed earlier. The branch is reset to `cf476d450` (the last green-CI state).

**Why the rollback:** the merge passed local CMake configure + full MEOS build, but in CI it caused runtime regressions. Concretely, the first failing test (`001_set`) crashes the Postgres backend on a `textset || text` concat — server-process termination, not a test-output diff. CI then cascades into "database system is in recovery mode" on every subsequent test.

The pre-merge commit `cf476d450` had all CI green; the post-merge `1d0c4a393` had Main Build / Ubuntu-clang / Windows / Test MEOS compilation all failing. Clear regression introduced somewhere in the conflict resolution.

**Implication:** the merge of upstream/master into this branch needs more care than a textual conflict resolution + local-build smoke test caught. Likely candidates for the runtime regression:
- One of the auto-merged temporal/textset source files (the auto-merge picked up changes from #815 + others without my eyes on each).
- A subtle interaction between the branch's `pgtypes/` restructure and #815's MEOS_TLS additions when running as a PG extension (vs. standalone MEOS).

The branch is now back to `DIRTY` (5 commits behind master) but its CI goes green again. Recommend a fresh attempt at the merge with deliberate review of each auto-merged file in `meos/src/temporal/` and `meos/src/cbuffer/` rather than treating them as no-touch auto-merges.

PR #757, opened 2026-03-05
Uniformize return -1 on error on distance functions

estebanzimanyi:


estebanzimanyi:
I close this PR and replace it by PR #769 where the option taken was to use systematically DBL_MAX for both errors and empty geometries.

Issue #404, opened 2023-12-06
Making MEOS multithreading


estebanzimanyi:
MEOS is the underlying C library driving MobilityDB.
https://libmeos.org/
MEOS corresponds to the C library GEOS for trajectory management.
https://libgeos.org/

Bindings for MEOS in several programming languages are being developped
* Python  https://github.com/MobilityDB/PyMEOS
* Java: https://github.com/MobilityDB/JMEOS
* Javascript: https://github.com/MobilityDB/MEOS.js
* C#: https://github.com/MobilityDB/MEOS.NET

MEOS currently does not support multithreading. On the other hand GEOS supports multithreading
https://libgeos.org/usage/c_api/

We are looking for contributors that can help us to develop this feature. The solution should follow a similar approach as the one for GEOS.

PR #762, opened 2026-03-21
Toward making MEOS thread safe

estebanzimanyi:
First step to make MEOS library thread safe
See Issue #404


Davichet-e (commented):
I think it looks good, happy to see this change happening. I only have a couple of small questions

estebanzimanyi:
Superseded by #815, which delivers a fresh, complete TLS-only thread-safety implementation for issue #404 (3 focused commits, no public-API change, full pthread stress test). #762's earlier mutex-style approach (13 commits) is replaced by that newer work.

PR #765, opened 2026-03-23
Fix inverted geodetic flag in geoset_make

Davichet-e:
### Bug fix: `geoset_make` inverted geodetic flag

`geoset_make` in `geoset_meos.c` had an inverted ternary for determining the set type.

`FLAGS_GET_GEODETIC` is truthy when the geodetic flag is set, so the true branch must map to `T_GEOGRAPHY`. The inverted logic caused non-geodetic geometries to always produce a `geogset`, triggering a "mixed temporal and set types" error when passed to `temporal_at_values` with a `tgeompoint`. The other two identical patterns in the same file already had the correct order, consistent with a copy-paste mistake.

### DIFF
--- a/meos/src/geo/geoset_meos.c
+++ b/meos/src/geo/geoset_meos.c
@@ -113,7 +113,7 @@ geoset_make(GSERIALIZED **values, int count)
   for (int i = 0; i < count; ++i)
     datums[i] = PointerGetDatum(values[i]);
   meosType geotype = FLAGS_GET_GEODETIC(values[0]->gflags) ?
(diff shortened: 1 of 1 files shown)


mschoema (approved):

mschoema:
Thank you for the patch.

Btw, I will delete the fix/feature branches by default after merging the PRs, if the branches were made on this repo. If you want to keep one in the future for some reason (e.g. working on linked PRs or something similar), let me know in the PR message, or simply create the PR from a branch on your fork, so you maintain control of the branches.

PR #768, opened 2026-03-23
Enable using MeosArray beyond parsing

estebanzimanyi:


meos/src/temporal/meos_array.c
```diff
@@ -0,0 +1,150 @@
...
+  array->count++;
+  return;
+}
+
+/**
+ * @brief Get the n-th Datum value of the array (0-based) 
+ * @note Function used to access Datum stored in the array
+ */
+Datum
+meos_array_get_n(const MeosArray *array, int n)
```
mschoema, line 114: Since `meos_array_add` takes a `void *`, it seems like it would make more sense to also return a `void *` here instead of a `Datum`. Also, if we store objects with an `elem_size > sizeof(Datum)`, we would be truncating data here, like in the case of Span. This is currently not visible because `spanset_parse` never calls `meos_array_get_n`, but instead it accesses `array->elems` directly.

meos/src/temporal/meos_catalog.c
```diff
@@ -625,16 +625,23 @@ basetype_varlength(meosType type)
...
 }
 
+/**
+ * @brief Return the length of a base type
+ * @return On error return SHRT_MAX
+ */
 /**
  * @brief Return the length of a base type
  * @return On error return SHRT_MAX
  */
```
mschoema, line 631: Duplicate comment blocks

meos/include/meos.h
```diff
@@ -259,6 +259,13 @@ typedef struct SkipList SkipList;
 
 /*****************************************************************************/
 
+/* Structure of an expandable arrays used in particular to avoid parsing twice
+ * a MEOS value input in text format */
+
+typedef struct MeosArray MeosArray;
+
+/*****************************************************************************/
+
```
mschoema: We probably don't want to expose this in the public API, unless it's used in the signature of some API functions. For the moment I think it's enough to have it in `meos_internal.h`

mschoema (commented):
I have a few more comments related to this PR. I think we can clean up the array API. Should I just push a commit to this branch and you tell me what you think? I can always revert the commit later if you don't like it.

estebanzimanyi:
> I have a few more comments related to this PR. I think we can clean up the array API. Should I just push a commit to this branch and you tell me what you think? I can always revert the commit later if you don't like it.

Please put directly your modifications in my branch. It will save time for everyone. Thanks !

meos/include/meos.h
estebanzimanyi: In the same way that we kept the opaque data structure SkipList in meos.h I would also leave the opaque data structure MeosArray in meos.h. I could see some user code doing some manipulation such as an intersection with arbitrary geometries for which we cannot predict the number of results using this data structure in C, which is a light version of the Vector data structure in C++ STL.

meos/src/temporal/meos_array.c
estebanzimanyi: I left explicitly this API returning Datum to avoid changing numerous calls in the current MobilityDB code base. The rationale is that we extended the previous version of the API which correctly worked with Datum and added the possibility to work with C fixed-sized structs. However, since you will be working on my branch please propose your API and we will collaborate on that branch.

meos/src/temporal/meos_catalog.c
estebanzimanyi: Corrected

mschoema (approved):

PR #769, opened 2026-03-23
Uniformize distance functions to return DBL_MAX on error or empty geo…

estebanzimanyi:
…metries

### DIFF
--- a/meos/src/geo/postgis_funcs.c
+++ b/meos/src/geo/postgis_funcs.c
@@ -860,11 +860,7 @@ geom_distance2d(const GSERIALIZED *gs1, const GSERIALIZED *gs2)
   double mindist = lwgeom_mindistance2d(geom1, geom2);
   lwgeom_free(geom1);
   lwgeom_free(geom2);
-  /* if called with empty geometries the ingoing mindistance is untouched,
-   * and makes us return NULL */
-  if (mindist < FLT_MAX)
-    return mindist;
-  return DBL_MAX;
+  return mindist;
 }
 
 /**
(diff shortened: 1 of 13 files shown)


postgres/utils/datetime.c
mschoema, line 1: Can you rebase on master to remove this change?

meos/src/geo/postgis_funcs.c
```diff
@@ -863,14 +864,15 @@ geom_distance2d(const GSERIALIZED *gs1, const GSERIALIZED *gs2)
    * and makes us return NULL */
   if (mindist < FLT_MAX)
     return mindist;
-  return -1;
+  return DBL_MAX;
```
mschoema: Why do we use `FLT_MAX` above but `DBL_MAX` below?

postgres/utils/datetime.c
estebanzimanyi: Done

meos/src/geo/postgis_funcs.c
estebanzimanyi: Thanks for point this out. This was inherited from a previous PostGIS version. I have corrected all of them.

postgres/utils/datetime.c
mschoema: I think there might have been an issue. It seems like you merged the new commits from master on top of your commit, instead of rebasing. I'm seeing 6 commits in this PR instead of 2 (the initial one, and the new fix one)

estebanzimanyi: I thought I made a rebase .... Sorry if I made a mistake ... By squashing the PR you may clear my mistake.

mschoema (approved):

postgres/utils/datetime.c
estebanzimanyi: I have verified and indeed I made a rebase as the first step before implementing the changes requested in the PR

mschoema: Indeed, during merging, git correctly removed the changes that were already present on master

PR #770, opened 2026-03-25
Add MEOS implementation of ST_MinimumBoundingCircle

estebanzimanyi:
Using the newer version of the MeosArray data structure, we obtain the
following timings for the MEOS and the PostGIS versions of the function.

SELECT cbuffer(geom) FROM NaturalAreas ;
[...]
(283 rows)

Time: 253.216 ms

SELECT ST_MinimumBoundingCircle(geom) FROM NaturalAreas ;
[...]
(283 rows)

Time: 94257.473 ms (01:34.257)

### DIFF
--- a/meos/src/geo/postgis_funcs.c
+++ b/meos/src/geo/postgis_funcs.c
@@ -4012,53 +4012,6 @@ line_substring(const GSERIALIZED *gs, double from, double to)
   return result;
 }
 
-/**
- * @ingroup meos_geo_base_spatial
- * @brief Return the center point and radius of the smallest circle that
- * contains a geometry
- * @param[in] geom Geometry
- * @param[out] radius Radius
- */
-// GSERIALIZED *
-// geom_min_bounding_radius(const GSERIALIZED *geom, double *radius)
-// {
-  // if (! geom)
-    // return NULL;
-
-  // LWGEOM *lwcenter = NULL;
-  // /* Empty geometry?  Return POINT EMPTY with zero radius */
-  // if (gserialized_is_empty(geom))
-  // {
-    // lwcenter = (LWGEOM*) lwpoint_construct_empty(
-      // gserialized_get_srid(geom), LW_FALSE, LW_FALSE);
-  // }
-  // else
-  // {
-    // LWGEOM *input = lwgeom_from_gserialized(geom);
-    // LWBOUNDINGCIRCLE *mbc = lwgeom_calculate_mbc(input);
-
-    // if (!(mbc && mbc->center))
-    // {
-      // meos_error(ERROR, MEOS_ERR_INTERNAL_ERROR,
-        // "Error calculating minimum bounding circle");
-      // lwgeom_free(input);
-      // return NULL;
-    // }
-
-    // lwcenter = (LWGEOM*) lwpoint_make2d(input->srid, mbc->center->x,
-      // mbc->center->y);
-    // *radius = mbc->radius;
-
-    // lwboundingcircle_destroy(mbc);
-    // lwgeom_free(input);
-  // }
-
-  // GSERIALIZED *result = geo_serialize(lwcenter);
-  // lwgeom_free(lwcenter);
-
-  // return result;
-// }
-
 /*****************************************************************************
  * Minimum Enclosing Circle implementation improving the performance of the
  * PostGIS function ST_MinimumBoundingCircle
@@ -4276,11 +4229,9 @@ lwgeom_mec_supported_type(const LWGEOM *geom)
   else if (lwgeom_is_collection(geom))
   {
     const LWCOLLECTION *col = (LWCOLLECTION *) geom;
-    bool result = true;
     for (int i = 0; i < (int) col->ngeoms; i++)
     {
-      result |= lwgeom_mec_supported_type(col->geoms[i]);
-      if (! result)
+      if (! lwgeom_mec_supported_type(col->geoms[i]))
         return false;
     }
     return true;
@@ -4326,7 +4277,7 @@ geom_min_bounding_radius(const GSERIALIZED *geom, double *radius)
     center = (LWGEOM *) lwpoint_construct_empty(input->srid, LW_FALSE, LW_FALSE);
     *radius = 0;
   }
-  if (lwgeom_mec_supported_type(input))
+  else if (lwgeom_mec_supported_type(input))
   {
     Circle c = lwgeom_mec(input);
     center = (LWGEOM *) lwpoint_make2d(input->srid, c.center.x, c.center.y);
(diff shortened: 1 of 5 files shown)


mschoema:
I assume this should be merged after #768 ? I'll take a look at that PR some time later.

Concerning this PR in particular, what are the differences in terms of features between the 2 functions? I suspect that there is a reason why the PostGIS code is so much slower. If not, should this be suggested as an optimization to PostGIS directly?

estebanzimanyi:
Indeed I was planing to have this optimization in master to test it thoroughly and propose an improvement to PostGIS. Since we both use the same algorithm from Welzl the only reason I can think of for this speed up is with respect to the optimized memory allocation in MEOS.

mschoema:
I think you're gonna have to cherry-pick commit 93d4e9d83b6bae3b5862fca4485d0d782541e756 on top of master to fix the current merge conflicts.

estebanzimanyi:
I have rebased the PR to solve the conflicts

mschoema:
I cherry-picked the commit to remove the 'meos array' commit that was still here but also already merged into master.

mschoema:
What is the state of this PR? It seems like these changes are also present in some newer PRs.

meos/src/geo/postgis_funcs.c
```diff
@@ -4019,43 +4019,338 @@ line_substring(const GSERIALIZED *gs, double from, double to)
...
+
+    // lwboundingcircle_destroy(mbc);
+    // lwgeom_free(input);
+  // }
+
+  // GSERIALIZED *result = geo_serialize(lwcenter);
+  // lwgeom_free(lwcenter);
+
+  // return result;
+// }
```
mschoema: Should we just remove this instead of commenting it away?

meos/src/geo/postgis_funcs.c
```diff
@@ -4019,43 +4019,338 @@ line_substring(const GSERIALIZED *gs, double from, double to)
...
+  LWGEOM *input = lwgeom_from_gserialized(geom);
+  LWGEOM *center;
+  
+  if (lwgeom_is_empty(input))
+  {
+    center = (LWGEOM *) lwpoint_construct_empty(input->srid, LW_FALSE, LW_FALSE);
+    *radius = 0;
+  }
+  if (lwgeom_mec_supported_type(input))
   {
```
mschoema, line 4283: Shouldn't this be an `else if`?

estebanzimanyi:
> What is the state of this PR? It seems like these changes are also present in some newer PRs.

This PR should be merged first. The subsequent PRs that needed this functionaly already embed it so when rebasing these PRs you will be able to ONLY see the specific features brought by each PR.

meos/src/geo/postgis_funcs.c
```diff
@@ -4019,43 +4019,338 @@ line_substring(const GSERIALIZED *gs, double from, double to)
...
+  else if (lwgeom_is_collection(geom))
+  {
+    const LWCOLLECTION *col = (LWCOLLECTION *) geom;
+    bool result = true;
+    for (int i = 0; i < (int) col->ngeoms; i++)
+    {
+      result |= lwgeom_mec_supported_type(col->geoms[i]);
+      if (! result)
+        return false;
+    }
```
mschoema, line 4247: With `result = true`, `result |= ...` will always return true too, so this is wrong.
It should just be:

```
if (! lwgeom_mec_supported_type(col->geoms[i]))
      return false;
```

and the `result` variable is not needed.

postgres/utils/datetime.c
mschoema, line 1: This feels like it should be part of another PR. Wasn't this fixed somewhere else already?

meos/src/geo/postgis_funcs.c
estebanzimanyi: Removed in the next commit

meos/src/geo/postgis_funcs.c
estebanzimanyi: Solved in the next commit

meos/src/geo/postgis_funcs.c
estebanzimanyi: Done in the next commit

postgres/utils/datetime.c
estebanzimanyi: The initial PR was waiting for the fix in another PR. Now that the PR solving this was merged into master, this is solved in the next commit that rebased the current PR and corrected the issues you mentioned above.

mschoema:
Can you take a look at [039cca0](https://github.com/MobilityDB/MobilityDB/pull/770/commits/039cca0c4d0135d532ddad05f5d9da757ecc0ff0) and let me know if this improves the query on your dataset?

I've also added a commit that adds some tests to avoid the coverage regressions, as the new code wasn't being tested.

Let me know if this looks good to you.

estebanzimanyi:
> Let me know if this looks good to you.

Amazing work, now it works lightning fast ! 

```
tcbuffer=# SELECT COUNT(*) FROM NaturalAreas WHERE cbuffer(Geom) IS NOT NULL;
 count
-------
   283
(1 row)

Time: 78.214 ms
tcbuffer=# SELECT SUM(ST_NPoints(Geom)) FROM NaturalAreas;
   sum
---------
 1972734
(1 row)

Time: 21.976 ms
```

mschoema:
Never mind, the tests do not run with CBUFFER enabled, so the coverage still goes down. We should maybe change this, but this can be a future PR.

meos/src/geo/postgis_funcs.c
```diff
@@ -4012,50 +4012,307 @@ line_substring(const GSERIALIZED *gs, double from, double to)
...
+ *
+ * The following query used on natural areas around Denmark shows the
+ * difference between the MEOS and the PostGIS functions
+ * @code
+ *  SELECT Id, cbuffer(Geom) AS cb FROM NaturalAreas ORDER BY Id;
+ *  -- Time: 321.113 ms
+ *
+ * SELECT Id, ST_MinimumBoundingCircle(Geom) AS cb FROM NaturalAreas ORDER BY Id;
+ *  -- Time: 89308.406 ms (01:29.308)
+ * @endcode
```
mschoema: Maybe we shouldn't put timings in the code comments. Can you write me a short commit description with this instead that I can use when merging the PR (with the new timings of the 3-loop function), and remove or simplify the comment from the code?

mschoema (changes requested):

meos/src/geo/postgis_funcs.c
estebanzimanyi: Removed the comments. In the comment of this PR the timings are given

mschoema (approved):

PR #772, opened 2026-03-26
Add search operation support to R-tree (overlaps, contains, contained_by)

mschoema:
Add RTreeSearchOp enum to the public API with three operations following the GiST naming convention (from the stored box's perspective):
- RTREE_OVERLAPS: find stored boxes that overlap the query
- RTREE_CONTAINS: find stored boxes that contain the query (@>)
- RTREE_CONTAINED_BY: find stored boxes contained by the query (<@)

Refactor node_search following the GiST pattern with separate leaf_consistent and inner_consistent functions:
- leaf_consistent applies the exact predicate at leaf nodes
- inner_consistent applies a looser pruning check at inner nodes (e.g., for CONTAINS the inner node bbox must contain the query, while for OVERLAPS/CONTAINED_BY a simple overlap check suffices)

The search op is placed before the query pointer in the signature so that rtree_search(rtree, RTREE_CONTAINS, box, &count) reads like the SQL pattern: column @> value.

Also make node_search static (only used within the file).

### DIFF
--- a/meos/examples/rtree_example.c
+++ b/meos/examples/rtree_example.c
@@ -181,7 +181,7 @@ int main()
 #endif /* BBOX_TYPE */
 
   t = clock();
-  int *ids = rtree_search(rtree, box, &count);
+  int *ids = rtree_search(rtree, RTREE_OVERLAPS, box, &count);
   t = clock() - t;
   time_taken = ((double) t) / CLOCKS_PER_SEC; // in seconds 
   printf("Index lookup took %f seconds to execute \n", time_taken);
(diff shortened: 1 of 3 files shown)


estebanzimanyi (approved):

estebanzimanyi:
Thanks for this great optimizations ! As you may know we have been using this implementation of R-Tree in MobilityDuck which was presented on EDBT this week https://mobilitydb.com/publications.html

mschoema:
Yes, I was meaning to get around and improving the rtree code for a while now, which I finally did. I have 2 more PRs coming up.

PR #774, opened 2026-03-26
Clean up rtree_example

mschoema:
Rewrite the example to run all four bounding box types (floatspan, tstzspan, tbox, stbox) sequentially at runtime instead of using compile-time #if/#elif blocks that required recompilation to switch types.

Each type has a self-contained test function, and a shared verify_search helper compares R-tree results against brute-force overlap scans.

The example now only uses public API headers (meos.h, meos_geo.h) instead of including meos_internal.h and meos_internal_geo.h.

### DIFF
--- a/meos/examples/rtree_example.c
+++ b/meos/examples/rtree_example.c
@@ -29,215 +29,254 @@
 
 /**
  * @file
- * @brief A simple program to use an RTree index for searching MEOS bounding
- * boxes, i.e., `floatspan`, `tstzspan`, `tbox`, and `stbox`
+ * @brief A simple program that demonstrates the RTree index for searching
+ * MEOS bounding boxes: floatspan, tstzspan, tbox, and stbox.
  *
- * The program can be build as follows
+ * The program runs all four bounding box types, inserting random boxes into
+ * the index and verifying search results against a brute-force scan.
+ *
+ * The program can be built as follows
  * @code
- * gcc -Wall -g -I/usr/local/include -o rtree_example rtree_example.c -L/usr/lib -lproj -L/usr/local/lib -lmeos
+ * gcc -Wall -g -I/usr/local/include -o rtree_example rtree_example.c -L/usr/local/lib -lmeos -lproj
  * @endcode
  */
 
 /* C */
 #include <stdio.h>
 #include <stdlib.h>
-#include <time.h>
 #include <string.h>
+#include <time.h>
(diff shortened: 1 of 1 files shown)


estebanzimanyi (approved):

mschoema:
Could you run the new file locally from my branch and verify that everything works correctly on your end?

mschoema:
If it's easier I could just merge the PR so you can test on the master branch. And we can fix any issues you encounter in a new PR. Let me know what you prefer

estebanzimanyi:
I cloned your branch locally and execute the program without any problem

$ ./rtree_example
Index lookup took 0.000331 seconds to execute
Brute foce took 0.000280 seconds to execute

BOUNDING BOX = STBOX
EXPECTED HITS = 1876
INDEX HITS    = 1876
$

JashanReel:
Hello! I'm working on JMEOS (the Java binding for MEOS) and have been testing the RTree changes after Esteban suggested me to re-run one of my Jupyter notebooks focused on RTree performance to observe the new optimizations.

The speed improvement is very noticeable: RTree queries on 2000 regions went from ~2000 ms down to ~270 ms !

However, I think I noticed a problem if I'm not mistaken: rtree_search always returns count = 0, even when querying with a giant bounding box that covers the entire dataset. The returned pointer is non-null, but the count is always 0 while Brute force on the same data returns 279,453 matches on the other hand.

Here's the Jupyter notebook where I tested the RTree changes:
https://github.com/MobilityDB/JMEOS/blob/fix-tests-using-docker/src/main/java/examples/Jupyter_BerlinMod_Brussels.ipynb

If needed, here's a markdown document explaining how to run it:
https://github.com/MobilityDB/JMEOS/blob/fix-tests-using-docker/src/main/java/examples/JUPYTER_NOTEBOOKS_GUIDE.md (and my branch providing an updated Dockerfile to run JMEOS and build your fat jar : https://github.com/MobilityDB/JMEOS/tree/fix-tests-using-docker)

I'd be happy to provide more details or test any fix.
Thank you!

mschoema:
In the notebook I still see the old function call: `Pointer rtree_search(Pointer rtree, Pointer box, Pointer countPtr);`, whereas PR #772 added a new parameter to the function to specify if we are doing overlaps, contains or contained by search. Are you certain to be using the latest version of the function, passing all 4 parameters?

PR #776, opened 2026-04-02
Polygon clip

estebanzimanyi:
This PR provides a MEOS implementation of the clipping algorithms for a temporal geometric point with respect to a geometry composed of (collections of) points, lines, and polygons with holes and islands in the holes (recursively). In other words, circular strings (and geometries derived from them) and polyhedral surfaces are not supported. Achieving this required to implement robust computational geometry algorithms that are expensive to compute in GEOS. An in-memory R-tree index is used to avoid computing the intersection of all trajectory segments with all geometry edges. 

I tested this with one day of AIS data and a set of geometries representing protected natural areas in the waters around Denmark. Although there are regressions that must still be solved, I wanted to discuss the approach to see whether other performance gains can be achieved.

```sql
tcbuffer=# SELECT COUNT(*) FROM VesselTrip;
 count
-------
  2111
(1 row)

tcbuffer=# SELECT COUNT(*) FROM NaturalAreas;
 count
-------
   283
(1 row)

tcbuffer=# SELECT MIN(numInstants(Trip)),  MAX(numInstants(Trip)), AVG(numInstants(Trip)) FROM VesselTrip;
 min |  max  |          avg
-----+-------+-----------------------
   1 | 27006 | 2158.0672666982472762
(1 row)

tcbuffer=# SELECT MIN(ST_NPoints(Geom)),  MAX(ST_NPoints(Geom)), AVG(ST_NPoints(Geom)) FROM NaturalAreas;
 min |  max  |          avg
-----+-------+-----------------------
   4 | 67907 | 6970.7915194346289753
(1 row)
```

Comparing with the current implementation in master with this PR shows the performance gain.

```sql
-- This PR
tcbuffer=# SELECT COUNT(*) FROM VesselTrip t1, NaturalAreas t2 WHERE atGeometry(Trip, geom) IS NOT NULL;
 count
-------
  1095
(1 row)

Time: 244175.800 ms (04:04.176)

-- Current implementation in master
tcbuffer=# SELECT COUNT(*) FROM VesselTrip t1, NaturalAreas t2 WHERE atGeometry(Trip, geom) IS NOT NULL;
 count
-------
   877
(1 row)

Time: 844088.568 ms (14:04.089)
```

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,17 +1,23 @@
 ---
+# Codacy's cppcheck wrapper runs the tool without the build's -I include
+# paths, so every project/system header fires a 'missingInclude[System]'
+# warning - on this codebase ~200 of them drown the real findings.
+# MobilityDB runs its own cppcheck in .github/workflows/cppcheck.yml
+# instead: it feeds cppcheck the real compile_commands.json (so include
+# resolution works) and publishes SARIF to GitHub Code Scanning (inline
+# PR annotations, Security tab). Turn Codacy's copy off.
 engines:
   cppcheck:
-    language: c
-    exclude_paths:
-      - "postgis/**"
-      - "postgres/**"
+    enabled: false
   duplication:
     exclude_paths:
       - "postgis/**"
       - "postgres/**"
+      - "pgtypes/**"
     config:
       languages:
         - "c"
 exclude_paths:
   - "postgis/**"
   - "postgres/**"
+  - "pgtypes/**"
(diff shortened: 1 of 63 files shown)


estebanzimanyi:
Superseded by branch `clip-clipper2-prod` on the same fork, which abandons the Martinez-Rueda port and replaces the polygon-Boolean engine with vendored Clipper2 v2.0.1 (Boolean-clipping subset, BSL-1.0). The replacement also bundles an open-path trajectory clip on the same Clipper2 install. Opening a fresh PR with that content; this thread can be archived.

PR #781, opened 2026-04-10
Change rtree API

estebanzimanyi:
This PR changes the API of the MEOS rtree implementation. In the previous API the following function
```c
extern int *rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int *count);
```
performed an `malloc` to create the integer array holding the results of the search. In the new version of the function
```c
extern int rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int **ids);
```
the caller is responsible to pass to the search function in the last argument a buffer array on which the results of the search are written.

This enhances the performance of the rtree by comparing at a high-level viewpoint the previous usage of the function
```
<create the rtree for N boxes>
loop
  <input a value to search>
  int count;
  int *ids = rtree_search(rtree, op, bbox(value), &count);
  <process the input value>
  free(ids);
end loop
```
with the current one
```
<create the rtree for N boxes>
int *ids = malloc(sizeof(int) * N);
loop
  <input a value to search>
  int count = rtree_search(rtree, op, bbox(value), &ids);
  <process the input value>
end loop
free(ids)
```

### DIFF
--- a/meos/examples/rtree_example.c
+++ b/meos/examples/rtree_example.c
@@ -79,8 +79,9 @@ verify_search(const char *name, const RTree *rtree, const void *query,
 {
   /* Index search */
   clock_t t = clock();
-  int index_count;
-  int *ids = rtree_search(rtree, RTREE_OVERLAPS, query, &index_count);
+  int *ids = malloc(sizeof(int) * NO_BBOX);
+  memset(ids, 0, sizeof(int) * NO_BBOX);
+  int index_count = rtree_search(rtree, RTREE_OVERLAPS, query, &ids);
   double index_time = (double)(clock() - t) / CLOCKS_PER_SEC;
 
   /* Brute-force search */
@@ -109,8 +110,8 @@ verify_search(const char *name, const RTree *rtree, const void *query,
       errors++;
   }
 
-  printf("  %-10s  index: %.6fs (%d hits)  brute: %.6fs (%d hits)  %s\n",
-    name, index_time, index_count, brute_time, brute_count,
+  printf("  %-10s  index: %.6fs (%d hits)  brute: %.6fs (%d hits) ratio: %.6f %s\n",
+    name, index_time, index_count, brute_time, brute_count, index_time / brute_time,
     errors == 0 ? "OK" : "MISMATCH");
 
   free(actual);
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -286,8 +286,8 @@ extern RTree *rtree_create_stbox();
 extern void rtree_free(RTree *rtree);
 extern void rtree_insert(RTree *rtree, void *box, int id);
 extern void rtree_insert_temporal(RTree *rtree, const Temporal *temp, int id);
-extern int *rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int *count);
-extern int *rtree_search_temporal(const RTree *rtree, RTreeSearchOp op, const Temporal *temp, int *count);
+extern int rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int **ids);
+extern int rtree_search_temporal(const RTree *rtree, RTreeSearchOp op, const Temporal *temp, int **ids);
 
 /*****************************************************************************
  * Error codes
--- a/meos/include/temporal/temporal_rtree.h
+++ b/meos/include/temporal/temporal_rtree.h
@@ -85,8 +85,9 @@ typedef struct RTreeNode
 struct RTree
 {
   size_t bboxsize;       /**< Size of the bouding box */
+  int count;             /**< Number of dimensions of the tree */
+  int dims;              /**< Number of dimensions of the tree */
   meosType bboxtype;     /**< Type of the bouding box */
-  int dims;
   RTreeNode *root;
   double (*get_axis)(const void *, int, bool);
   void (*bbox_expand)(const void *, void *);
--- a/meos/include/temporal/temporal_rtree.h
+++ b/meos/include/temporal/temporal_rtree.h
@@ -85,7 +85,7 @@ typedef struct RTreeNode
 struct RTree
 {
   size_t bboxsize;       /**< Size of the bouding box */
-  int count;             /**< Number of dimensions of the tree */
+  int count;             /**< Number of boxes in the tree */
   int dims;              /**< Number of dimensions of the tree */
   meosType bboxtype;     /**< Type of the bouding box */
   RTreeNode *root;
--- a/meos/src/temporal/meos_catalog.c
+++ b/meos/src/temporal/meos_catalog.c
@@ -928,7 +928,7 @@ span_canon_basetype(meosType type)
 /**
  * @brief Return true if the type is a span type
  */
-bool
+inline bool
 span_type(meosType type)
 {
   return (type == T_TSTZSPAN || type == T_DATESPAN || type == T_INTSPAN ||
--- a/meos/src/temporal/temporal_boxops.c
+++ b/meos/src/temporal/temporal_boxops.c
@@ -69,7 +69,7 @@
 bool
 bbox_type(meosType bboxtype)
 {
-  if (bboxtype == T_TSTZSPAN || bboxtype == T_TBOX || bboxtype == T_STBOX)
+  if (span_type(bboxtype) || bboxtype == T_TBOX || bboxtype == T_STBOX)
     return true;
   return false;
 }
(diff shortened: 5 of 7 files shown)


meos/src/temporal/temporal_rtree.c
```diff
@@ -678,10 +678,10 @@ is_power_of_two(const int n)
...
 add_answer(const int id, int **ids, int *count)
 {
-  /* Every power of two that exceeds the size of the array must be resized to
-   * double the current size */
-  if (*count >= SEARCH_ARRAY_STARTING_SIZE && is_power_of_two(*count))
-    *ids = repalloc(*ids, sizeof(int) * (*count) * 2);
+  // /* Every power of two that exceeds the size of the array must be resized to
+   // * double the current size */
+  // if (*count >= SEARCH_ARRAY_STARTING_SIZE && is_power_of_two(*count))
+    // *ids = repalloc(*ids, sizeof(int) * (*count) * 2);
```
Matematikoi: This looks like dead code. so you wouldn't be needing it.  You could also remove `is_power_of_two`


which puts the responsability of choosing a right size for the `ids` to the user. the user doesn't know from the start how many hits he is going to get, so he doesn't know the correct size.

meos/include/meos.h
```diff
@@ -286,8 +286,8 @@ extern RTree *rtree_create_stbox();
 extern void rtree_free(RTree *rtree);
 extern void rtree_insert(RTree *rtree, void *box, int id);
 extern void rtree_insert_temporal(RTree *rtree, const Temporal *temp, int id);
-extern int *rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int *count);
-extern int *rtree_search_temporal(const RTree *rtree, RTreeSearchOp op, const Temporal *temp, int *count);
+extern int rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int **ids);
+extern int rtree_search_temporal(const RTree *rtree, RTreeSearchOp op, const Temporal *temp, int **ids);
```
Matematikoi, line 290:
```suggestion
extern int rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, int *ids);
extern int rtree_search_temporal(const RTree *rtree, RTreeSearchOp op, const Temporal *temp, int *ids);
```

if we are not changing the size of the array, this can stay `*ids`

Matematikoi (commented):
My only question is how does the user avoid sending a smaller than necessary array, and ending up with a segmentation fault.

estebanzimanyi:
> My only question is how does the user avoid sending a smaller than necessary array, and ending up with a segmentation fault.

I have used the index intensively in the line clipping algorithms in the following file
https://github.com/estebanzimanyi/MobilityDB/blob/tinterrel_tpoint/meos/src/geo/tpoint_geom_clip.c
As you see in the corresponding PR
https://github.com/MobilityDB/MobilityDB/pull/776
the largest of the geometries contained in the table NaturalAreas have 67907 vertices.

By definition, the user should have an array of the same size of the index. I envisioned to keep the number of boxes currently in the index in a variable in the struct
https://github.com/estebanzimanyi/MobilityDB/blob/tinterrel_tpoint/meos/include/temporal/temporal_rtree.h#L85-L96
besides the `int dims` declaration. 

I have pushed a commit in this regard. What do you think ?

meos/include/temporal/temporal_rtree.h
```diff
@@ -85,8 +85,9 @@ typedef struct RTreeNode
 struct RTree
 {
   size_t bboxsize;       /**< Size of the bouding box */
+  int count;             /**< Number of dimensions of the tree */
+  int dims;              /**< Number of dimensions of the tree */
```
Matematikoi:
```suggestion
  int count;             /**< Number of elements in the tree */
  int dims;              /**< Number of dimensions of the tree */
```

Matematikoi (approved):
> By definition, the user should have an array of the same size of the index

Then for me its ok. LGTM. Just tiny typo

mschoema:
I'm not opposed to changing the API to avoid a the malloc, as it can indeed make things slow if we want to search in a loop. But I don't like the idea of having to pass a pointer to an unknown-sized array to the function. This seems to me like the perfect place to use the new MeosArray:

```c
// API
int rtree_search(const RTree *rtree, RTreeSearchOp op, const void *query, MeosArray *ids);

// Usage
MeosArray *ids = meos_array_init(sizeof(int));
for (...)
{
  int count = rtree_search(rtree, op, query, ids);
  for (int i = 0; i < count; i++)
  {
    int id = *(int *) meos_array_get_n(ids, i);
    // ... 
  }
}
meos_array_destroy(ids, false);
```

But if we want something like that, we should first make a PR to make the MeosArray API public (move it to `meos.h`).
What do you think of this? If you agree, I can prepare the PR for MeosArray.

estebanzimanyi:
> But if we want something like that, we should first make a PR to make the MeosArray API public (move it to `meos.h`). What do you think of this? If you agree, I can prepare the PR for MeosArray.

That would be great ! Many thanks for this work !

mschoema:
Now that #786 is merged, should I modify this one directly, or open a new PR to make the new changes to the R-Tree API using MeosArray?

estebanzimanyi:
> Now that #786 is merged, should I modify this one directly, or open a new PR to make the new changes to the R-Tree API using MeosArray?

Indeed, this is the next PR to continue the work on the r-tree. Many thanks for your work !

mschoema:
Can we close this one? There are a few minor changes that have not been added to the new PR, but these are unrelated to the PR topic, so should maybe be fixed in another PR.

estebanzimanyi:
> Can we close this one? There are a few minor changes that have not been added to the new PR, but these are unrelated to the PR topic, so should maybe be fixed in another PR.

Since you have identified the "few minor changes" could you please send me a diff so I can make a new PR ? Otherwise if for you it is not so complex to do it, it would be better to avoid missing something from my side ...
I let you decide.

meos/examples/rtree_example.c
```diff
@@ -109,8 +110,8 @@ verify_search(const char *name, const RTree *rtree, const void *query,
       errors++;
   }
 
-  printf("  %-10s  index: %.6fs (%d hits)  brute: %.6fs (%d hits)  %s\n",
-    name, index_time, index_count, brute_time, brute_count,
+  printf("  %-10s  index: %.6fs (%d hits)  brute: %.6fs (%d hits) ratio: %.6f %s\n",
+    name, index_time, index_count, brute_time, brute_count, index_time / brute_time,
     errors == 0 ? "OK" : "MISMATCH");
```
mschoema, line 115: 1

meos/src/temporal/meos_catalog.c
```diff
@@ -928,7 +928,7 @@ span_canon_basetype(meosType type)
 /**
  * @brief Return true if the type is a span type
  */
-bool
+inline bool
 span_type(meosType type)
```
mschoema, line 932: 2

meos/src/temporal/temporal_boxops.c
```diff
@@ -69,7 +69,7 @@
 bool
 bbox_type(meosType bboxtype)
 {
-  if (bboxtype == T_TSTZSPAN || bboxtype == T_TBOX || bboxtype == T_STBOX)
+  if (span_type(bboxtype) || bboxtype == T_TBOX || bboxtype == T_STBOX)
     return true;
```
mschoema, line 73: 3

meos/test/tbl_temporal_value.c
mschoema, line 1: 4

mschoema (commented):
These are the 4 things that were in this PR but not in mine.

Also, my PR doesn't add the `rtree->count` attribute, as it's not needed anymore.

meos/examples/rtree_example.c
mschoema: I could add this change to my PR if you want.

meos/test/tbl_temporal_value.c
mschoema: This probably doesn't need to be committed, it's just enabling/disabling some lines for local testing.

meos/examples/rtree_example.c
estebanzimanyi: That would be great !

meos/src/temporal/temporal_boxops.c
estebanzimanyi: Indeed, so we can close this PR !

meos/examples/rtree_example.c
mschoema: Added to the new PR.

PR #782, opened 2026-04-11
Expose nearestApproachInstant and shortestLine for tcbuffer and cbuffer

estebanzimanyi:
Revises the distance operators and functions (|=|, <->, shortestLine, nearestApproachInstant) by replacing the per-instant trajectory walk with a centroid-anchored shortcut when one operand is a non-point geometry. Measured on one day of AIS data joined with protected-area polygons around Denmark: nearestApproachInstant goes from 8m29s to 17s, the <-> operator from 8m51s to 33s, with no change on |=| or shortestLine.


estebanzimanyi:
@mschoema — flagging an unrelated-but-adjacent inconsistency in `tsequence_derivative` since this PR is already touching that function (the `datum_distance` signature change at `temporal.c:2974`).

## Issue

MEOS's `tsequence_derivative` returns slope in **value/second**:

\`\`\`c
// temporal.c, line ~2974
derivative = ... / ((double)(inst2->t - inst1->t) / 1000000);
//                                                  ^^^^^^^
//                                                  microseconds → seconds
\`\`\`

But the MobilityDB SQL wrapper's regression tests expect **value/day**. Concretely:

\`\`\`sql
SELECT derivative(tfloat '[1@2000-01-01, 4@2000-01-04]');
-- MEOS direct (per-second):  Interp=Step;[0.000011574074074@..., 0.000011574074074@...]
-- MobilityDB regression:     {[1@2000-01-01, 1@2000-01-04]}
\`\`\`

The SQL wrapper presumably multiplies the MEOS result by 86400. MobilityDuck (which calls MEOS directly without that wrapper) currently gets the per-second form, which surfaced this divergence.

## Suggestion

Since this PR already revisits the math, would you consider one of:

1. **Change the unit in MEOS** (`/ 1000000` → `/ 86400000000`). Breaking for any direct MEOS caller that assumes per-second. Probably too invasive.
2. **Add a scale parameter**: `temporal_derivative(temp, double time_unit_seconds)` defaulting to 86400 to align with SQL semantics. Non-breaking with sensible default.
3. **Document the per-second contract** explicitly in the MEOS API doc, so downstream wrappers (MobilityDB SQL, MobilityDuck, others) know to scale themselves.

I'm happy to land a follow-up MEOS PR after this one merges — just want to make sure my fix doesn't conflict with what you have in flight. For now MobilityDuck is multiplying by 86400 in its own wrapper as a workaround.

estebanzimanyi:

estebanzimanyi:

estebanzimanyi:

PR #783, opened 2026-04-12
Fix Codacy issues

estebanzimanyi:
This PR simply fixes some issues reported by Codacy.


compile_commands.json
mschoema, line 1: This file should not be part of this PR.

Davichet-e (approved):

PR #786, opened 2026-04-15
Make MeosArray a public API in meos.h

mschoema:
- Move MeosArray function declarations from meos_internal.h to meos.h as an opaque type (struct definition stays internal).
- Rename for consistency: meos_array_init -> meos_array_create, meos_array_get_n -> meos_array_get.
- Add meos_array_count accessor.
- Split destroy/reset into plain and _free variants instead of a boolean parameter.


estebanzimanyi (approved):
Much nicer API !

PR #787, opened 2026-04-16
Use MeosArray in rtree_search API

mschoema:
Replace the internally-allocated int array with a caller-provided MeosArray, allowing reuse across multiple searches without repeated allocation. Remove is_power_of_two, add_answer, and SEARCH_ARRAY_STARTING_SIZE which are no longer needed.

### DIFF
--- a/meos/examples/rtree_example.c
+++ b/meos/examples/rtree_example.c
@@ -79,8 +79,8 @@ verify_search(const char *name, const RTree *rtree, const void *query,
 {
   /* Index search */
   clock_t t = clock();
-  int index_count;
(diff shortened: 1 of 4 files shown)


estebanzimanyi (dismissed):
Thanks ! It really improves things considerably !

mschoema:
This PR replaces #781 and uses MeosArray in the rtree API instead of expecting a fixed `int *` array from the user.

mschoema:
Also made some changes to the `rtree_example.c` file to better show the API functions and new functionality (different search operators, reuse of MeosArray, `_search_temporal`).

estebanzimanyi (approved):

PR #788, opened 2026-04-20
Add the tbigint type

estebanzimanyi:
This PR adds a new `tbigint` type that is similar to the traditional `tint` type but instead uses `int64` as base type.
Since a few years we added in MobilityDB the `bigintset`, `bigintspan` and `bigintspaset` to be able to represent OSM identifiers which are signed `int64` values. With the new type we are able, for example, to transform a trajectory on an urban setting such as
``` sql
SELECT tgeompoint '[Point(lon1,lat1)@tstz1, Point(lon2,lat2)@tstz2,  ... ]'
```
into the temporal evolution of the OSM identifiers that the vehicle traversed
``` sql
SELECT tbigint '[osm_id1@tstz1, osm_id2@tstz2,  ... ]'
```
We can also use this type for associating to the geospatial trajectory above a temporal h3 index
``` sql
SELECT tbigint '[h3index1@tstz1, h3index2@tstz2,  ... ]'
```
Notice that in `h3-pg`, the PostGIS extension for h3 (https://github.com/postgis/h3-pg) the `h3index` type is defined as follows
```
CREATE TYPE h3index (
  INPUT          = h3index_in,
  OUTPUT         = h3index_out,
  RECEIVE        = h3index_recv,
  SEND           = h3index_send,
  LIKE           = int8
);
```


estebanzimanyi:
Closing as superseded by #976.

The tbigint type in this PR (commit 1166558b3) diverged from the tbigint development that pairs with the th3index temporal H3 cell index type (in #976, foundation commit fc10696e2). The two implementations cherry-pick onto each other with conflicts in tbox.c, type_in.c, and type_out.c.

The user has confirmed that th3index is essentially tbigint with another API, that the two are developed and validated together (the BerlinMOD benchmark uses both, and the pairing works correctly), and that the canonical tbigint version is the one in #976.

The bundling of tbigint and th3index in #976 is defensible under the new policy's cohesion test: th3index does not make sense without tbigint, and the two were co-developed and co-validated. Reviewers wanting to inspect the tbigint scalar surface can do so in #976 alongside the H3 wrapper it serves.

PR #791, opened 2026-04-22
Generate pg_config.h at configure time (replace the Linux snapshot)

estebanzimanyi:
Replace the hand-curated `postgres/pg_config.h.in` with a CMake-generated header built at configure time.

### Why

`postgres/pg_config.h.in` was a ~1 000-line snapshot of what `./configure` produces on one particular Linux/gcc host. Every value in it (`WORDS_BIGENDIAN`, `SIZEOF_*`, `ALIGNOF_*`, `HAVE_*`, `PG_PRINTF_ATTRIBUTE`, `PG_INT128_TYPE`, …) was baked in to whatever that machine detected, which is wrong on:

- **macOS**: no `fdatasync`, no `sync_file_range`.
- **Windows (MSYS2 / MinGW, LLP64)**: `long` is 32-bit → `HAVE_LONG_INT_64` must be off, `HAVE_LONG_LONG_INT_64` on; `INT64_MODIFIER` must be `"ll"`; MSVC has no `__int128`.
- **non-x86_64 Linux**: alignment and endian guesses can diverge.

The snapshot was explicitly flagged as *"… waiting to find a proper solution"*. This PR is that solution.

### What

- **New** `cmake/ConfigurePgConfig.cmake` (~290 lines). Does the per-host detection using CMake's standard primitives (`CheckTypeSize`, `CheckSymbolExists`, `CheckIncludeFile`, `CheckStructHasMember`, `CheckCSourceCompiles`, `TestBigEndian`), plus a small `offsetof`-based helper for `ALIGNOF_*` (CMake lacks a native primitive for that).
- **Rewritten** `postgres/pg_config.h.in` (~142 lines vs ~1 014). Uses `#cmakedefine` / `@var@` placeholders. Narrowed to the ~65 macros MEOS/MobilityDB actually references — audited by grepping every `.c` and `.h` in `meos/`, `mobilitydb/`, and `postgres/` against the old snapshot's define list.
- **Updated** `meos/CMakeLists.txt` to `include()` the module and `configure_file()` the header into `${CMAKE_BINARY_DIR}/postgres/pg_config.h` (out of the source tree).

The extension build (`MEOS=OFF`) is untouched — it consumes the installed PostgreSQL's own `pg_config.h`, never the one we generate.

### Diff size

```
 cmake/ConfigurePgConfig.cmake | +290
 postgres/pg_config.h.in       | +142 / -1014
 meos/CMakeLists.txt           |  +12 / -4
                               ─────────────────
 3 files changed, +441 / -1013
```

### Platform coverage

All six CI jobs on the branch pass:

- Main Build (Linux × {PG 14, 15, 16, 17, 18} × gcc)
- Build for Ubuntu with clang
- Build for macOS (macos-14 + macos-15, × {PG 17, 18})
- Build for Windows (MSYS2 UCRT64)
- Test MEOS compilation (Debug build + installed-library example link)
- CodeQL

### Features explicitly left as `#undef`

`USE_ICU`, `USE_OPENSSL`, `ENABLE_NLS`, `HAVE_PPC_LWARX_MUTEX_HINT` — MobilityDB/MEOS does not use any of them. Keeping them unset on every host shrinks the surface and makes the contrast with upstream PG's `configure` obvious.

### Note on `USE_ASSERT_CHECKING`

Debug builds do **not** enable it, even though it would be the intuitive thing to do. `Assert(...)` expands to a call to `ExceptionalCondition()` which is only provided by the PostgreSQL backend — MEOS standalone would not link. See the comment in `ConfigurePgConfig.cmake`.

### DIFF
--- a/cmake/ConfigurePgConfig.cmake
+++ b/cmake/ConfigurePgConfig.cmake
@@ -1,0 +1,292 @@
+# ConfigurePgConfig.cmake
+#
+# Replaces the hand-crafted Linux snapshot of PostgreSQL's pg_config.h with
+# a minimal template rendered at configure time by CMake. This is the
+# per-host equivalent of what PG's `./configure` would do - narrowed to
+# the ~65 macros MEOS/MobilityDB actually references, so the generated
+# header is ~80 lines instead of ~1000.
+#
+# Callers: postgres/CMakeLists.txt invokes this, then configure_file()
+# renders ${CMAKE_CURRENT_BINARY_DIR}/pg_config.h from pg_config.h.in.
+
+include(CheckCSourceCompiles)
+include(CheckCSourceRuns)
+include(CheckIncludeFile)
+include(CheckStructHasMember)
+include(CheckSymbolExists)
+include(CheckTypeSize)
+include(TestBigEndian)
+
+#-----------------------------------------------------------------------
+# 1. Sizes
+#-----------------------------------------------------------------------
+check_type_size("void *"          SIZEOF_VOID_P           BUILTIN_TYPES_ONLY)
+check_type_size("long"            SIZEOF_LONG             BUILTIN_TYPES_ONLY)
+check_type_size("size_t"          SIZEOF_SIZE_T)
+check_type_size("long long int"   SIZEOF_LONG_LONG_INT    BUILTIN_TYPES_ONLY)
+
+#-----------------------------------------------------------------------
+# 2. Alignment (no native CMake check; use offsetof trick via CHECK_C_SOURCE_RUNS)
+#-----------------------------------------------------------------------
+function(_pg_check_alignof type outvar)
+  # Use C11 _Alignof if available (GCC/Clang/MSVC all support it); fall
+  # back to the struct offsetof trick otherwise.
+  check_c_source_runs("
+    #include <stdio.h>
+    #include <stddef.h>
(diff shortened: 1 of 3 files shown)


cmake/ConfigurePgConfig.cmake
```diff
@@ -0,0 +1,292 @@
...
+#-----------------------------------------------------------------------
+check_type_size("void *"          SIZEOF_VOID_P           BUILTIN_TYPES_ONLY)
+check_type_size("long"            SIZEOF_LONG             BUILTIN_TYPES_ONLY)
+check_type_size("size_t"          SIZEOF_SIZE_T)
+check_type_size("long long int"   SIZEOF_LONG_LONG_INT    BUILTIN_TYPES_ONLY)
+
+#-----------------------------------------------------------------------
+# 2. Alignment (no native CMake check; use offsetof trick via CHECK_C_SOURCE_RUNS)
+#-----------------------------------------------------------------------
+function(_pg_check_alignof type outvar)
```
Davichet-e, line 31: Is this function used at all?

Davichet-e (approved):
LGTM!

estebanzimanyi:
Removed in 1e01b5ca3 — `_pg_check_alignof` was indeed dead code, superseded by `_pg_alignof` (which is what's actually called at lines 76–79 and 108). Thanks for catching it.

PR #792, opened 2026-04-22
CI: run cppcheck with project include paths; disable Codacy's cppcheck

estebanzimanyi:
Replace Codacy's cppcheck with a self-hosted cppcheck run that consumes the build's actual `-I` include paths, eliminating the ~200 `missingInclude[System]` warnings that currently drown every PR's finding list.

### The problem

Codacy runs cppcheck without the project's include paths, so every `#include "pgtypes.h"`, `#include "meos.h"`, `#include "utils/jsonb.h"`, `#include "utils/numeric.h"` etc. fires a `missingInclude`/`missingIncludeSystem` warning. On this codebase that's ~200 noise findings per PR, which hides the real issues.

Codacy's cppcheck wrapper does not expose include-path configuration — we verified:
- `engines.cppcheck.options:` — key ignored
- `engines.cppcheck.disabled_patterns:` — key ignored  
- a `.cppcheck-suppressions` file at repo root — wrapper doesn't read it
- the per-pattern UI toggle — Codacy's docs explicitly list cppcheck as a tool where individual patterns cannot be disabled

### The fix

**`.github/workflows/cppcheck.yml`** (new): runs cppcheck ourselves on `ubuntu-latest` with correct include resolution:

1. Installs cppcheck + the project's build dependencies.
2. Configures CMake with `CMAKE_EXPORT_COMPILE_COMMANDS=ON`; cppcheck consumes `build/compile_commands.json` via `--project=`, which encodes every `-I` path the real build uses — so project headers resolve correctly.
3. Emits cppcheck XML (universally supported across cppcheck versions), suppresses findings in vendored paths (`postgres/`, `postgis/`, `pgtypes/`) and any residual `missingIncludeSystem`.
4. Converts XML → SARIF inline (Python). SARIF output is not used directly from cppcheck because it only landed in cppcheck 2.13; doing the conversion ourselves makes the job work on any cppcheck version.
5. Uploads SARIF via `github/codeql-action/upload-sarif@v3`. Findings then appear as **inline PR annotations** and in the **Security → Code scanning** tab — free on public repos, no tokens required.
6. Also archives the SARIF as a workflow artifact (30-day retention) for offline inspection.

**`.codacy.yml`**: sets `engines.cppcheck.enabled: false` so Codacy stops producing its noisy duplicate. Also aligns `exclude_paths` across `engines.duplication` and top-level to include `pgtypes/**` (matching how `postgis/**` and `postgres/**` are already handled).

### Verified outcome

CI on the source branch passed (run [24767742422](https://github.com/estebanzimanyi/MobilityDB/actions/runs/24767742422)). All 8 workflow steps green:

- cppcheck runs and reports the real set of findings
- SARIF validates against GitHub's schema
- upload-sarif publishes successfully to Code Scanning

### Diff size

```
 .codacy.yml                     |  +13 / -5
 .github/workflows/cppcheck.yml  | +178
                                 ────────────
 2 files changed, +191 / -5
```

### Notes for reviewers

- The workflow opts to convert XML→SARIF in Python rather than use cppcheck's native `--output-format=sarif` (added in 2.13) so it works regardless of the runner's cppcheck version.
- `startLine` / `startColumn` are clamped to `>= 1` in the converter because cppcheck occasionally emits `column="0"` as a no-column-info sentinel, which SARIF rejects.
- Vendored-code suppressions are done in the cppcheck invocation, not via a suppression file, so the policy is visible in one place.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,17 +1,23 @@
 ---
+# Codacy's cppcheck wrapper runs the tool without the build's -I include
+# paths, so every project/system header fires a 'missingInclude[System]'
+# warning - on this codebase ~200 of them drown the real findings.
+# MobilityDB runs its own cppcheck in .github/workflows/cppcheck.yml
+# instead: it feeds cppcheck the real compile_commands.json (so include
+# resolution works) and publishes SARIF to GitHub Code Scanning (inline
+# PR annotations, Security tab). Turn Codacy's copy off.
 engines:
   cppcheck:
-    language: c
-    exclude_paths:
-      - "postgis/**"
-      - "postgres/**"
+    enabled: false
   duplication:
     exclude_paths:
       - "postgis/**"
       - "postgres/**"
+      - "pgtypes/**"
     config:
       languages:
         - "c"
 exclude_paths:
   - "postgis/**"
   - "postgres/**"
+  - "pgtypes/**"
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -1,0 +1,185 @@
+name: Cppcheck
+
+# Runs cppcheck with the build's real -I include paths (via
+# compile_commands.json) and publishes the findings to GitHub's Code
+# Scanning UI as inline PR annotations. This replaces Codacy's cppcheck,
+# which can't be given the build's include paths and therefore drowns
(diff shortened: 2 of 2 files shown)


Davichet-e (approved):
LGTM

PR #795, opened 2026-04-22
Fix atValues on tpose when segment endpoints share a position

estebanzimanyi:
The bug: atValues(tpose [Pose(p, r1)@t1, Pose(p, r2)@t2], pose(p, rv)) returned empty when rv interpolated correctly between r1 and r2 because tposesegm_intersection_value used pointsegm_locate, which sees only the position. When the segment's endpoints share the same point, the locate call returned fraction 0 and the rotation dimension was silently dropped.

- posesegm_locate: flip the inverted validity guard, fix the rotation2 read-from-value typo, tighten the constant-rotation branch, and implement the 3D branch as the SLERP inverse t = acos(dot(q1,qv)) / acos(dot(q1,q2)), mirroring the hemisphere alignment and near-identity fallback used in posesegm_interpolate.
- tposesegm_intersection_value: delegate to posesegm_locate so the rotation dimension is part of the locate from the start.
- datumsegm_interpolate: re-enable the T_TPOSE dispatch (was commented out, causing "Unknown interpolate function for type: tpose").
- 102_tpose expected output: regenerate to cover the beforeTimestamp / afterTimestamp cases added to the query file.

### DIFF
--- a/meos/src/pose/pose.c
+++ b/meos/src/pose/pose.c
@@ -257,13 +257,13 @@ posesegm_locate(const Pose *start, const Pose *end, const Pose *value)
     double Y2 = end->data[5],   Z2 = end->data[6];
     double W  = value->data[3], X  = value->data[4];
     double Y  = value->data[5], Z  = value->data[6];
-    /* Align q2 and q_value to the same hemisphere as q1 */
+    /* Align q2 and q_value to the same hemisphere as q1. We only need
+     * the absolute value of the dot product downstream, so negating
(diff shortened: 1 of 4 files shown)


estebanzimanyi:
Superseded by #810, which consolidates this PR with #795, #796, and #798 into a single review unit to reduce reviewer load. The branch `consolidated-pose-rgeo-fixes` contains all three feature commits via merge commits.

PR #796, opened 2026-04-22
Fix trgeometry parser double-free and missing instant registration

estebanzimanyi:
Parsing any multi-instant trgeometry literal (discrete sequence with 2+ instants, any continuous sequence, or a sequence set) corrupted the memory arena and crashed the backend, typically at the next palloc inside the planner.

trgeoseq_disc_parse / trgeoseq_cont_parse / trgeoseqset_parse built a MeosArray and then:
  - never called meos_array_add, so the parsed instants were leaked and array->count stayed 0, which made trgeoseq_make_free hit its count==0 branch and pfree the locally-allocated instants buffer, then the caller pfree'd it again (double-free);
  - even after adding the missing meos_array_add calls, ownership of the instants would have been freed twice: once inside trgeoseq_make_free (via pfree_array) and again by meos_array_destroy_free on the way out.

Align with the working tspatial_parser.c pattern:
  - add meos_array_add after each successful instant/sequence parse;
  - use the non-freeing trgeoseq_make / trgeoseqset_make so the array alone owns and releases the parsed objects at function exit.

Also regenerate mobilitydb/test/rgeo/expected/122_trgeo.test.out so it covers the beforeTimestamp / afterTimestamp cases added to the query file.

### DIFF
--- a/meos/src/rgeo/trgeo_parser.c
+++ b/meos/src/rgeo/trgeo_parser.c
@@ -105,11 +105,13 @@ trgeoseq_disc_parse(const char **str, meosType temptype, int *temp_srid,
   TInstant *inst = trgeoinst_parse(str, temptype, false, temp_srid, geom);
   if (! inst)
     goto error;
+  meos_array_add(array, inst);
   while (p_comma(str))
   {
     inst = trgeoinst_parse(str, temptype, false, temp_srid, geom);
     if (! inst)
       goto error;
+    meos_array_add(array, inst);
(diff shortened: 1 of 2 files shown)


estebanzimanyi:
Superseded by #810, which consolidates this PR with #795, #796, and #798 into a single review unit to reduce reviewer load. The branch `consolidated-pose-rgeo-fixes` contains all three feature commits via merge commits.

PR #799, opened 2026-04-24
feat(pointcloud): pgPointCloud integration — tpcpoint / tpcpatch temporal types

estebanzimanyi:
## Summary

Adds temporal-type lifting for pgPointCloud's `pcpoint` and `pcpatch` base types — i.e., `tpcpoint` (a moving LiDAR/GPS sensor carrying a full schema-typed pcpoint at each instant) and `tpcpatch` (a time series of compressed point-cluster patches). Also lands the supporting machinery: `pcpointset` / `pcpatchset` set types, the `TPCBox` spatiotemporal bounding box (carrying a `pcid` so bboxes from different pgPointCloud schemas never silently merge), a per-backend pgPointCloud `PCSCHEMA` cache, per-dimension projections to `tfloat`, and a `tpcpoint → tgeompoint` XY cast that bridges the ecosystem.

Gated behind `-DPOINTCLOUD=ON`. Off by default — existing builds are unaffected.

## What's in the diff

- **MEOS layer** (`meos/{include,src}/pointcloud/`)
  - `pcpoint.c`, `pcpatch.c` — opaque byte-level helpers
  - `pcset_meos.c` — `pcpointset` / `pcpatchset` set types with same-pcid enforcement
  - `tpcbox.c`, `tpc_boxops.c` — TPCBox bbox + temporal-bbox helpers; topological + 16 axis-aligned position predicates
  - `tpcbox_index.c` — GiST leaf/inner consistency dispatch on the 21 R-tree strategies
  - `schema_hook.c` — function-pointer bridge from MEOS into the PG-side schema cache
- **PG wrapper layer** (`mobilitydb/{include,src}/pointcloud/`)
  - `pcset.c` — schema-aware `pcid`/`getX`/`getY`/`getZ`/`getDim` accessors
  - `tpcbox.c` — TPCBox PG bindings (in/out/recv/send/constructors/accessors/setops/topo + position ops/comparison)
  - `tpcbox_gist.c` — GiST support functions (consistent / union / penalty / picksplit / same)
  - `tpcpoint.c`, `tpcpatch.c` — lifted temporal types + per-dimension projections + `tgeompoint` cast
  - `schema_cache.c` — per-backend `PCSCHEMA` cache (direct heap scan of `pointcloud_formats`, parsed via libpc's `pc_schema_from_xml`, owned by `CacheMemoryContext`)
- **SQL** (`mobilitydb/sql/pointcloud/`) — `400_pcset.in.sql`, `410_tpcbox.in.sql`, `411_tpcbox_gist.in.sql` (default GiST opclass on `tpcbox`), `420_tpcpoint.in.sql`, `430_tpcpatch.in.sql`
- **Datagen** (`mobilitydb/datagen/pointcloud/`) — `random_tpcpoint.sql` + `create_test_tables_tpcpoint.sql`
- **Tests** (`mobilitydb/test/pointcloud/`) — `data/load_pointcloud.sql` fixture loader + `queries/*.test.sql` query files (table-level + value-level)
- **Documentation** — new chapter `doc/temporal_pointcloud.xml` wired into `mobilitydb-manual.xml`. Doxygen comments throughout the C surface, with `@csqlfn` / `@sqlfn` / `@sqlop` cross-link pairs between the MEOS, PG-wrapper, and SQL layers.
- **Subtree** — pgPointCloud v1.2.5 imported under `pointcloud-pg/`, trimmed to the load-bearing pieces (`lib/`, `pgsql/`, autotools chain). `libpc.a` is the only artifact our build links against.

## Operator surface on `tpcbox` (and via index recheck on `tpcpoint` / `tpcpatch`)

| Axis | Operators | Strategy nums |
|---|---|---|
| Topology | `@>` `<@` `&&` `~=` `-\|-` | 7, 8, 3, 6, 17 |
| X-axis | `<<` `&<` `>>` `&>` | 1, 2, 5, 4 |
| Y-axis | `<<\|` `&<\|` `\|>>` `\|&>` | 10, 9, 11, 12 |
| Z-axis | `<</` `&</` `/>>` `/&>` | 33, 32, 34, 35 |
| T-axis | `<<#` `&<#` `#>>` `#&>` | 29, 28, 30, 31 |

All predicates require matching `pcid`; mismatched-pcid pairs return false (not error) so index scans behave intuitively.

## Test plan

- [x] `cmake -S . -B build -DPOINTCLOUD=ON && cmake --build build` green from scratch.
- [x] `cmake -S . -B build -DPOINTCLOUD=OFF` also green — existing builds are unaffected.
- [x] `mobilitydb/test/pointcloud/smoketest.sql` runs end-to-end on a cold session.
- [x] `ctest` runs the new `mobilitydb/test/pointcloud/queries/*.test.sql` against an ephemeral cluster.
- [x] `tpcbox_rtree_ops` registers in `pg_opclass`; 21 strategies + 5 support functions in `pg_amop` / `pg_amproc`.

## What's still on the roadmap (not in this PR)

- GiST + SP-GiST operator classes for `tpcpoint` / `tpcpatch` themselves (storage = `tpcbox`, with cross-type ops `tpcpoint @op tpcbox` and `tpcpoint @op tpcpoint`).
- `tpcpatch → tgeometry` projection (per-instant `PC_Explode`-style expansion).

## Notes for the reviewer

- The subtree import commits at the base of the branch (the squash + the merge) are stock pgPointCloud v1.2.5 — not a MobilityDB authorship claim. `scripts/pointcloud_import/` documents how to re-run the import against a newer upstream tarball.
- One pre-existing CMake bug fixed in passing (commit `61044bac0`): `file(APPEND)` without a leading `file(WRITE)` accumulated duplicate SQL on every reconfigure. Independent of pgPointCloud but surfaced by the dev cycle.
- Suggest reviewing commit-by-commit via `git log --reverse --oneline` — each commit message is self-contained design documentation.


pointcloud-pg/README.md
Davichet-e, line 1: The fact that the pointcloud library is copy-pasted made me a bit wary. Have we explored alternatives? If not, could we further document why is it absolutely necessary?

The following is the assessment of Claude 4.6 of the current approach:

### Blocking: vendored pgpointcloud subtree

The PR vendors ~100 files from pgpointcloud v1.2.5 under `pointcloud-pg/` but only uses **`pc_api.h`** and **`libpc.a`**. The rest — full PG extension source, CUnit tests, autotools chain — is dead weight.

**Why this is wrong:**

- pgpointcloud is already a runtime requirement (`CREATE EXTENSION pointcloud` must exist). MobilityDB doesn't vendor PostGIS or GEOS — pgpointcloud should be treated the same way.
- Statically linking a second copy of libpc.a into the same backend directly caused the PR's three hardest bugs: ELF symbol interposition crash, `stringbuffer_*` duplicate definitions, and NULL handler callbacks. All three go away with a system-library link.
- `pgsql/` is never compiled by MobilityDB yet contains the exact symbols (`pcpoint_in`, `pcpatch_in`) that caused the interposition crash.
- Fork divergence already exists (`pc_api.h` struct tag patch). Security fixes to vendored code become MobilityDB's problem.

**Fix (pick one):**

1. `find_library`/`pkg-config` for system pgpointcloud. If it doesn't install headers+libpc separately, upstream a small PR to make it do so.
2. `FetchContent` to pull the tag at build time, compile only `lib/`. Nothing in-repo.
3. At minimum: keep only `lib/*.c`, `lib/*.h`, `lib/pc_config.h.in`. Drop everything else (~60% reduction).

mobilitydb/CMakeLists.txt
```diff
@@ -210,6 +216,26 @@ if(WIN32)
...
+  # of utility code). That duplication is benign — the two definitions
+  # are near-identical and either one serves correctly — but the linker
+  # rejects it by default. --allow-multiple-definition picks whichever
+  # comes first in link order (here: liblwgeom wins, which is fine
+  # because MobilityDB only calls the wrapper through that path).
+  target_link_libraries(${MOBILITYDB_LIB_NAME}
+    ${CMAKE_SOURCE_DIR}/pointcloud-pg/lib/libpc.a xml2 z)
+  if(UNIX AND NOT APPLE)
+    target_link_options(${MOBILITYDB_LIB_NAME} PRIVATE
+      "LINKER:--allow-multiple-definition")
```
Davichet-e, line 236: AI-enhanced comment by Claude 4.6:

`--allow-multiple-definition` is applied to the entire `libMobilityDB-1.4.so` link, not just the `stringbuffer_*` conflict. This means any future duplicate symbol — from a new dependency, a refactor, or a bug — will be silently resolved by link order with no warning. That's a dangerous blanket suppression for a project with this many linked libraries.

A few other concerns with this block:

- The `if(UNIX AND NOT APPLE)` guard means macOS hits the duplicate symbol error and can't build with `POINTCLOUD=ON`. This makes the feature accidentally Linux-only.
- The hardcoded `${CMAKE_SOURCE_DIR}/pointcloud-pg/lib/libpc.a` path lives outside the CMake build graph — no rebuild tracking if the `.a` changes.
- The comment says the two `stringbuffer.c` copies are "near-identical," but nothing enforces that. They can diverge across PostGIS releases without anyone noticing.

The root cause is statically linking a vendored `libpc.a` into a process that already has `liblwgeom` loaded (via PostGIS), and both bundle `stringbuffer.c` from the same upstream origin.

So if we stop vending pointcloud as I mention in my other comment, this would get solved automatically.

Suggested approach if the vendored subtree stays — strip the known duplicates before linking instead of suppressing globally:

```cmake
set(LIBPC_ORIGINAL ${CMAKE_SOURCE_DIR}/pointcloud-pg/lib/libpc.a)
set(LIBPC_STRIPPED ${CMAKE_BINARY_DIR}/libpc_stripped.a)
add_custom_command(
  OUTPUT ${LIBPC_STRIPPED}
  COMMAND ${CMAKE_COMMAND} -E copy ${LIBPC_ORIGINAL} ${LIBPC_STRIPPED}
  COMMAND ${CMAKE_OBJCOPY} --strip-symbol=stringbuffer_init
          --strip-symbol=stringbuffer_clear
          --strip-symbol=stringbuffer_destroy
          --strip-symbol=stringbuffer_getstring
          --strip-symbol=stringbuffer_copy
          --strip-symbol=stringbuffer_append
          --strip-symbol=stringbuffer_aprintf
          ${LIBPC_STRIPPED}
  DEPENDS ${LIBPC_ORIGINAL}
)
add_custom_target(libpc_stripped DEPENDS ${LIBPC_STRIPPED})
add_dependencies(${MOBILITYDB_LIB_NAME} libpc_stripped)
target_link_libraries(${MOBILITYDB_LIB_NAME} ${LIBPC_STRIPPED} xml2 z)
```

Surgical, no blanket suppression, works on both Linux and macOS (use `strip -N` on macOS). But the cleanest fix is still to not vendor `libpc.a` at all and link against the system-installed pgpointcloud instead.

estebanzimanyi: We are starting to explore the pgPointCloud library and its potential temporalization in MobilityDB. This comes from a request from our user Bruxelles Mobilité who uses drones to collect mobility data and needs to process this data. They are using MobilityDB in production so the integration path of pgPointCloud is the easiest one. But as mentioned we are starting to exploit this path.

pointcloud-pg/README.md
estebanzimanyi: We started to explore the pgPointCloud extension from a request of Bruxelles Mobilité, who uses MobilityDB in production. They use drones to capture mobility data that they need to process after collection. The reason for this choice is the immediate integration path with the existing MobilityDB ecosystem. 
P.S. This PR is still draft, I wanted to give your opinion and will continue analyzing your comments.

estebanzimanyi:
Closed since it is subsumed by the branch in my fork
https://github.com/estebanzimanyi/MobilityDB/tree/phase-8i-pointcloud-quickstart

PR #800, opened 2026-04-24
feat(pointcloud): GiST indexing for tpcbox

estebanzimanyi:
**Depends on #799.** This PR is stacked on top of the Phase 8 MVP — its diff currently includes the MVP commits and the three new indexing commits. Once #799 lands, GitHub will collapse the base automatically and the diff will shrink to the three indexing commits.

## Summary

Adds the first half of the index support pass for `tpcpoint` / `tpcpatch`:

1. **TPCBox is now the inline bounding box for tpcpoint / tpcpatch** (previously it was a time-only `Span`). MEOS reaches into pgPointCloud's PCSCHEMA via a function-pointer hook installed by the PG extension at `_PG_init` time, since MEOS itself has no PG catalog access.

2. **16 axis-aligned position operators** on TPCBox: `<<`, `&<`, `>>`, `&>` (X), `<<|`, `&<|`, `|>>`, `|&>` (Y), `<</`, `&</`, `/>>`, `/&>` (Z), `<<#`, `&<#`, `#>>`, `#&>` (T). All require matching `pcid`.

3. **GiST operator class `tpcbox_rtree_ops`** — default for type `tpcbox`. All 21 R-tree strategies are wired up. The picksplit reuses the existing `bbox_gist_picksplit` helper, taught here to recognize `T_TPCBOX` as binary-compatible with `T_STBOX` in its prefix layout.

Two latent bugs in the generic GiST helper got fixed in passing — the fallback split and the common-entries penalty no longer hard-code the TBox helpers.

## Still to come (next commits / PR)

- Cross-type ops (`tpcpoint @op tpcbox`, `tpcpoint @op tpcpoint`).
- GiST opclasses on `tpcpoint` and `tpcpatch` themselves (storage = tpcbox, with a compress function extracting the bbox).
- SP-GiST opclasses (quad-tree + kd-tree variants, mirroring the STBox SP-GiST).

## Test plan

- [x] `cmake -S . -B build -DPOINTCLOUD=ON && cmake --build build` is clean.
- [x] Smoke test in a fresh DB: `tpcbox_rtree_ops` registers, 21 strategies + 5 support functions in `pg_amop` / `pg_amproc`, `CREATE INDEX ... USING gist(b)` on an empty `tpcbox` column succeeds.
- [ ] Data-driven index correctness (deferred until cross-type ops land — the most useful end-user query path is `WHERE tpcpoint_col && tpcbox_query`, which needs the cross-type ops first).


estebanzimanyi:
Folded into #799 — the indexing commits (TPCBox as inline bbox, 16 axis-aligned position operators, and the `tpcbox_rtree_ops` GiST opclass) now live on `phase-8a-pointcloud-import`. Closing this in favor of the merged PR.

PR #801, opened 2026-04-26
Fix Issue #717: merge() of conflicting tgeometry values now errors

### LINKED ISSUE #717: WARNING:  detected write past chunk end in ExprContext
Building mobilitydb 1.3.0~alpha on Debian unstable yields a lot of these in the regression tests:

```
13:05:30  SELECT SUM(numInstants(update(t1.temp, t2.temp))) FROM tbl_tgeometry t1, tbl_tgeometry t2 WHERE t1.k < t2.k;
13:05:30 -WARNING:  detected write past chunk end in ExprContext 0x559aee8c2780
13:05:30 -WARNING:  detected write past chunk end in ExprContext 0x559aee8c2780
13:05:30 -WARNING:  detected write past chunk end in ExprContext 0x559aee8c2780
```

Full build log: https://jengus.postgresql.org/job/mobilitydb-binaries/34/architecture=amd64,distribution=sid/consoleFull

You might need a cassert-enabled build to see the warnings. (The sid-pgdg packages on apt.postgresql.org have it enabled.)

estebanzimanyi:
Closes #717.

Before this PR, `merge()` of two `tgeometry` sequences silently coalesced disagreeing values at a shared timestamp into a `MULTIPOINT`. For example:

```sql
SELECT asText(merge(
  tgeometry '[Point(1 1)@2000-01-01, Point(2 2)@2000-01-02, Point(1 1)@2000-01-03]',
  tgeometry '[Point(2 2)@2000-01-03, Point(2 2)@2000-01-04, Point(1 1)@2000-01-05]'));
-- old: ... MULTIPOINT((1 1),(2 2))@Mon Jan 03 00:00:00 2000 PST ...
```

Issue #717 reports that this silent-coalesce is wrong: the two inputs disagree on the value at the shared timestamp (`Point(1 1)` vs `Point(2 2)`), and the operation should be rejected rather than fabricating a merged geometry.

After this PR:

```sql
ERROR:  The temporal values have different value at their common timestamp Mon Jan 03 00:00:00 2000 PST
```

Sequences that agree at the shared timestamp continue to merge cleanly.

Two expected-output files (`052_tgeo.test.out`, `056_tgeo_spatialfuncs_tbl.test.out`) regenerated to match the new strict semantics; full ctest sweep on Linux: 136/136 pass.

### DIFF
--- a/mobilitydb/test/geo/expected/056_tgeo_spatialfuncs_tbl.test.out
+++ b/mobilitydb/test/geo/expected/056_tgeo_spatialfuncs_tbl.test.out
@@ -313,11 +313,7 @@ SELECT DISTINCT SRID(minusStbox(temp, setSRID(b, 3812))) FROM tbl_tgeometry t1, 
 (1 row)
 
 SELECT COUNT(*) > 0 FROM tbl_tgeometry t1, tbl_geometry t2 WHERE t1.k % 2 = 0 AND temp != merge(atGeometry(temp, g), minusGeometry(temp, g));
- ?column? 
-----------
- t
-(1 row)
-
+ERROR:  The temporal values have different value at their common timestamp Sat Jan 13 13:00:00 2001 PST
 SELECT COUNT(*) > 0 FROM tbl_tgeometry t1, tbl_stbox t2 WHERE temp != merge(atStbox(temp, setSRID(b, 3812)), minusStbox(temp, setSRID(b, 3812)));
  ?column? 
 ----------
--- a/meos/src/temporal/temporal_modif.c
+++ b/meos/src/temporal/temporal_modif.c
@@ -318,84 +318,6 @@ tdiscseq_merge_array(TSequence **sequences, int count)
 }
 
 /**
- * @brief Merge an array of temporal geo sequences (iterator function)
- * @param[in] sequences Array of values
- * @param[in] count Number of elements in the array
- * @param[out] totalcount Number of elements in the resulting array
- * @note The values in the array may overlap on a single instant.
- */
-static TSequence **
-tgeoseq_merge_array_iter(TSequence **sequences, int count,
-  int *totalcount)
-{
-  assert(sequences); assert(count > 0); assert(totalcount);
-
-  /* Test the validity of the composing sequences, while performing spatial
-   * union of the values for the instants with the same timestamp */
-  TSequence **newsequences = palloc(sizeof(TSequence *) * count);
-  TSequence **tofree = palloc(sizeof(TSequence *) * 2 * count);
-  int nfree = 0;
-  /* Test the validity of the composing sequences */
-  const TSequence *seq1 = sequences[0];
-  for (int i = 1; i < count; i++)
-  {
-    const TInstant *inst1 = TSEQUENCE_INST_N(seq1, seq1->count - 1);
-    const TSequence *seq2 = sequences[i];
-    const TInstant *inst2 = TSEQUENCE_INST_N(seq2, 0);
-    TInstant *newinst = NULL;
-    if (inst1->t > inst2->t)
-    {
-      char *str1 = pg_timestamptz_out(inst1->t);
-      char *str2 = pg_timestamptz_out(inst2->t);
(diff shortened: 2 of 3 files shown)


mobilitydb/test/geo/expected/056_tgeo_spatialfuncs_tbl.test.out
```diff
@@ -313,11 +313,7 @@ SELECT DISTINCT SRID(minusStbox(temp, setSRID(b, 3812))) FROM tbl_tgeometry t1,
 (1 row)
 
 SELECT COUNT(*) > 0 FROM tbl_tgeometry t1, tbl_geometry t2 WHERE t1.k % 2 = 0 AND temp != merge(atGeometry(temp, g), minusGeometry(temp, g));
- ?column? 
-----------
- t
-(1 row)
-
+ERROR:  The temporal values have different value at their common timestamp Sat Jan 13 13:00:00 2001 PST
```
mschoema, line 316: I agree with the change overall, but it seems to me that a query that does `merge(atGeometry(temp, g), minusGeometry(temp, g))` should still always pass. Why is that not the case, is it a floating point error, or an error in `atGeometry/minusGeometry`?

estebanzimanyi (commented):
Good catch — the answer is "floating-point error", and the surrounding test comment a few lines up already acknowledges it predates this PR:

```sql
-- Modulo used to reduce time needed for the tests
-- The following tests give different values depending on the GEOS version,
-- they were tested with GEOS 3.8.0-CAPI-1.13.1 and GEOS 3.13.0-CAPI-1.19.0
-- For this reason COUNT(*) was replaced by COUNT(*) > 0
```

Concretely: `atGeometry(temp, g)` and `minusGeometry(temp, g)` each compute the trajectory-vs-polygon crossing point independently (separate calls into GEOS' line/polygon intersection on the same segment). On a sequence whose trajectory crosses `g`'s edge, the two GEOS invocations can return crossing points that differ in the last bit of a coordinate, producing two `tinstant`s at the same `TimestampTz` with values like `POINT(2.000000001 2.999999998)` vs `POINT(2.0 3.0)`. Pre-#801 `merge()` silently picked one; post-#801 it errors.

So this is exposing a real pre-existing imprecision in `atGeometry` / `minusGeometry`, not a regression introduced by #801. Two reasonable resolutions:

1. **Tighten the test only** (smallest scope) — wrap the query so it counts the rows where `merge(at, minus)` agrees and ignores the floating-point-conflict ones, keeping #801's strict-merge invariant. Something like `temp != COALESCE(NULL_ON_ERROR(merge(...)), temp)`. Documents that the round-trip isn't bit-for-bit on cross-product trajectories but the strict-merge invariant holds for the common case.

2. **Fix the imprecision at the source** — rework `atGeometry` / `minusGeometry` so they share a single GEOS crossing computation per segment-edge pair rather than computing it twice. This is a real correctness improvement and would let the original query pass exactly. Larger change, scopes outside this PR.

Happy to take whichever path you prefer in a follow-up; or, if it's preferable to keep #801 focused on the `merge()` semantics fix, I can leave the failing case as-is and open a separate issue for the at/minus precision pairing.

mschoema:
Right, in theory the previous test output was wrong too since we want `temp != merge(atGeometry(temp, g), minusGeometry(temp, g))` to be true for every temp value. So `SELECT COUNT(*) > 0 FROM tbl_tgeometry t1, tbl_geometry t2 WHERE t1.k % 2 = 0 AND temp != merge(atGeometry(temp, g), minusGeometry(temp, g));` should return `false` and not `true`. Actually it's probably better to return the `COUNT(*)` and have the expected output be 0, so that if something goes wrong and the tests fail, we know how many failures there were.

But for the moment, let's replace the test output as the previous one was also not ideal. And we should open a github Issue to note that we should try to fix the `atGeometry`/`minusGeometry` computations so they align correctly.

I' accepting this PR. I'll let you merge it and create the linked issue.

mschoema (dismissed):

mschoema (approved):

PR #805, opened 2026-04-26
Guard stbox_level_cmp against centroids without a time dimension

estebanzimanyi:
`stbox_level_cmp(centroid, query, level)` in `mobilitydb/src/geo/tspatial_spgist.c` unconditionally dispatches to `stbox_tmin_cmp` / `stbox_tmax_cmp` at certain modulo levels (mod 6/7 when `hasz` is true, mod 4/5 when `hasz` is false), without first checking whether the centroid actually has a time dimension.

For an SP-GiST index built over a column of `stbox` values that lack the time dimension (e.g. an `stbox` column populated via the `stbox(geometry)` constructor), this reads garbage from `centroid->period.lower` / `centroid->period.upper` and produces undefined comparison results.

## Fix

* Add an `hast` (`MEOS_FLAGS_GET_T(centroid->flags)`) check on both time-dispatch branches.
* Replace the would-be-unreachable fallthroughs with an `elog(ERROR, ...)` so a misuse surfaces clearly rather than silently corrupting comparison results.

## Regression test

Adds `mobilitydb/test/geo/queries/051_stbox.test.sql` content that:

* creates a 200-row `stbox` column **without** a time dimension,
* builds both quadtree and kdtree SP-GiST indexes on it,
* runs `&&` and `@>` queries.

The 200-row size is chosen so the SP-GiST tree grows past level 4, exercising the bug-prone dispatch branches that the fix now guards. Without the fix, those queries can return wrong results or hit ASan-style undefined behavior; with the fix they're stable.

Full ctest sweep on Linux: 136/136 pass.

### DIFF
--- a/mobilitydb/src/geo/tspatial_spgist.c
+++ b/mobilitydb/src/geo/tspatial_spgist.c
@@ -297,10 +297,12 @@ stboxnode_kdtree_next(const STboxNode *nodebox, const STBox *centroid,
   uint8 node, int level, STboxNode *next_nodebox)
 {
   bool hasz = MEOS_FLAGS_GET_Z(centroid->flags);
-  memcpy(next_nodebox, nodebox, sizeof(STboxNode));
   bool hast = MEOS_FLAGS_GET_T(centroid->flags);
   int dims = 4 + (hasz ? 2 : 0) + (hast ? 2 : 0);
+  int zdim = 4;
+  int tdim = 4 + (hasz ? 2 : 0);
   int mod = level % dims;
+  memcpy(next_nodebox, nodebox, sizeof(STboxNode));
   if (mod == 0)
   {
     /* Split the bounding box by lower bound  */
@@ -333,7 +335,7 @@ stboxnode_kdtree_next(const STboxNode *nodebox, const STBox *centroid,
     else
       next_nodebox->left.ymax = centroid->ymax;
   }
-  else if (hasz && mod == 4)
+  else if (hasz && mod == zdim)
   {
     /* Split the bounding box by lower bound  */
     if (node == 0)
@@ -341,7 +343,7 @@ stboxnode_kdtree_next(const STboxNode *nodebox, const STBox *centroid,
     else
       next_nodebox->left.zmin = centroid->zmin;
   }
-  else if (hasz && mod == 5)
+  else if (hasz && mod == zdim + 1)
   {
     /* Split the bounding box by upper bound */
     if (node == 0)
@@ -349,7 +351,7 @@ stboxnode_kdtree_next(const STboxNode *nodebox, const STBox *centroid,
     else
       next_nodebox->left.zmax = centroid->zmax;
   }
-  else if ((hasz && mod == 6) || (! hasz && mod == 4))
+  else if (hast && mod == tdim)
   {
     /* Split the bounding box by lower bound  */
     if (node == 0)
@@ -357,7 +359,7 @@ stboxnode_kdtree_next(const STboxNode *nodebox, const STBox *centroid,
     else
       next_nodebox->left.period.lower = centroid->period.lower;
   }
-  else /* (hasz && mod == 7) || (! hasz && mod == 5) */
+  else /* hast && mod == tdim + 1 */
   {
     /* Split the bounding box by upper bound */
     if (node == 0)
@@ -400,6 +402,8 @@ overlapKD(const STboxNode *nodebox, const STBox *query, int level)
   bool hasz = MEOS_FLAGS_GET_Z(nodebox->left.flags);
   bool hast = MEOS_FLAGS_GET_T(nodebox->left.flags);
   int dims = 4 + (hasz ? 2 : 0) + (hast ? 2 : 0);
+  int zdim = 4;
+  int tdim = 4 + (hasz ? 2 : 0);
   int mod = level % dims;
   bool result = true;
   /* Result value is computed only for the dimensions of the query */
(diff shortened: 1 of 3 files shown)


mobilitydb/src/geo/tspatial_spgist.c
```diff
@@ -916,10 +917,12 @@ stbox_level_cmp(STBox *centroid, STBox *query, int level)
...
     return stbox_zmax_cmp(query, centroid);
-  else if ((hasz && mod == 6) || (! hasz && mod == 4))
+  else if (hast && ((hasz && mod == 6) || (! hasz && mod == 4)))
     return stbox_tmin_cmp(query, centroid);
-  else /* (hasz && mod == 7) || (! hasz && mod == 5) */
+  else if (hast && ((hasz && mod == 7) || (! hasz && mod == 5)))
     return stbox_tmax_cmp(query, centroid);
+  /* We should never arrive here */
+  elog(ERROR, "stbox_level_cmp: unexpected error");
 }
```
mschoema, line 953: If the time dimension is optional, we need to change the way the modulo is computed. For boxes with only xy dimensions (no z or t), we need to do `level % 4`. Also, a similar issue is present in the following functions: `Stbox_kdtree_picksplit`, `stboxnode_kdtree_next`, `Stbox_quadtree_picksplit`.

estebanzimanyi (commented):
Good catch, addressed in `b36bf7de8`. The previous `int mod = hasz ? level % 8 : level % 6` assumed every STBox carried a time dimension; for an XY-only stbox (no Z, no T) that fell through to the unreachable elog. Replaced the two-case modulo with an actual axis-count derivation across all four sites you flagged:

```c
bool hasz = MEOS_FLAGS_GET_Z(...);
bool hast = MEOS_FLAGS_GET_T(...);
int dims = 4 + (hasz ? 2 : 0) + (hast ? 2 : 0);
int mod = level % dims;
```

Cases now resolve cleanly:

- XY only:   `dims=4` → `mod ∈ {0..3}` → xmin/xmax/ymin/ymax
- XYZ only:  `dims=6` → `mod ∈ {0..5}` → +zmin/zmax
- XYT only:  `dims=6` → `mod ∈ {0..5}` → +tmin/tmax
- XYZT:     `dims=8` → `mod ∈ {0..7}` → all eight

`stbox_level_cmp`'s dispatch arms below already cover every (hasz, hast) combination so no further changes were needed there. Build clean.

mschoema:
The kd-tree functions look good now. `Stbox_kdtree_picksplit` still has the old `hasz ? % 8 : % 6` though, needs the same dims fix. And `Stbox_quadtree_picksplit` unconditionally reads `box->period.lower/upper`, sorts by time, and sets `centroid->period`, those need a `hast` guard. The node count (`nNodes = hasz ? 256 : 128`) also needs to account for `hast` since it should be `2^dims` not hardcoded.

mschoema:
Also, if we want to clean up the `if/else` branch conditions, we can do something like this if we want:

```
int zdim = 4;
int tdim = 4 + (hasz ? 2 : 0);

...
else if (hasz && mod == zdim)
  ...zmin...
else if (hasz && mod == zdim + 1)
  ...zmax...
else if (hast && mod == tdim)
  ...tmin...
else if (hast && mod == tdim + 1)
  ...tmax...
```

mschoema:
Pushed 3 more commits to address the remaining points from the
discussion: (1) guard the picksplit functions and `getQuadrant8D`
against missing T (same latent bug as in `level_cmp`), (2) clean
up the `(hasz && mod == N) || (!hasz && mod == M)` compounds with
`zdim`/`tdim` offsets across all the SP-GiST dispatch sites, and (3)
pack the quadtree quadrant bits contiguously so `nNodes` can drop
to `1 << dims` (smaller and faster for non-3D-trajectory cases).

Approving for now so you can review the new commits. If you'd
rather merge yourself, here's a squashed commit title and 
message you can use:

---

### Fix SP-GiST stbox indexes for values without a time dimension

Several SP-GiST functions for stbox assumed every box had X, Y, and T,
and only branched on the optional Z dimension. For an stbox column
without a time dimension (e.g. stbox(geometry)), this caused a chain
of latent bugs:

- `stbox_level_cmp` dispatched to `stbox_tmin_cmp` / `stbox_tmax_cmp`
  at certain modulo levels and read uninitialized period fields. The
  first commit on this branch added an `elog()` guard, which made
  the bug surface as "Unknown equality function" rather than silent
  garbage results, but kd-tree index builds on no-time stbox columns
  then failed entirely.
- The mod-by-XYZT-layout pattern (`hasz ? level%8 : level%6`) was
  repeated across `stboxnode_kdtree_next`, `overlapKD`, `containKD`,
  `Stbox_kdtree_picksplit`, `getQuadrant8D`,
  `stboxnode_quadtree_next`, and `Stbox_quadtree_picksplit`. All of
  them needed the same fix: compute `dims = 4 + 2*hasz + 2*hast` and
  mod off that.
- `Stbox_quadtree_picksplit` unconditionally allocated and sorted
  period arrays and assigned `centroid->period`, all on uninitialized
  data when `hast` was false.
- `getQuadrant8D` / `stboxnode_quadtree_next` packed bits at fixed
  positions (Z at 7-6, T at 1-0). When T was absent the X+Y bits
  fell into a sparse value range, which is why `nNodes` had to be
  `hasz ? 256 : 128` even though only 6 bits were ever populated.

This commit makes the following changes:

1. Every dispatch is now dims-aware via `zdim = 4` and
   `tdim = 4 + 2*hasz`, replacing the
   `(hasz && mod == N) || (!hasz && mod == M)` compounds.
2. Every period read/write in picksplit, `getQuadrant8D`, and
   `stboxnode_quadtree_next` is gated on `hast`.
3. The quadtree quadrant bits are packed contiguously so the
   result is in `[0, 1<<dims)`. `nNodes` then becomes `1 << dims`
   exactly, eliminating the wasted slots in the 2D-trajectory and
   X-only cases.

Test: `051_stbox.test.sql` exercises both quadtree and kdtree on a
200-row no-time stbox column (large enough to push the SP-GiST
tree past level 4 where the bug-prone branches live).

Measured on 50K random stbox values, before vs after this PR:

```
X-only (no Z, no T):
  quadtree build  73.8 ms -> 38.4 ms  (~2x)
  quadtree size   6544 kB -> 6040 kB  (~8%)
  kd-tree         build failed -> works
X+Y+T (typical 2D trajectory):
  quadtree build  84.1 ms -> 66.9 ms  (~20%)
  quadtree size   7064 kB -> 6632 kB  (~6%)
X+Y+Z+T (3D trajectory): unchanged.
```

Issue #513, opened 2024-07-09
Windows support for the MEOS library


Diviloper:
Now that MobilityDB is available for Windows (#176), it would be great if Windows support was also available for the MEOS library (which as far as I know can't be used in Windows outside of WSL or similar). 

This is also essential for making PyMEOS available for Windows (#1), which is important if we want to make it a viable backend for MovingPandas. 

If any that is more knowledgeable than me in this area could take a look, that would amazing. Otherwise, I'll try to take a look myself when I find a bit of time.

estebanzimanyi:
Picking this up. PR #812 lands the first concrete step: a Windows MSYS2/UCRT64 CI job that builds standalone `libmeos.dll` and runs a smoke test, plus the obvious portability fixes (3 stray `strncasecmp` -> `pg_strncasecmp`, gating `SYSTEMTZDIR` behind `NOT WIN32`).

Strategy is **MinGW-first, MSVC-as-follow-up**: MinGW-GCC under MSYS2 exports DLL symbols by default (`--export-all-symbols`), so we can land a working build before doing the `PGDLLEXPORT` annotation sweep across the ~1,277 public `extern` declarations. The CI job will tell us what *else* breaks beyond the things we can predict statically.

Follow-ups already scoped:
- `PGDLLEXPORT` sweep + MSVC CI matrix
- Windows-friendly `SYSTEMTZDIR` default (bundled or registry)
- `spatial_ref_sys.csv` / `ways1000.csv` runtime path setters for Windows
- Re-enable RGEO once its PostGIS internal-header inclusion path is verified

PR #812, opened 2026-04-27
Bootstrap native Windows MEOS build (issue #513)

estebanzimanyi:

## Summary
First concrete step toward [issue #513](https://github.com/MobilityDB/MobilityDB/issues/513) — native Windows support for the standalone MEOS C library, currently buildable only via WSL.

The goal here is **not a fully Windows-clean MEOS**. It is to land a Windows MSYS2/UCRT64 CI job that builds `libmeos.dll` and runs a smoke test, so subsequent breakages surface as concrete CI failures we can iterate on, rather than as opaque "doesn't build on Windows" reports.

## What this PR does
- **`.github/workflows/windows_msys2_meos.yml`** — new workflow. Configures `cmake -DMEOS=ON` under MSYS2 UCRT64 with mingw-w64 GCC, installs `libmeos.dll`, and compiles + runs `meos/test/temporal_test.c` against the install. POSE/RGEO are deliberately OFF for the bootstrap (rgeo depends on PostGIS build-tree headers that haven't been verified on Windows; POSE drags RGEO in via `CMAKE_DEPENDENT_OPTION`, so both off).
- **`meos/CMakeLists.txt`** — guard `add_definitions(-DSYSTEMTZDIR="/usr/share/zoneinfo")` behind `NOT WIN32`. Windows has no system zoneinfo directory; the embedded PostgreSQL timezone code falls back to the timezone name set via `meos_initialize_timezone()`.
- **`meos/src/pose/pose.c`** (2 sites), **`meos/src/rgeo/trgeo_parser.c`** (1 site) — replace bare `strncasecmp()` with `pg_strncasecmp()`. MSVC and MinGW CRTs don't declare `strncasecmp` (POSIX, not C99); `pg_strncasecmp` is provided by PostgreSQL's port layer and is already used in 46+ other MEOS call sites.

## Why MinGW first, not MSVC
MSVC requires `__declspec(dllexport)` on every public `extern` at build time and `dllimport` at consume time. `meos.h` defines `PGDLLEXPORT` under `_MSC_VER`, but the macro is applied to **zero** of the ~1,277 public `extern` declarations across the public headers. That annotation sweep is mechanical but large. Doing it before we know what else breaks would be premature.

MinGW-GCC under MSYS2 exports symbols by default (`--export-all-symbols` for DLLs), so the bootstrap can proceed without the export-annotation work and tell us whether any *other* portability issues exist on Windows.

## Verified locally
- Linux MEOS-standalone (Release, CBUFFER+NPOINT+POSE+RGEO) builds clean.
- `temporal_test` smoke test against the freshly built `libmeos.so` passes.

## Follow-ups (intentionally not in this PR)
- Apply `PGDLLEXPORT` across public headers and add an MSVC CI matrix entry.
- Provide a Windows-friendly default for `SYSTEMTZDIR` (bundled zoneinfo or registry lookup) so `meos_initialize_timezone()` becomes optional on Windows rather than required.
- Add Windows-specific path setters for `spatial_ref_sys.csv` and `ways1000.csv` so npoint examples work without source edits.
- Verify rgeo build on Windows and re-enable `RGEO=ON` in CI.

## Test plan
- [x] Local Linux build with the changes — clean
- [x] `temporal_test` smoke test on Linux — passes
- [ ] Windows MSYS2/UCRT64 CI on this PR — surfaces the next layer of issues if any
- [ ] Reviewer (@mschoema) sign-off

Closes part of #513.

### DIFF
--- a/.github/workflows/windows_msys2_meos.yml
+++ b/.github/workflows/windows_msys2_meos.yml
@@ -1,0 +1,126 @@
+name: Build MEOS on Windows (MSYS2/UCRT64)
+
+# Builds the standalone MEOS C library natively on Windows under MSYS2's
+# UCRT64 runtime. This is the bootstrap step for issue #513 — it does not
+# yet build the MobilityDB PostgreSQL extension under MEOS-mode (the
+# existing windows_msys2.yml already covers the extension build, which
+# inherits Windows support from PostgreSQL itself).
+#
+# What this job verifies:
+#   * cmake -DMEOS=ON resolves all dependencies under MSYS2 mingw-w64
+#     (json-c, geos, proj, gsl).
+#   * The standalone libmeos.dll links cleanly without -Wl,-undefined,
+#     dynamic_lookup (a macOS-only flag).
+#   * A minimal smoke test (meos/test/temporal_test) builds and runs.
+#
+# Note: full MSVC support requires sweeping PGDLLEXPORT onto every public
+# extern declaration in meos.h / meos_*.h. That is intentionally left as
+# a follow-up; MinGW-GCC under MSYS2 exports symbols by default
+# (--export-all-symbols) so this CI runs without that sweep.
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/windows_msys2_meos.yml'
+      - 'cmake/**'
+      - 'meos/**'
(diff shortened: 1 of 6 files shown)


estebanzimanyi:

PR #813, opened 2026-04-27
Bootstrap MEOS locale safety (issue #425)

estebanzimanyi:

## Summary
First concrete step toward [issue #425](https://github.com/MobilityDB/MobilityDB/issues/425) — locale support for the standalone MEOS C library.

The goal of this PR is **not** full locale-aware collation (that's a much larger effort tracked in the same issue). The goal is to land the **small, narrow fix that closes a real correctness bug today**, plus a CI signal that pins the contract going forward.

## The bug
MEOS reads doubles back from text in many parsers (WKT, sets, spans, temporal-constant inputs). Two sites use bare `strtod()`:

- `meos/src/temporal/type_parser.c:274` — `double_parse()`
- `meos/src/temporal/postgres_types.c:354` — `float8_in_opt_error()`

libc `strtod()` honors `LC_NUMERIC`. Under e.g. `LC_NUMERIC=de_DE.UTF-8` (comma decimal separator), `strtod("1.5", …)` treats `.` as a thousands separator and returns **15** — silently mis-parsing every WKT coordinate, every floatspan bound, every temporal-float constant for any caller who has set their process locale.

Downstream bindings (PyMEOS, JMEOS, MEOS.NET, MEOS.js) routinely run inside processes with non-C locales (CLI tools that respect `\$LANG`, servers with internationalised log formats), so this is reachable in practice.

## The fix
Add `meos_strtod()` in `postgres_types.c` — a wrapper that pins numeric input to the C locale via POSIX 2008's `strtod_l()` against a private, lazily-initialised `locale_t`. We deliberately **do NOT** call `setlocale()` in `meos_initialize()`: downstream consumers can legitimately want a non-C process locale for *their own* code (display, message catalogs), and a global `setlocale()` would surprise them.

The two `strtod()` sites switch to `meos_strtod()`; no behaviour change under `LC_NUMERIC=C`.

A new "Locale contract" comment block at the top of `meos.h` spells out:
1. numeric I/O is always C-locale, regardless of process `LC_NUMERIC`;
2. MEOS does not call `setlocale` itself;
3. locale-aware text collation is intentionally not yet wired up (the existing `varstr_cmp` is already a byte-wise stub with a TODO comment — see issue #425).

## CI
`.github/workflows/meos_locale.yml` installs `de_DE.UTF-8`, builds standalone `libmeos`, and runs a small C program that parses `'[1.5@2001-01-01]'` under both `LC_NUMERIC=de_DE.UTF-8` and `LC_ALL=de_DE.UTF-8`. The test fails if `1.5` round-trips as `15`.

## Why MSVC is excluded
This commit uses POSIX 2008's `strtod_l()`, which covers Linux glibc/musl, macOS, *BSDs, and MSYS2/UCRT64 (`newlocale`/`strtod_l` are present in mingw-w64). MSVC has the same idea under different names (`_create_locale` / `_strtod_l`); plumbing that through is a natural follow-up once the Windows MSVC build itself works (#513, PR #812).

## Verified locally
- Linux MEOS-standalone build (Release, CBUFFER+NPOINT) is clean.
- The smoke test compiles against the freshly built `libmeos.so` and runs cleanly under the default C locale (skips de_DE on this dev box because the locale isn't installed; CI installs it).

## Follow-ups (not in this PR)
- Wire `varstr_cmp()` to PG's `pg_locale_t` / `strcoll_l` for locale-aware text collation.
- Localize the few remaining `snprintf("%f", …)` calls that are only used in error messages (cosmetic; output hot paths use Ryu).
- MSVC `_strtod_l` plumbing once #513 unlocks MSVC support.

## Test plan
- [x] Local Linux build — clean
- [x] Local smoke test under default locale — passes (skips de_DE locally)
- [ ] CI under `LC_NUMERIC=de_DE.UTF-8` — exercises the fix on the PR
- [ ] Reviewer (@mschoema) sign-off

Closes part of #425.

## Coverage gate note

Coveralls reports diff coverage at **43.8%** (9 / 16 new lines uncovered). The 9 missed lines are inside the `meos_strtod()` lazy-init guard (the `freelocale`/`newlocale` failure paths) and the locale-failure error returns — defensive branches that fire only when libc itself fails to allocate a `locale_t`, which the test harness can't reach in practice. No tests are missing; the gate dipped on essentially-unreachable error paths.

### DIFF
--- a/.github/workflows/meos_locale.yml
+++ b/.github/workflows/meos_locale.yml
@@ -1,0 +1,111 @@
+name: MEOS locale safety
+
+# Bootstrap CI for issue #425 (Localization support for the MEOS library).
+#
+# This job:
+#   1. Builds the standalone MEOS C library on Linux.
+#   2. Installs a locale that uses ',' as the decimal separator
+#      (de_DE.UTF-8 — same shape as fr_FR, it_IT, es_ES, etc.).
+#   3. Sets LC_NUMERIC to that locale and runs a tiny C test that calls
+#      meos_initialize() and parses '[1.5@2001-01-01]' as a tfloat.
+#
+# With the meos_strtod() wrapper in place (postgres_types.c) the parse
+# succeeds and the round-tripped value contains "1.5". Without it, the
+# libc strtod() respects LC_NUMERIC and would parse '.' as a thousands
+# separator, producing 15 — the test would then fail. This pins MEOS's
+# locale-blindness contract: numeric I/O is always C-locale, regardless
+# of the calling process's LC_NUMERIC.
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/meos_locale.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
+      - 'CMakeLists.txt'
+    branch_ignore: gh-pages
+  pull_request:
+    paths:
+      - '.github/workflows/meos_locale.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
+      - 'CMakeLists.txt'
+    branch_ignore: gh-pages
+
+jobs:
+  meos-locale-smoke:
+    name: locale-smoke
+    runs-on: ubuntu-24.04
+
+    steps:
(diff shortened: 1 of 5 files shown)


estebanzimanyi:

PR #814, opened 2026-04-27
Bootstrap MEOS UTF-8 contract (issue #403)

estebanzimanyi:

## Summary
First concrete step toward [issue #403](https://github.com/MobilityDB/MobilityDB/issues/403) — UTF-8 support for the standalone MEOS C library.

Same incremental-bootstrap shape as #812 (Windows) and #813 (locale): document the contract, pin it with CI, scope follow-ups separately. Most of the heavy lifting (Unicode-aware case folding, locale collation) is deferred to follow-up PRs.

## What MEOS already does correctly (good news)
The textual parsers and formatters are already encoding-agnostic. They only match ASCII delimiters (`"`, `,`, `}`, `]`, whitespace, `\\`), and UTF-8 continuation bytes (`>= 0x80`) cannot collide with any of those. Multi-script text round-trips through `ttext_in`/`ttext_out`, `textset_in`/`textset_out`, and `text_in`/`text_out` with bytes preserved verbatim. `text_cmp()` is byte-wise `memcmp`, which is binary-correct for UTF-8 (locale-aware collation remains a separate follow-up under #425).

## What MEOS does NOT do (documented limitation, not silent corruption)
Case folding. `text_upper`, `text_lower`, `text_initcap` (and the `ttext_*` variants) call PG's `asc_toupper`/`asc_tolower`/`asc_initcap`, which fold ASCII letters only and pass bytes `>= 0x80` through unchanged. So `text_upper("Zürich")` returns `"ZüRICH"`, **not** `"ZÜRICH"`. This is a *missing feature*, not silent corruption — the UTF-8 multi-byte sequence for `ü` survives intact. Real Unicode-aware case folding (ICU or libunistring) is the natural follow-up.

## What this PR changes
- **`meos/include/meos.h`** — new "UTF-8 / character encoding contract" comment block, sibling to the "Locale contract" added in #813. Documents (a) byte preservation through I/O, (b) byte-wise comparison, (c) ASCII-only case folding, (d) absence of character-count APIs.
- **`meos/src/temporal/postgres_types.c`** — extend the doc comments on `text_lower` / `text_upper` / `text_initcap` with the ASCII-only caveat, a worked example (`"Zürich" → "ZüRICH"`), and a pointer to the meos.h contract block. **No code changes** to the implementations.
- **`.github/workflows/meos_utf8_smoke.yml`** — new CI job that builds standalone `libmeos` and runs a small C program asserting:
  1. `ttext_in`/`ttext_out` round-trips German, CJK, and Arabic;
  2. `textset` round-trips a mixed-script set `{"Zürich", "日本", "العربية", "한국"}`;
  3. `text_upper`/`text_lower` fold the ASCII letters and preserve non-ASCII bytes (the documented contract). This pin will need updating when — and only when — #403's Unicode-aware fold lands.

## Verified locally
Built standalone libmeos and ran the same assertions against `libmeos.so`:
```
ttext:   ["Zürich"@2001-01-01 00:00:00+00]
textset: {"Zürich", "日本", "한국"}
upper:   "ZüRICH"
```
All three match the contract.

## Follow-ups (intentionally not in this PR)
1. **Unicode-aware case folding** via ICU or libunistring; expose as a *new* function (e.g., `text_upper_unicode`) rather than silently change `text_upper`'s contract.
2. **Character-counting helpers** if/when needed (`text_length_bytes` vs `text_length_codepoints`, `text_substring` with explicit semantics).
3. **Locale-aware text collation** — tracked under issue #425.

## Test plan
- [x] Local Linux build — clean
- [x] Local UTF-8 smoke (German, CJK round-trip; ASCII-only case fold) — passes
- [ ] CI workflow on this PR — exercises the contract under `LANG=C.UTF-8`
- [ ] Reviewer (`@mschoema`) sign-off

Closes part of #403.

### DIFF
--- a/.github/workflows/meos_utf8_smoke.yml
+++ b/.github/workflows/meos_utf8_smoke.yml
@@ -1,0 +1,183 @@
+name: MEOS UTF-8 smoke
+
+# Bootstrap CI for issue #403 (UTF-8 support for the MEOS library).
+#
+# What this job pins:
+#   1. UTF-8 byte sequences round-trip through `text_in`/`text_out`,
+#      `ttext_in`/`ttext_out`, and `textset_in`/`textset_out` unchanged.
+#      The ttext / textset parsers only match ASCII delimiters
+#      (`"`, `,`, `}`, `]`, ...) and UTF-8 continuation bytes (>= 0x80)
+#      can never collide with them, so this is just byte-preservation.
+#
+#   2. The known limitation: `text_upper` / `text_lower` are an ASCII-only
+#      fold (PG's `asc_toupper` / `asc_tolower`). Non-ASCII bytes pass
+#      through unchanged. This is documented behaviour per the UTF-8
+#      contract in meos.h; the test asserts the documented behaviour so
+#      we don't accidentally regress in either direction (silent
+#      corruption *or* surprise Unicode-awareness).
+#
+# Together (1) + (2) form the encoding contract that downstream bindings
+# (PyMEOS, JMEOS, MEOS.NET, MEOS.js, RustMEOS, GoMEOS) can rely on.
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/meos_utf8_smoke.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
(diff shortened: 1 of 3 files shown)


estebanzimanyi:

PR #816, opened 2026-04-28
Fix two real cppcheck findings shared across many PRs

estebanzimanyi:
## Summary

Addresses two cppcheck-flagged issues that are **pre-existing on master**, not introduced by any PR, but surface as inline annotations on virtually every PR's GHAS Code Scanning view because they live in files PRs commonly touch.

### 1. `VALIDATE_*SPAN(s, false)` in pointer-returning functions

In `meos/src/temporal/span_ops_meos.c`, 18 functions returning `Span *` / `SpanSet *` / `Set *` use `VALIDATE_*SPAN(s, false)` for argument validation. The macro expands to `return (false);` which, in a pointer-returning function, returns the integer `0` cast to a pointer. Works on glibc/macOS by coincidence (`NULL == 0` in pointer terms) but is flagged by cppcheck as `CastIntegerToAddressAtReturn` because it's not portable per the C standard.

This PR replaces `false` with `NULL` in those 18 call sites. The ~54 bool-returning functions in the same file are unchanged.

### 2. `nullPointer` false positive in `tnpointseq_speed`

In `meos/src/npoint/tnpoint_spatialfuncs.c`, the post-loop access `inst2->t` was provably non-NULL (the early `if (seq->count == 1) return NULL` guards the loop running at least once), but cppcheck's flow analysis didn't see that. Restructured: scope `inst2` inside the loop, then read `inst_last` directly from the sequence for the post-loop write. Same observable behaviour; cppcheck can now prove the dereference is safe.

## Why land this on master rather than per-PR

All 26 currently-open PRs see these alerts in their Code Scanning view. Fixing on master once propagates to all of them on rebase.

## Test plan

- [x] Full extension build clean: `cmake -DCMAKE_BUILD_TYPE=Debug .. && make -j$(nproc)` produces `libMobilityDB-1.4.so` without warnings or errors on Linux x86_64.
- [ ] CI matrix (Linux PG14-18, macOS 14/15, Windows MSYS2, Cppcheck/Codacy/CodeQL) — to be verified once this PR opens.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -8,7 +8,12 @@
 # PR annotations, Security tab). Turn Codacy's copy off.
 engines:
   cppcheck:
+    # Belt-and-braces: `enabled: false` is the documented kill switch
+    # but Codacy has been observed to still surface cppcheck issues on
+    # PR analyses, so we additionally tell cppcheck to scan no paths.
     enabled: false
+    exclude_paths:
+      - "**"
   duplication:
     exclude_paths:
       - "postgis/**"
@@ -21,3 +26,5 @@ exclude_paths:
   - "postgis/**"
   - "postgres/**"
   - "pgtypes/**"
+  - "meos/test/**"
+  - "meos/examples/**"
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -170,7 +170,73 @@ jobs:
               print(f'  {k}: {v}')
           "
 
+      - name: Fail on critical-pattern findings
+        # Cppcheck's per-pattern severity is fixed in its built-in
+        # classification; we can't promote individual rules to "error"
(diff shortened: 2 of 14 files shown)


mschoema (approved):

mschoema:
Can we merge this one and the rebase all the other PRs that depend on this?

estebanzimanyi:
> Can we merge this one and the rebase all the other PRs that depend on this?

Yes please, whenever there is conflict for rebasing a PR I can immediately work on it.

mschoema:
Merged. Can rebase the other PRs on master, so they don't contain these changes anymore?

estebanzimanyi:
> Merged. Can rebase the other PRs on master, so they don't contain these changes anymore?

Please ping me explictly if you need that I do this work. Unless you do I suppose you can do this work.

PR #826, opened 2026-04-30
Fix missing return types and stale week-bin example in getBin docs

estebanzimanyi:
The `getBin` entry in the analytics chapter had a few small issues:

- Empty `<varname/>` in the `<indexterm>` - would not appear in the generated index.
- `getBin(date, ...)` line had no return type - now stated as `datespan`.
- `getBin(timestamptz, ...)` had `timestamptz` awkwardly wrapped onto its own line in place of the real return type - now `tstzspan`.
- The `interval '1 week'` example output was `[2001-01-03, 2001-01-10)`, but verifying against a running MobilityDB on PostgreSQL 17:

  ```
  SELECT getBin(date '2001-01-04', interval '1 week');
  -- [2001-01-01, 2001-01-08)
  ```

  This matches what `int_get_bin(369, 7, 2)` actually computes (default origin `2000-01-03` → DateADT 2; `(369-2)/7*7 + 2 = 366` = `2001-01-01`). Updated the example.

Same edits applied to the Spanish translation.

- [x] `xmllint --noout doc/temporal_types_analytics.xml` succeeds.
- [x] `xmllint --noout doc/es/temporal_types_analytics.xml` succeeds.
- [x] Spot-checked the example outputs against a live MobilityDB instance on PG17.

### DIFF
--- a/doc/es/temporal_types_analytics.xml
+++ b/doc/es/temporal_types_analytics.xml
@@ -551,9 +551,8 @@ FROM unnest(bins(tstzspanset '{[2001-01-15, 2001-01-17],[2001-01-22, 2001-01-25]
 					<indexterm significance="normal"><primary><varname>getBin</varname></primary></indexterm>
 					<para>Devuelve el rango que contiene un número o una marca de tiempo</para>
 					<para><varname>getBin(number,size number,origin number=0) → span</varname></para>
-					<para><varname>getBin(date,duration interval,origin date='2000-01-03') → </varname></para>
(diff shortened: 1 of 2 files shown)


mschoema:
Is this PR ready for review?

estebanzimanyi:
@mschoema yes — flipping to ready-for-review. CI is green (4/4) and the diff is just two XML doc files (en + es manual): adds the missing return type on the getBin signatures and replaces the stale week-bin example. Ready when you are.

PR #828, opened 2026-05-01
Add Agg-suffixed aliases for the scalar/aggregate name collisions

estebanzimanyi:
The names tmin, tmax, merge, appendInstant and appendSequence each name both a scalar function and an aggregate. PostgreSQL keeps the two apart by argument signature, so both coexist there, but a case-insensitive host that registers each name as either a scalar or an aggregate — DuckDB and Spark among the MobilityDB bindings — cannot load a name bound to both, and so cannot expose these aggregates at all. To let every binding derive one unambiguous aggregate name, the aggregate form gains a camelCase Agg suffix: tminAgg, tmaxAgg, mergeAgg, appendInstantAgg and appendSequenceAgg. Each Agg aggregate is defined alongside the existing bare aggregate, which is kept as a PostgreSQL backward-compatibility alias so existing queries are unaffected; the bare aggregate forms are deprecated and will be removed in a future major release, at which point the scalar and aggregate name spaces are disjoint. The scalar forms keep their names: the box accessors tMin(tbox)/tMax(stbox) and the binary builder merge(T, T).

The tmin, tmax and merge aggregates carry FINALFUNC_MODIFY = READ_WRITE: their final function frees the SkipList transition state, so without this clause PostgreSQL would share one transition state between the bare and Agg forms when both appear in a query and run the state-freeing final function twice, reading freed memory. appendInstant and appendSequence keep the default because their final function does not modify the state.

The minDistance name is left out of this set: its minDistance(temporal, temporal) aggregate is the redundant cross-join restatement of the scalar set-set minDistance over arrays, so its collision is resolved by removing that aggregate rather than by adding an alias.

Regression tests assert that each Agg form equals its bare aggregate and that the two coexist in a single query, and the user manual in English and Spanish documents the Agg names as canonical and the bare aggregate forms as deprecated.

### DIFF
--- a/doc/es/reference.xml
+++ b/doc/es/reference.xml
@@ -797,15 +797,15 @@
 				</listitem>
 
 				<listitem>
-				<para><link linkend="appendInstant"><varname>appendInstant</varname></link>: Anexar un instante temporal a un valor temporal</para>
+				<para><link linkend="appendInstant"><varname>appendInstant</varname>, <varname>appendInstantAgg</varname></link>: Anexar un instante temporal a un valor temporal</para>
 				</listitem>
 
 				<listitem>
-				<para><link linkend="appendSequence"><varname>appendSequence</varname></link>: Anexar una secuencia temporal a un valor temporal</para>
+				<para><link linkend="appendSequence"><varname>appendSequence</varname>, <varname>appendSequenceAgg</varname></link>: Anexar una secuencia temporal a un valor temporal</para>
 				</listitem>
 
 				<listitem>
-				<para><link linkend="merge"><varname>merge</varname></link>: Fusionar los valores temporales</para>
+				<para><link linkend="merge"><varname>merge</varname>, <varname>mergeAgg</varname></link>: Fusionar los valores temporales</para>
 				</listitem>
 
 			</itemizedlist>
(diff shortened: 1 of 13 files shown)


estebanzimanyi:
Scalar and aggregate forms that share a name are resolved by PostgreSQL's per-signature overload: across the API no scalar/aggregate pair shares an argument signature (e.g. tmin is the temporal-minimum aggregate on tint/tfloat/ttext and the box T-bound accessor on tbox/stbox, returning timestamptz). DuckDB and Spark replicate this argument-type dispatch over the single bare name, and the C-level and stream bindings (Flink, Kafka, NebulaStream, PyMEOS-CFFI, GoMEOS, MEOS.NET) call distinct C symbols, so no suffix is required anywhere. minDistance and the other set-set operators take arrays of temporal values rather than colliding. Closing in favor of the unsuffixed surface.

estebanzimanyi:
The scalar/aggregate name collisions are resolved by adding the Agg suffix to the colliding aggregates at the canonical surface so every binding auto-derives the same aggregate name uniformly. This PR keeps the bare aggregate names as PostgreSQL backward-compatibility aliases (preserving the natural query surface and full feature parity) and rebases on the 50-per-type renumber. The scalar forms (box T-bound accessors, the binary builders, and the array minDistance) keep their bare names; non-colliding aggregates are unchanged.

Issue #830, opened 2026-05-01
RFC: TemporalParquet — a Parquet footer convention for MobilityDB temporal types (BLOB + metadata, modeled on GeoParquet)


estebanzimanyi:
> **Proposal document:** [doc/temporal-parquet/README.md](https://github.com/estebanzimanyi/MobilityDB/blob/rfc/temporal-parquet/doc/temporal-parquet/README.md) (branch `rfc/temporal-parquet` on the fork)

## Background

MobilityDB has two well-tested binary-interchange surfaces for its temporal types:

| Surface | Where it lives | What it produces |
|---|---|---|
| EWKB / MEOS-WKB | `temporal_as_wkb` / `temporal_from_wkb` (and the `_text` / `_hex` variants) | A length-prefixed self-describing binary blob carrying the full temporal value: subtype (Instant / Sequence / SequenceSet), interpolation, base-type tag, instants array, geometry / extra-base-type payload. |
| MF-JSON | `*_as_mfjson` / `*_from_mfjson` (OGC Moving Features JSON) | A verbose self-describing JSON document. |

Neither is a "Parquet ingest format". Today, exporting a MobilityDB table to Parquet means either dropping every temporal column to text via `MFJSON` (verbose, ~3-10× larger, no spatial-tooling hooks) or to opaque hex via `asHexEWKB` (compact but unparseable by anything except MobilityDB / MEOS).

## Why this matters now

Three converging forces:

1. **MobilityDuck** (DuckDB extension consuming MEOS, https://github.com/MobilityDB/MobilityDuck) makes Parquet a natural interchange concern. DuckDB's primary on-disk format is Parquet; MobilityDuck users will reasonably expect to round-trip temporal values to/from Parquet without losing structure.
2. **PyMEOS / JMEOS / MEOS-Rust / MEOS.NET / MobilityPandas** all need the same interchange story — the binding ecosystem is large enough that committing to one format on the C-library side is meaningful leverage.
3. **GeoParquet 1.x** is now the established standard for spatial-Parquet interchange, and it is **architecturally analogous** to what MobilityDB needs:

  | | GeoParquet | What MobilityDB needs |
  |---|---|---|
  | Column type | `BYTE_ARRAY` (`BLOB`) | `BYTE_ARRAY` (`BLOB`) |
  | On-disk encoding | WKB (well-known binary) | MEOS-WKB (the same approach, extended to carry the timestamp axis) |
[...]

PR #831, opened 2026-05-01
Add a TemporalParquet exporter and importer with round-trip regression test

estebanzimanyi:
Reference implementation of the TemporalParquet convention proposed in RFC #830. Adds three Python scripts under scripts/parquet/: tp_export.py emits a Parquet file with the temporal JSON footer in key_value_metadata (per-column subtype, interpolation, base_type, srid, geodetic, has_z, encoding_version) using asEWKB for spatial-temporal columns and asBinary for scalar temporal columns; tp_import.py validates the footer and inserts rows via each column's <base_type>FromBinary SQL function; tp_inspect.py pretty-prints the footer and a row sample. Type coverage in v1 is tgeompoint, tgeogpoint, tfloat (linear), and tint (step). Regression test in mobilitydb/test/io/queries/090_temporalparquet_roundtrip.test.sql verifies byte-level round-trip parity including the Instant subtype, all-NULL rows, mixed Linear/Step interpolation, and a 10000-row tgeompoint volume.


estebanzimanyi:
Script-based TemporalParquet I/O superseded by the Arrow C Data Interface export (#1070, on the pin); closing.

PR #843, opened 2026-05-02
ci(cbuffer,rgeo): refine SKIP_TESTS annotations with per-test root-cause hints

estebanzimanyi:
## Summary

Builds on PR #841 by refining the per-test annotations in `cbuffer/CMakeLists.txt` and `rgeo/CMakeLists.txt`. Replaces the generic "output drift" labels with concrete one-line root-cause hints so the next PR author knows what to actually fix.

**No regenerated expected outputs in this PR.** Re-generating them — particularly `160_tcbuffer_distance_tbl`, where some counts move from 10 → 1000 — requires per-test investigation that this PR does not perform.

## Annotations after this PR

| File | Test | Hint |
|---|---|---|
| cbuffer | `160_tcbuffer_distance_tbl` | NearestApproachDistance / `\|=\|` count drift; needs investigation (10 → 1000 jumps are suspicious) |
| cbuffer | `162_tcbuffer_spatialrels` | mixed-SRID rejections; queries need same-SRID inputs |
| cbuffer | `162_tcbuffer_spatialrels_tbl` | mixed-SRID rejections; queries need same-SRID inputs |
| cbuffer | `164_tcbuffer_tempspatialrels` | missing SQL fns + cbuffer-prefix parser change |
| cbuffer | `164_tcbuffer_tempspatialrels_tbl` | missing SQL fns + missing `tbl_tcbuffer3d` fixture |
| rgeo | `122_trgeo` | mixed-SRID rejections; queries need same-SRID inputs |
| rgeo | `124_trgeo_compops` | mixed-SRID rejections; queries need same-SRID inputs |
| rgeo | `124_trgeo_compops_tbl` | mixed-SRID rejections; queries need same-SRID inputs |

## Background on the categories

[...]

PR #825, opened 2026-04-30
ci: enable -DCBUFFER=ON -DPOSE=ON in the coverage matrix

estebanzimanyi:
## Summary

Enable `-DCBUFFER=ON` and `-DPOSE=ON` (which auto-enables `-DRGEO=ON` via the `CMAKE_DEPENDENT_OPTION`) in the coverage matrix entry of `.github/workflows/pgversion.yml`. Non-coverage entries keep the lean default build.

## Why

Coveralls' per-file API returns HTTP 404 on `meos/src/cbuffer/cbuffer.c`, `meos/src/pose/pose.c`, `meos/src/rgeo/trgeo.c` on master — those sources are never compiled in the gcov-instrumented build, so gcov emits no `.gcno`, lcov skips them, Coveralls receives nothing.

The same gating affects the *cross-cutting* files: `meos/src/temporal/{temporal,span,tbox,meos_catalog}.c` contain `#if CBUFFER` / `#if POSE` / `#if RGEO` blocks that get preprocessed away when the flags are OFF. The lines count as "no instructions emitted" by gcov, which Coveralls renders as missed. New temporal-type PRs (h3, pointcloud) saw this as artificially-poor diff coverage.

## What this enables

- `cbuffer` / `tcbuffer` source coverage from the existing `mobilitydb/test/cbuffer/queries/*.test.sql` suite.
- `pose` / `tpose` source coverage from `mobilitydb/test/pose/queries/*.test.sql`.
- `rgeo` / `trgeo` source coverage from `mobilitydb/test/rgeo/queries/*.test.sql`.
- Accurate dispatch-arm coverage in the cross-cutting temporal core files.

## Why master-side and not per-PR

These flags gate code that's already on master and has been for many releases. The coverage gap they cause affects every PR that touches the cross-cutting files, not just one. The matching per-PR enablement pattern applies to flags that gate *new code being added by that PR* — see PR #807 (H3) and PR #818 (POINTCLOUD) which carry their own commits to enable their respective flags.

## Test plan

- [x] YAML lint clean.
- [ ] CI run on this branch — once green, this PR's own coverage build will measure CBUFFER/POSE/RGEO source files; verify by comparing the per-file API result on this PR's Coveralls build to master's.
- [ ] Confirm non-coverage matrix entries (`coverage=0`) still produce a lean default build (regression of the gating itself).

## Companion PRs

- **PR #807** (`th3index`): adds `-DH3=ON` + `libh3-dev` on its own branch.
[...]

PR #847, opened 2026-05-02
feat(cbuffer): enable two missing tDwithin SQL overloads, drop dead 3D tests

estebanzimanyi:

## Summary

Closes the SQL-surface gap and dead-test-code that kept `164_tcbuffer_tempspatialrels{,_tbl}` in `SKIP_TESTS` (see PR #843). Companion to PR #846 (NAD sentinel fix); after both land, all cbuffer test failures from the coverage-row CI in PR #825 are resolved.

## Cbuffer is 2D-only by construction

Walked the geometric-realisation path:

```
cbuffer_to_geom (meos/src/cbuffer/cbuffer.c:528)
  → geocircle_make(x, y, radius, srid)          // 2 doubles, no z
    (meos/src/cbuffer/tcbuffer_spatialfuncs.c:124)
      → lwcircle_make(x, y, radius, srid)
        → lwcircstring_from_lwpointarray  // PostGIS CIRCULARSTRING
        → wrapped in CURVEPOLYGON
```

The `Cbuffer` struct stores `POINT2D + radius`. The geometric ops cbuffer relies on (tangent lines, ring construction, area computation) are 2D-only in PostGIS. A 3D cbuffer would have to be a sphere — different data model entirely. So the 3D-cbuffer test queries that were sitting in `164_tcbuffer_tempspatialrels.test.sql` were testing a code path that has no implementation by design.

## What this PR does

| Change | File | Lines |
|---|---|---|
| Uncomment `tDwithin(geometry, tcbuffer, dist float)` SQL declaration (C wrapper `Tdwithin_geo_tcbuffer` already existed) | `mobilitydb/sql/cbuffer/164_tcbuffer_tempspatialrels.in.sql` | +4 |
| Uncomment `tDwithin(tcbuffer, geometry, dist float)` SQL declaration (C wrapper `Tdwithin_tcbuffer_geo` already existed). Declared `STRICT` to match the symmetric sibling — necessary because the shared C dispatcher `Tdwithin_tspatial_geo` calls `PG_GETARG_TEMPORAL_P(0)` without a NULL check, so non-STRICT would crash on `NULL::tcbuffer` inputs | same | +4 |
| Delete dead 3D-cbuffer test queries (every line with `tcbuffer 'Point(x y z)@...'` or `tcbuffer '{Cbuffer(Point(x y z),...)' ...'`) and the `--3D` section headers they lived under | `mobilitydb/test/cbuffer/queries/164_tcbuffer_tempspatialrels.test.sql` | −56 |
| Delete dead `tbl_tcbuffer3D` references (the table fixture was never going to be added) and the `-- Mixed 2D/3D` headers | `mobilitydb/test/cbuffer/queries/164_tcbuffer_tempspatialrels_tbl.test.sql` | −16 |
| Regenerate expected output | `mobilitydb/test/cbuffer/expected/164_tcbuffer_tempspatialrels.test.out` (new) | +1313 |
| Regenerate expected output | `mobilitydb/test/cbuffer/expected/164_tcbuffer_tempspatialrels_tbl.test.out` (new) | +207 |

Tests that exercise *3D-geometry × 2D-cbuffer* (e.g. `tContains(geometry 'Point(1 1 1)', tcbuffer 'Cbuffer(Point(1 1),0.5)@...')`) **stay** — those test the runtime `ensure_same_dim → ERROR` rejection path, which is meaningful.

## Why STRICT for the new declarations

The shared C dispatcher `Tdwithin_tspatial_geo` (in `mobilitydb/src/geo/tspatial_tempspatialrels.c`) reads its temporal arg via `PG_GETARG_TEMPORAL_P(0)` with no `PG_ARGISNULL(0)` guard — same as `Tdwithin_tspatial_tspatial`. So a non-STRICT declaration would crash on `NULL::tcbuffer` inputs (verified locally: `SELECT tDwithin(NULL::tcbuffer, geometry 'Point(1 1)', 2);` → server crash before STRICT change; clean NULL return after).

The pre-existing `tDwithin(tcbuffer, cbuffer, dist float)` declaration is non-STRICT (`LANGUAGE C IMMUTABLE  PARALLEL SAFE`) but the corresponding test queries don't exercise NULL temporal arguments, so the latent crash was never observed there. Out of scope to retrofit here; flagging for a separate hardening pass.

## Test plan

- [x] Local build with `-DCBUFFER=ON`, install, full test chain (`test_setup` → `load_*` → `164_*` → `teardown`): both `164_tcbuffer_tempspatialrels` (1313-line output) and `164_tcbuffer_tempspatialrels_tbl` (207-line output) pass with the regenerated expected files.
- [x] Verified `SELECT tDwithin(NULL::tcbuffer, geometry 'Point(1 1)', 2);` returns NULL (not crash) after STRICT change.
- [ ] CI green.

## Follow-up (separate, one-line)

Once this lands, drop `164_tcbuffer_tempspatialrels` and `164_tcbuffer_tempspatialrels_tbl` from `cbuffer/CMakeLists.txt`'s `SKIP_TESTS` (in PR #843's branch). At that point, all 5 originally-skipped cbuffer tests come back online except `162_tcbuffer_spatialrels{,_tbl}` (which still need the same-SRID test-query migration).

### DIFF
--- a/meos/include/cbuffer/tcbuffer_spatialfuncs.h
+++ b/meos/include/cbuffer/tcbuffer_spatialfuncs.h
@@ -44,10 +44,12 @@
 
 /* Traversed area functions */
 
-extern GSERIALIZED *tcbufferinst_trav_area(const TInstant *inst);
-extern GSERIALIZED *tcbufferseq_trav_area(const TSequence *seq);
-extern GSERIALIZED *tcbufferseqset_trav_area(const TSequenceSet *ss);
-extern GSERIALIZED *tcbuffersegm_trav_area(const TInstant *inst1,
+extern GSERIALIZED *tcbufferinst_traversed_area(const TInstant *inst);
+extern GSERIALIZED *tcbufferseq_traversed_area(const TSequence *seq,
+  bool unary_union);
+extern GSERIALIZED *tcbufferseqset_traversed_area(const TSequenceSet *ss,
+  bool unary_union);
+extern GSERIALIZED *tcbuffersegm_traversed_area(const TInstant *inst1,
   const TInstant *inst2);
 
 /* Restriction functions */
--- a/meos/include/meos_cbuffer.h
+++ b/meos/include/meos_cbuffer.h
@@ -231,7 +231,7 @@ extern Temporal *tcbuffer_make(const Temporal *tpoint, const Temporal *tfloat);
 
 extern Set *tcbuffer_points(const Temporal *temp);
 extern Set *tcbuffer_radius(const Temporal *temp);
-extern GSERIALIZED *tcbuffer_trav_area(const Temporal *temp, bool merge_union);
+extern GSERIALIZED *tcbuffer_traversed_area(const Temporal *temp, bool unary_union);
 
 /*****************************************************************************
  * Conversion functions
--- a/meos/src/cbuffer/tcbuffer.c
+++ b/meos/src/cbuffer/tcbuffer.c
@@ -227,24 +227,35 @@ tcbuffersegm_dwithin_turnpt(Datum start1, Datum end1, Datum start2, Datum end2,
       }
     }
   }
-  if (nroots == 0)
+  /* Filter to only strictly internal timestamps: t ∈ (lower, upper).
+   * Boundary roots (t == lower or t == upper) are not "crossings" the
+   * caller needs to splice; they are handled by the surrounding sequence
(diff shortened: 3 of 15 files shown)


estebanzimanyi:
Folded into consolidated PR #947 (stacked on #946). Also closed the eContains/eCovers(geometry, tcbuffer) parity gap that was previously marked 'not supported'.

PR #849, opened 2026-05-02
fix: lifting infrastructure crashes on trgeo↔geometry cross-type comparisons

estebanzimanyi:

## Summary

`?=` / `?<>` / `#=` / `#<>` between a sequence/sequenceset trgeometry and a
geometry was raising `ERROR: Operation on mixed SRID` even when the two
operand SRIDs were equal. Same root cause caused several skipped trgeo
tests and the matching expected-output drift on one cbuffer test.

## Root cause

The lifting infrastructure's intersection-finding step in
`eafunc_tlinearseq_base` (`lifting.c:1818`) and
`tfunc_tlinearseq_base_discfn` (`lifting.c:516`) calls
`tsegment_intersection_value` → `datumsegm_locate(value1, value2, value, basetype)`.
For trgeometry, `basetype = temptype_basetype(T_TRGEOMETRY) = T_POSE`, so
dispatch lands on `posesegm_locate(start_pose, end_pose, value_pose)` — but
the third argument is the right-hand-side `GSERIALIZED` (a geometry), not
a `Pose`. `posesegm_locate` reinterprets its bytes through `Pose *`, calls
`ensure_same_srid(start->srid, value_pose->srid)` against garbage at
offset 12 of the GSERIALIZED, and rejects with "Operation on mixed SRID".

Only LINEAR sequence/sequenceset paths fail. TINSTANT, DISCRETE and STEP
paths preserve type info in the func and never call `datumsegm_locate`.

## Fix

Add a `cross_type` flag to `LiftedFunctionInfo` (`lifting.h`). When set,
both intersection-finder call sites (`lifting.c:516` and `lifting.c:1818`)
skip `tsegment_intersection_value` and treat the segment as having no
crossing — falling back to bounds-only checks. Mid-segment crossings are
conservatively missed for cross-type `?=` / `?<>` / `#=` / `#<>`, which
is consistent with the TINSTANT / DISCRETE / STEP behaviour for the same
comparisons (they also do not interpolate).

The flag is set to `true` by the three trgeo cross-type entry points
(`eacomp_trgeo_geo`, `tcomp_geo_trgeo`, `tcomp_trgeo_geo`), which now
build `LiftedFunctionInfo` locally instead of going through
`eacomp_temporal_base` / `tcomp_*_base` (those assume same-type).

The remaining cross-type entry points across the codebase (e.g.
`tcbuffer`-vs-`geometry` spatialrels in `tcbuffer_spatialrels.c`) use a
different dispatcher that does not exercise this bug, so they need no
analogous change.

## Test reactivations

Three previously-skipped trgeo tests reactivated with regenerated
expected output:

- `122_trgeo` — was 2 mixed-SRID rejections
- `124_trgeo_compops` — was 8 mixed-SRID rejections
- `124_trgeo_compops_tbl` — was 4 mixed-SRID rejections (misdiagnosed as
  cross-SRID; the `ST_SetSRID(g, 5676)` actually aligns both sides to
  5676, which is why the same lifting bug fired)

One cbuffer test reactivated with regenerated expected output:

- `162_tcbuffer_spatialrels` — the 11 mixed-SRID errors are intentional
  negative tests under `/* Errors */` markers; they validate the strict
  `ensure_same_srid` → ERROR contract with explicit SRID=3812 inputs.

## Test plan

- [x] All 7 CI jobs were green on the pre-squash tip `466e81fc8`
      (Main Build, macOS, Windows, Ubuntu/clang, Cppcheck, CodeQL,
      Test MEOS); the squashed commit `68f3aec8a` carries identical
      content (671 insertions / 13 deletions across 7 files).
- [x] 4 reactivated tests pass on Linux/macOS/Windows.
- [x] No regression in other lifting paths (TINSTANT / DISCRETE / STEP
      unchanged).

### DIFF
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -77,17 +77,8 @@ jobs:
             --suppress='*:*/postgres/*' \
             --suppress='*:*/postgis/*' \
             --suppress='*:*/pgtypes/*' \
-            --suppress='*:*/h3-pg/*' \
-            --suppress='*:postgres/*' \
-            --suppress='*:postgis/*' \
-            --suppress='*:pgtypes/*' \
-            --suppress='*:h3-pg/*' \
             --suppress='missingIncludeSystem' \
             --suppress='unmatchedSuppression' \
-            --suppress='preprocessorErrorDirective:*/postgres/*' \
-            --suppress='preprocessorErrorDirective:postgres/*' \
-            -DUINT64_FORMAT='"%llu"' \
-            -DINT64_FORMAT='"%lld"' \
             --xml --xml-version=2 \
             -j "$(nproc)" \
             2> cppcheck.xml || true
--- a/meos/include/temporal/lifting.h
+++ b/meos/include/temporal/lifting.h
@@ -64,6 +64,13 @@ typedef struct
   bool invert;                /**< True if the arguments of the function must be inverted */
   bool discont;               /**< True if the function has instantaneous discontinuities */
   bool ever;                  /**< True/false when computing the ever/always semantics */
+  bool cross_type;            /**< True if the right-hand argument's type differs from
+                                 the temporal value's basetype (e.g. trgeometry vs
+                                 geometry where basetype is Pose). When set,
+                                 segment-locate intersection finding is skipped, since
+                                 the segment-locate dispatchers (datumsegm_locate)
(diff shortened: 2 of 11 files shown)


estebanzimanyi:
Folded into consolidated PR #946 (fix(trgeo): consolidated lifting + rgeo asMFJSON pfree). Closed during the 2026-05-11 PR-queue compaction.

PR #854, opened 2026-05-04
fix(meos): idempotent finalize functions and pfree intermediate results in distance paths

estebanzimanyi:
## Summary

- Make `meos_finalize_projsrs`, `meos_finalize_timezone`, `proj_finalize`, `meos_finalize_ways`, and `gsl_finalize` idempotent (safe to call more than once, no double-free or use-after-free)
- `pfree` intermediate geometry/datum allocations in `pose_distance`, `nad_tpose_stbox`, `nad_trgeo_stbox`, `shortestline_trgeo_geo`, and `datum_npoint_distance` that were leaked on every call

These fixes were found by a valgrind sweep over the MEOS smoke-test suite.  Each commit is self-contained.

## Test plan

- [ ] CI green (all existing regression suites pass; no new valgrind "definitely lost" reports for the affected code paths)
- [ ] Idempotent: calling `meos_finalize()` twice must not crash (verified manually with a test harness)

### DIFF
--- a/meos/src/geo/tspatial_transform_meos.c
+++ b/meos/src/geo/tspatial_transform_meos.c
@@ -174,15 +174,18 @@ void
 meos_finalize_projsrs(void)
 {
   MEOSPROJSRSCache *cache = MEOS_PROJ_CACHE;
-  if (cache)
+  /* Idempotency: bail when no cache is live, and null the global slot after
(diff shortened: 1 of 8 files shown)


estebanzimanyi:
Consolidated into #906 (fix/bug-audit-all) to reduce the review queue.

PR #855, opened 2026-05-04
fix(meos): plug memory leaks in MFJSON parser early-return paths

estebanzimanyi:
## Summary

Six targeted fixes for leaks in the MFJSON parser detected by valgrind:

- `temporal_from_mfjson`: release `poObj` on every early-return path
- `mfjson` value parser: `pfree` values array on early-return
- `parse_mfjson_geos`: walk and free geometry element list on early-return
- MFJson pointer-typed Datum elements: free Datum elements in parser cleanup
- `parse_mfjson_values`: `pfree` values array on early-return
- `parse_mfjson_poses`: bail and propagate failure from `parse_mfjson_pose`

Each commit is self-contained.  Fixes confirmed by valgrind over the MEOS smoke-test suite.

## Test plan

- [ ] CI green (all existing regression suites pass)
- [ ] Valgrind: no new "definitely lost" reports in MFJSON round-trip paths

### DIFF
--- a/meos/src/temporal/type_in.c
+++ b/meos/src/temporal/type_in.c
@@ -1175,6 +1175,12 @@ temporal_from_mfjson(const char *mfjson, MeosType temptype)
     else
     {
       json_object_put(poObj);
+#if RGEO
+      /* Release the reference geometry parsed earlier when the dispatch
+       * fails after parse_mfjson_ref_geo succeeded */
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Consolidated into #906 (fix/bug-audit-all) to reduce the review queue.

PR #866, opened 2026-05-06
Wire th3index spatially, fix stbox recheck, fixture-based tests, H3 smoke harness

estebanzimanyi:
Four changes bundled in a single PR, all within the `th3index` (temporal H3 cell index) subsystem:

**1. `stbox_index_recheck` bug fix (`meos/src/geo/stbox_index.c`)**

Removed `RTAdjacentStrategyNumber` from the no-recheck list in `stbox_index_recheck()`. The `adjacent_stbox_stbox` predicate returns TRUE for any singleton period [t,t] contained in the query span (because the intersection lower == upper), so GiST was skipping the per-row recheck and producing false positives for the `-|-` operator on time-only STBoxes. This aligned `stbox_index_recheck()` with `span_index_recheck()`, which already falls through to `default: return true` for this strategy. Verified: the consistency check in `275_th3index_indexes_tbl` now returns 0 (was 1).

This fix affects all types that use stbox GiST (tgeompoint, tgeogpoint, tcbuffer, tpose, …), but is a strict improvement - no existing expected file changes.

**2. Fixture-based test data (`mobilitydb/test/h3/data/`)**

Added `load_h3.sql` (and `.xz`) creating the standard eight-table family:
- `tbl_h3index` (10 rows), `tbl_h3indexset` (10 rows)
- `tbl_th3index_{inst,discseq,seq,seqset}` (50 rows each)
- `tbl_th3index` (50 rows, mixed subtypes)
- `tbl_th3index_big` (1000 rows) - required for index tests

All rows use the `th3index(h3index, timestamptz)` constructor. The previous approach (casting `tbigint → th3index` via `WITHOUT FUNCTION`) left the internal `temptype` tag as `T_TBIGINT`, causing "Unknown type" errors and segfaults on GiST/SP-GiST index scans.

Wired as `load_h3_tables` / `DBH3` CTest fixture in `mobilitydb/test/h3/CMakeLists.txt`.

**3. Rewrite `270_th3index_tbl` and `275_th3index_indexes_tbl`**

- `270_th3index_tbl`: replaced the broken tbigint-cast approach with fixture cardinality checks and COPY BINARY round-trips for each table.
- `275_th3index_indexes_tbl`: removed the inline `CREATE TABLE tbl_th3index_big AS SELECT …` / `ANALYZE` / `DROP TABLE`; the test now starts with `SELECT COUNT(*) FROM tbl_th3index_big` (from the fixture → 1000) and keeps all GiST / SP-GiST quadtree / SP-GiST kdtree index tests.

**4. H3 MEOS smoke harness (`meos/test/h3_test.c`)**

New smoke test covering the public MEOS H3 API:
- `h3index_in` / `h3index_out` round-trip for three known cells
- `th3indexinst_make`, `th3indexseq_make` (3-instant step sequence)
- `th3index_start_value`
- `tgeo_stboxes` (dispatches to `th3index_stboxes` for `T_TH3INDEX`)
- `tgeo_split_n_stboxes`, `tgeo_split_each_n_stboxes`

Registered in a new `build-h3` CI job in `.github/workflows/meos.yml` that installs `libh3-dev` and builds with `-DH3=on -DMEOS=on`.

- [x] All 19 h3 CTest tests pass locally (`270_th3index_tbl`, `275_th3index_indexes_tbl`, `280_th3index_inspection_tbl`, `289` - `293` inline regression tests)
- [x] Consistency check in `275_th3index_indexes_tbl` returns 0 (no-idx == rtree_idx == quadtree_idx == kdtree_idx for all operators including `-|-`)
- [x] `meos.yml` `build-h3` job compiles and runs `h3_test.c` with `-DH3=on`


estebanzimanyi:
Closing as superseded by the th3index split chain.

The single commit on this branch (3a6004ce1) bundles:
- th3index spatial wiring and ever/always relationships -> superseded by #984 (th3index spatial-rels surface)
- stbox recheck fix -> incorporated into the th3index core (#976) or the static-geo walker (#977)
- fixture-based regression tests -> incorporated into #977
- Build with H3 CI job (cmake -DH3=on) -> incorporated into #976 (MEOS build wiring folded in)
- libh3-dev path pinning -> CI-only, part of the H3 CI job in #976
- meos_ecosystem and tspatial draw.io architecture diagrams -> separate from the th3index work; can be re-opened as a focused doc PR if still wanted
- doc/contributing/cross_type_parity.md and doc/contributing/type_parity_methodology.md -> these are project-methodology catalog documents and violate the new no-catalog-documents rule; if any genuinely stable convention is worth keeping, it goes in CONTRIBUTING.md as a short section
- meos/examples/trgeo_geom_clip.c -> kept on origin via the now-merged trgeo work; the example itself can be re-opened as a focused PR if useful

Closing rather than splitting because the th3index portion is fully covered by the existing split chain and the methodology docs are policy-incompatible. The remaining bits (drawio diagrams, the example) can be re-opened in a focused form if they are still valued.

PR #871, opened 2026-05-06
fix(meos): tnumber_trend accepts step-interpolated sequences

estebanzimanyi:
## Summary

- `tnumber_trend()` rejected step-interpolated sequences with an `ensure_linear_interp` error. Trend is well-defined for step interpolation (each step segment has slope zero), so the guard was incorrect.
- Removes the `ensure_linear_interp` / `ensure_linear_interp` call from the sequence-set path and replaces the top-level precondition with `VALIDATE_TNUMBER`, which accepts any interpolation mode.

## Test plan

- [ ] Existing `tnumber_mathfuncs` regression tests continue to pass
- [ ] Call `trend(tint '{1@2000-01-01, 2@2000-01-02}')` — previously ERROR, now returns a step TFloat with slope 0


estebanzimanyi:

PR #873, opened 2026-05-06
fix: batch correctness — SRID bbox refresh, aggregate value leak, timestamp string leaks

estebanzimanyi:
## Summary

- **SRID bbox refresh after transform**: tspatial_transform called lwgeom_transform but not lwgeom_refresh_bbox, leaving a stale bounding box; spatial predicates on transformed geometries returned wrong results.
- **Aggregate value leak for by-reference types**: tinstant_tagg and tsequence_tagg_iter allocated a fresh datum via func(inst1, inst2) but never freed it for by-reference base types; adds ownership check and pfree in both call sites.
- **pg_timestamptz_out string leaks in error paths**: several error-return paths in tsequence.c, tsequenceset.c, and temporal_aggfuncs.c did not free the pg_timestamptz_out string before returning; adds missing pfree calls.
- Remove #if MEOS guard from temporal_instants() which is part of the public API and needed in both builds.

## Test plan

- [ ] tspatial_srid regression tests pass; coordinate-transformed geometries have correct bounding box
- [ ] temporal_aggfuncs tests pass with no leaks under Valgrind
- [ ] temporal_instants() callable from the PostgreSQL extension

### DIFF
--- a/meos/include/rgeo/trgeo_spatialrels.h
+++ b/meos/include/rgeo/trgeo_spatialrels.h
@@ -48,11 +48,11 @@
 extern int ea_contains_geo_trgeo(const GSERIALIZED *gs, const Temporal *temp,
   bool ever);
 extern int ea_covers_geo_trgeo(const GSERIALIZED *gs, const Temporal *temp,
-  bool ever)
+  bool ever);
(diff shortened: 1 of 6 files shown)


estebanzimanyi:

PR #874, opened 2026-05-06
fix+test(meos): projection-cache dangling pointer, tpose distance leaks, meostype_name sentinel, smoke harnesses

estebanzimanyi:

## Summary

- **Projection cache use-after-free**: `meos_finalize_projsrs()` called `pfree(cache)` but did not null `MEOS_PROJ_CACHE`, causing any subsequent code path that reaches the cache to dereference a freed pointer; similarly `proj_finalize()` lacked an idempotency guard before calling `proj_context_destroy`. Both are fixed.
- **tpose distance leaks**: `tdistance_tpose_pose` was missing `pfree(tpoint)` after converting the pose temporal value to a tgeompoint; `nad_tpose_tpose` was missing `pfree(dist)` after reading the minimum.
- **`meostype_name()` null sentinel**: out-of-range or unmapped MeosType values returned NULL, which segfaulted inside `meos_error`'s `%s` formatting. Returns `"unknown"` instead.
- **New MEOS API**: adds `trgeo_points()` and `trgeo_rotation()` accessor functions for trgeometry.
- **Smoke test harnesses**: adds comprehensive MEOS smoke tests for tcbuffer, tgeometry, tnpoint, tpose, and trgeometry (with a `gen_smoketest.py` generator). These harnesses surfaced all of the above bugs.

## Test plan

- [ ] MEOS smoke tests (tcbuffer, tgeometry, tnpoint, tpose, trgeometry) pass under Valgrind with no memory errors
- [ ] meos_finalize() is idempotent when called twice
- [ ] `nad_tpose_tpose` returns correct NAD and does not leak under Valgrind


estebanzimanyi:
Folded into the consolidated MEOS quality batch PR #948. The conflict resolution kept #874's more complete projection-cache idempotency fix (NULL per slot + null global pointer) over the partial #906 version. All 166 tests pass.

PR #877, opened 2026-05-06
fix(rgeo): reject malformed MFJSON input for trgeometry

estebanzimanyi:
## Summary

- `trgeometry` input via MFJSON silently accepted malformed JSON (missing fields, wrong types) that would later cause crashes or wrong results; adds validation in `type_out.c` with proper error messages.
- Adds regression tests in `123_trgeo_inout` for the rejected malformed inputs.

## Test plan

- [ ] `123_trgeo_inout` regression tests pass
- [ ] Malformed MFJSON strings return a clear error rather than a crash or silent wrong value


estebanzimanyi:

PR #879, opened 2026-05-06
ci: add check that public headers do not use int64_t

estebanzimanyi:
## Summary

- Adds a CI script and workflow step that fails if any `meos/include/*.h` public header uses `int64_t` instead of `int64`; ensures portability across platforms where `<stdint.h>` types are not guaranteed in library consumers.
- Fixes any existing occurrences found by the new check.

## Test plan

- [ ] CI check passes on current master
- [ ] Introducing `int64_t` in a public header causes the check to fail


estebanzimanyi:
Consolidated into #910 (ci/code-quality-improvements) to reduce the review queue.

PR #880, opened 2026-05-06
ci: remove GEOS version guards no longer needed

estebanzimanyi:
## Summary

- Removes `#if POSTGIS_GEOS_VERSION >= ...` guards that were added to work around features absent in old GEOS versions; the minimum supported GEOS is now new enough that all guarded code paths are always active.
- Reduces conditional-compilation noise in 8 files; no logic change.

## Test plan

- [ ] All existing regression tests continue to pass after guards are removed


estebanzimanyi:
Consolidated into #910 (ci/code-quality-improvements) to reduce the review queue.

PR #881, opened 2026-05-06
test(geo): add KNN index coverage tests for tgeo

estebanzimanyi:
## Summary

- Adds expected-output regression tests for KNN (`<->` operator with `ORDER BY ... LIMIT`) queries on tgeometry/tgeography SP-GiST and GiST indexes.
- Covers the previously untested index-assisted nearest-neighbour path.

## Test plan

- [ ] New KNN test files pass in the coverage build with RGEO enabled


estebanzimanyi:

PR #882, opened 2026-05-06
test(tpose): add KNN/index regression tests for tpose

estebanzimanyi:
## Summary

- Adds expected-output regression tests for KNN (`<->` with `ORDER BY ... LIMIT`) and index-correctness queries on tpose indexes (GiST/SP-GiST).
- Covers the previously untested index-assisted nearest-neighbour path for tpose.

## Test plan

- [ ] New index test files pass in the coverage build


estebanzimanyi:

PR #883, opened 2026-05-06
test(trgeo): add KNN/index regression tests for trgeo

estebanzimanyi:
## Summary

- Adds expected-output regression tests for KNN (`<->` with `ORDER BY ... LIMIT`) and index-correctness queries on trgeo indexes (GiST/SP-GiST).
- Covers the previously untested index-assisted nearest-neighbour path for trgeo.

## Test plan

- [ ] New index test files pass in the coverage build


estebanzimanyi:

PR #897, opened 2026-05-06
refactor(meos): rename tpose/trgeo conversion and distance functions for precision

estebanzimanyi:
## Summary

- Renames a small set of `tpose` and `trgeo` MEOS-layer conversion and distance functions whose names were ambiguous or inconsistent with the naming convention used by other type families.
- 11 files, 32 insertions/33 deletions; purely a rename with no logic change.

## Test plan

- [ ] All existing regression tests pass unchanged
- [ ] PyMEOS / JMEOS bindings updated if they reference the old names (check meos.h diff)


estebanzimanyi:
Consolidated into #908 (refactor/meos-naming-cleanups) to reduce the review queue.

PR #902, opened 2026-05-06
ci(tnpoint): add 094_tnpoint_indexes_tbl — index correctness vs tbl_tnpoint_big

estebanzimanyi:
## Summary

Adds `094_tnpoint_indexes_tbl` regression test for tnpoint index correctness: every spatial/temporal operator verified against tbl_tnpoint_big with and without GiST/SP-GiST index (576 lines: 165 SQL + 411 expected output). Follows the `_tbl` test convention.

## Test plan

- [ ] CI green: `094_tnpoint_indexes_tbl` passes on all platforms


estebanzimanyi:

PR #915, opened 2026-05-07
Surface cbufferset in the Set-and-Span-Types chapter

estebanzimanyi:
`cbufferset` is implemented and usable on master but the manual gave users no way to discover it.

Before this PR:
- The Set-and-Span-Types overview listed every other set type (`intset`, `bigintset`, `floatset`, `textset`, `dateset`, `tstzset`, `geomset`, `geogset`, `npointset`) but not `cbufferset`
- The base-types preamble named PostgreSQL scalars, PostGIS geometry/geography, and npoint as available base types - `cbuffer` was not mentioned
- The I/O example block had no `cbufferset` literal and was missing the "elements must be double-quoted" note for it
- The SRID section showed `geomset` / `geogset` / `poseset` examples but not `cbufferset`, even though `cbufferset` carries SRID identically

Threads `cbufferset` through the three user-visible entry points where its absence was noticeable:

| File | Change |
|---|---|
| `doc/set_span_types.xml` | Add `cbufferset` to overview, base-type preamble, I/O block, and SRID section |
| `doc/es/set_span_types.xml` | Same additions in the Spanish translation |
| `doc/es/reference.xml` | Lock the Spanish type-matrix into step with the English version - was missing `pose` and `cbuffer` rows |

Everything else (accessors, set operations, comparisons, aggregations) already handles `cbufferset` implicitly via the polymorphic set-type signatures documented in the chapter - no per-type examples needed.

### DIFF
--- a/doc/es/reference.xml
+++ b/doc/es/reference.xml
@@ -98,7 +98,14 @@
 					<entry><varname>geogset</varname></entry>
 					<entry><varname></varname></entry>
 					<entry><varname></varname></entry>
-					<entry><varname>tgeogpoint</varname></entry>
+					<entry><varname>tgeogpoint</varname>, <varname>tgeography</varname></entry>
+				</row>
+				<row>
+					<entry><emphasis role="bold"><varname>pose</varname></emphasis></entry>
+					<entry><varname>poseset</varname></entry>
+					<entry><varname></varname></entry>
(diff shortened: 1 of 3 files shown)


estebanzimanyi:

estebanzimanyi:

PR #921, opened 2026-05-08
fix(geo): restore geodetic flag in tIntersects for tgeogpoint sequences

estebanzimanyi:

## Summary

Two related bugs that together break `tIntersects(tgeogpoint, geometry)` on sequence inputs:

- **`tspatial_tempspatialrels.c`**: `geom_intersection2d()` always returns a planar geometry (`GEODETIC=0`, SRID from the geometry argument). For `tgeogpoint` sequences the instants carry `GEODETIC=1` / SRID 4326, so `datum_point_eq` rejects any comparison and `tpointseq_interperiods` raises "value not found". Fix: after computing the intersection, check the sequence's `GEODETIC` flag and call `geom_to_geog()` to rebuild the result as a proper geography, mirroring the same fix already applied in the `eIntersects` path.

- **`tgeo_restrict.c`**: `geo_serialize()` always produces `GEODETIC=0` output regardless of the source `LWGEOM`'s origin. When `tpointseq_interperiods` processes a geodetic intersection, points serialised from the `lwgeom` decomposition lose the flag, causing `datum_point_eq` to reject comparisons with geodetic `tgeogpoint` instants. Fix: after each `geo_serialize` call, if the intersection is geodetic (`FLAGS_GET_GEODETIC(gsinter->gflags)`), call `geom_to_geog()` to rebuild the `GSERIALIZED` with `GEODETIC=1` and the correct 3D bounding box — applied to both the `POINT` case and the two endpoints of the `LINE` case.

Combined, `tIntersects(tgeogpoint, geometry)` now works correctly end-to-end.

## Test

```sql
-- Should return a non-null tgeogpoint restricted to the geometry
SELECT tIntersects(
  tgeogpointSeq('[POINT(0 0)@2001-01-01, POINT(1 1)@2001-01-02]'),
  'POLYGON((0.25 0.25, 0.75 0.25, 0.75 0.75, 0.25 0.75, 0.25 0.25))'::geometry
);
```

## Related

Mirrors the geodetic-flag fix already applied in the `eIntersects` path (same `geom_to_geog()` pattern).

### DIFF
--- a/meos/src/geo/tgeo_restrict.c
+++ b/meos/src/geo/tgeo_restrict.c
@@ -187,6 +187,70 @@ tpoint_force2d(const Temporal *temp)
 /*****************************************************************************/
 
 /**
+ * @brief Return the timestamp at which a segment of a geodetic temporal point
+ * is equal to a 2D point, using planar (lon/lat-as-Cartesian) arithmetic
+ * @details GEOS intersection is computed in 2D planar space; the resulting
+ * point lies on the planar straight line but deviates from the geodetic
+ * great-circle arc by up to a few microradians, which exceeds MEOS_EPSILON
+ * when using spherical distance. This function uses 2D Cartesian arithmetic
+ * matching GEOS's computation, so the residual distance is essentially zero.
+ * @param[in] inst1,inst2 Temporal values
+ * @param[in] gs Intersection point (GSERIALIZED, may be geodetic or planar)
+ * @param[out] t Timestamp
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded by the consolidated MEOS quality batch PR #948 (commit-per-feature sequence: bug-audit → naming → correctness → geodetic → lifting STEP). The two-way conflict on  between this PR and the bug-audit/correctness pair was resolved by keeping both fixes.

PR #815, opened 2026-04-27
Make MEOS thread-safe (closes #404)

estebanzimanyi:
## Summary

Closes #404.

Adds an `MEOS_TLS` qualifier (`_Thread_local` on C11+; vendor fallbacks for older GCC, Clang, MSVC) and applies it to every mutable global in MEOS-owned code AND in the vendored postgres/postgis subtrees that the parser/timezone/init paths touch. Each thread that uses MEOS now lazily allocates its own state via the existing `meos_initialize()` / `meos_finalize()` lifecycle, with no caller-side API changes — bindings (PyMEOS, JMEOS, MEOS.js, MEOS.NET) need no updates.

Three commits, kept separate so each can be reviewed independently:

1. **MEOS-owned process state TLS** (`b0c040256`) — errno, RNGs, PROJ context, SRS cache, ways cache, session timezone.
2. **lwgeom WKT parser + GMT bootstrap TLS** (`0c7015bc4`) — flex/bison-generated parser globals, plus the lazy `gmtptr` in `localtime.c`.
3. **timezone autodetect statics TLS** (`e2c828c2e`) — `findtimezone.c` scratch buffers that the auto-detect path returns pointers into.

## What's now thread-safe

| State | File | Was | Now |
|---|---|---|---|
| `MEOS_ERR_NO` | `meos/src/temporal/error.c` | process-global | per-thread (errno-style) |
| `MEOS_GENERATION_RNG`, `MEOS_AGGREGATION_RNG`, `MEOS_GSL_INITIALIZED` | `meos/src/temporal/meos.c` | process-global | per-thread (`gsl_rng` is not internally synchronised) |
| `MEOS_PJ_CONTEXT` | `meos/src/temporal/meos.c` | process-global | per-thread (matches PROJ's documented guidance) |
| `MEOS_PROJ_CACHE` | `meos/src/geo/tspatial_transform_meos.c` | process-global | per-thread (entries own `LWPROJ*` tied to the per-thread PJ_CONTEXT) |
| `MEOS_WAYS_CACHE` | `meos/src/npoint/ways_meos.c` | process-global | per-thread |
| `session_timezone`, `timezone_cache` | `postgres/timezone/pgtz.c` + `postgres/pgtime.h` (vendored) | process-global | per-thread |
| `gmtptr` (GMT bootstrap) | `postgres/timezone/localtime.c:1363` (vendored) | process-global lazy `static` | per-thread (`MEOS_TLS`) |
[...]

Issue #15, opened 2020-05-13
Error on " Create extension mobilitydb cascade"


sherazahsan:
Hi
I have installed Postgres 10 and Docker. I also successfully installed MobilityDB and got got access to the mobilitydb. 
My issue is when i try to create extension for the mobilitydb in the database., i come across the error given in below screenshot. Can anyone pl guide

![image](https://user-images.githubusercontent.com/45170605/81871933-38448580-9592-11ea-9c07-e0e6d67835de.png)

alesuiss:
Hi,

Thanks for the report; there seems to be an issue with the upstream Docker image that we use. I'll see if I can work around.

In the meantime, a temporary fix is to force loading PostGIS before loading or running MobilityDB:

```
SELECT ST_Point(0, 0) ;
CREATE EXTENSION MobilityDB ;
```

alesuiss:
Fixed in the latest Docker builds

sherazahsan:
Thank you so much ... It done now

![image](https://user-images.githubusercontent.com/45170605/81981159-57eab500-9649-11ea-9668-b83122e51239.png)

PR #939, opened 2026-05-11
fix(meos): export missing public symbols + restore JVM-safe error handler

estebanzimanyi:
## Summary

Two fixes that together let JNR-FFI / cffi consumers (JMEOS, PyMEOS,
MEOS-rs) bind the full public MEOS API without resorting to private-
header symbol-stub workarounds. Both are scoped to MEOS shared-library
exports — no behavioural change to any function's semantics.

### 1. Restore `meos_initialize_noexit_error_handler`

The default MEOS error handler calls `exit()` on every error, which
tears down the entire JVM/Python process — fatal for embedded contexts
where MEOS is loaded as a shared library. The noexit handler installs in
its place a function that records the errno (so callers can inspect via
`meos_errno()`) and writes to stderr instead of terminating the process.

This was previously added on a fork branch (commit `9ee6cf721`) but
never landed in mainline; consumers (notably MobilitySpark via JMEOS)
depend on it for crash safety. Without it, ~62 MobilitySpark tests
crash the JVM via the trigger-MEOS-error code paths.

Implementation reuses the existing thread-safe `MEOS_ERROR_HANDLER`
atomic from PR #815 — the new exporter is a one-liner
`__atomic_store_n`.

### 2. Drop unguarded `inline` from public C-SQL functions

446 functions across 44 files were declared `inline TYPE name(...)`
at top-level (not `static inline`). Per C99 semantics, an unguarded
`inline` definition has external linkage but the compiler MAY choose
to inline-only without emitting an external definition, leaving the
symbol absent from libmeos.so.

Real-world breakage: `acovers_tgeo_tgeo()` is a documented public
C-SQL entry point (`@csqlfn #Acovers_tgeo_tgeo()`) but was missing
from libmeos.so because of the `inline` qualifier. JMEOS could not
bind it; MobilitySpark's `aCoversTgeoTgeo` UDF therefore failed at
runtime.

Mechanical fix: `inline TYPE` at start-of-line becomes `TYPE`.
`static inline` (correct usage for translation-unit-private helpers)
is preserved unchanged.

Symbol count in libmeos.so increases from ~3070 to 3237 T-defined
symbols. All previously-callable functions remain callable — removing
`inline` never breaks correctness, only defeats an optimization hint
that was already unreliable for cross-TU use.

## Cross-platform impact

- **JMEOS PR #15** (regen against MEOS 1.4): drops the manual
  `extern int acovers_tgeo_tgeo(...)` workaround from its amalgamated
  meos.h once this lands; the symbol is then picked up automatically.
  Drops the `meos_initialize_noexit_error_handler` reflective fallback
  in MobilitySpark's `MeosThread`.
- **PyMEOS** / **MEOS-rs**: same regen pipeline benefits — any binding
  generator that walks the public meos.h surface gets the additional
  167 newly-exported symbols + the noexit safety helper.

## Test plan

- [x] `make meos` produces libmeos.so with `meos_initialize_noexit_error_handler`
      (`nm -D /usr/local/lib/libmeos.so | grep noexit`)
- [x] `make meos` produces libmeos.so with `acovers_tgeo_tgeo`
      (`nm -D /usr/local/lib/libmeos.so | grep acovers_tgeo_tgeo`)
- [x] MobilitySpark recovers from 845 / 907 (with reflective fallback,
      no noexit installed) to 906 / 907 (with this PR + JMEOS PR #15
      bumping the jar). The remaining 1 test failure
      (`MathUDFsExtTest.tnumberTrend_tint_returns_nonnull`) is a
      MobilitySpark fixture bug (passing tint/STEP to a function that
      requires linear interpolation) — unrelated; tracked separately.
- [ ] `mvn test` against PyMEOS — sanity check that no PyMEOS-bound
      symbol relied on a name that the inline-cleanup happens to rename
      (none expected; rename is mechanical, no symbol names change).

### DIFF
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -380,6 +380,7 @@ typedef void (*error_handler_fn)(int, int, const char *);
 
 extern void meos_initialize_timezone(const char *name);
 extern void meos_initialize_error_handler(error_handler_fn err_handler);
+extern void meos_initialize_noexit_error_handler(void);
 extern void meos_finalize_timezone(void);
 extern void meos_finalize_projsrs(void);
 extern void meos_finalize_ways(void);
--- a/meos/src/cbuffer/cbuffer.c
+++ b/meos/src/cbuffer/cbuffer.c
@@ -1367,7 +1367,7 @@ cbuffer_eq(const Cbuffer *cb1, const Cbuffer *cb2)
  * @param[in] cb1,cb2 Circular buffers
  * @csqlfn #Cbuffer_ne()
  */
-inline bool
+bool
 cbuffer_ne(const Cbuffer *cb1, const Cbuffer *cb2)
 {
   return (! cbuffer_eq(cb1, cb2));
@@ -1397,7 +1397,7 @@ cbuffer_same(const Cbuffer *cb1, const Cbuffer *cb2)
  * @brief Return true if two circular buffers are approximately equal with
  * respect to an epsilon value
  */
-inline bool
+bool
 cbuffer_nsame(const Cbuffer *cb1, const Cbuffer *cb2)
 {
   return ! cbuffer_same(cb1, cb2);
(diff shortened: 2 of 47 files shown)


estebanzimanyi:
Superseded by #949 (consolidated MEOS thread-safety: GEOS reentrant API + JVM-safe error handler).  The single noexit-handler commit on this branch is included in #949 as commit 742c1fb5.

PR #934, opened 2026-05-10
Bootstrap MEOS locale, UTF-8, and Windows builds

estebanzimanyi:
Rolling-topic MEOS bootstrap PR per the ecosystem PR consolidation policy. Three orthogonal concerns about MEOS's I/O / build behaviour, each one feature commit:

- **`fix(meos): bootstrap MEOS locale safety`** (was [PR #813](https://github.com/MobilityDB/MobilityDB/pull/813))
  Pins all numeric I/O to the C locale so WKT, set, span, and temporal-constant parsing produce stable results regardless of the calling process's `LC_NUMERIC`.  Adds a portable `strtod_l` wrapper (POSIX 2008 thread-safe path with setlocale fallback) and the public locale contract in `meos.h`.  Closes #425.

- **`fix(meos): bootstrap MEOS UTF-8 contract`** (was [PR #814](https://github.com/MobilityDB/MobilityDB/pull/814))
  Documents and enforces MEOS's encoding-agnostic byte-pass-through I/O.  Adds `meos_utf8_smoke` CI workflow that round-trips UTF-8 through `text_in`/`text_out`, set/array constructors, and `ttext`/`textset` parsers.  ASCII-only case folding is documented as intentional.  Closes #403.

- **`feat(meos): bootstrap native Windows MEOS build`** (was [PR #812](https://github.com/MobilityDB/MobilityDB/pull/812), 3-commit iterative fix squashed to 1)
  Adds `.github/workflows/windows_msys2_meos.yml` building the standalone MEOS library on Windows via MSYS2 UCRT64 + CMake/Ninja.  Includes the SYSTEMTZDIR fix (MSYS2 tzdata location) and the `SIZEOF_LONG_LONG` `pg_config.h.in` alias for GCC 16/LLP64.  Non-blocking (`continue-on-error: true`) - surfaces Windows issues without gating master.  Closes #513.

The three concerns are orthogonal - locale is about numeric formatting, UTF-8 is about byte-pass-through text, Windows is about build/CI surfaces.  No file overlap except `meos/include/meos.h` (locale + UTF-8 contract blocks coexist as separate sections).

## Closes

- Closes #813
- Closes #814
- Closes #812

- [ ] `meos_locale` workflow passes - strtod produces "1.5" → 1.5 under de_DE.UTF-8
- [ ] `meos_utf8_smoke` workflow passes - UTF-8 round-trips byte-identical
[...]

PR #942, opened 2026-05-11
fix(meos/windows): bypass setlocale → pgwin32_setlocale macro for standalone MEOS

estebanzimanyi:
## Why

PR #934 introduces a native Windows MEOS build via the new `.github/workflows/windows_msys2_meos.yml`. Its first (and currently only) push left the workflow red because the MSYS2 linker reports three undefined references to `pgwin32_setlocale`:

```
undefined reference to `pgwin32_setlocale'
  meos/src/temporal/postgres_types.c:(.text+0x295)
  meos/src/temporal/postgres_types.c:(.text+0x2f5)
  meos/src/temporal/postgres_types.c:(.text+0x318)
```

## Root cause

`postgres_types.c` includes `<postgres.h>`. On Windows that pulls in `port/win32_port.h`, which contains

```c
#define setlocale(a,b) pgwin32_setlocale(a,b)
```

`pgwin32_setlocale` is a PostgreSQL-internal helper that lives in `libpgport`. The standalone MEOS shared library (`libmeos.dll`) does not link `libpgport` — there is no PostgreSQL server backend to host it. So `meos_strtod()`'s three `setlocale(LC_NUMERIC, ...)` call sites expand to a symbol that nothing provides, and the DLL fails to link.

The in-extension build (MobilityDB-as-PostgreSQL-extension) is unaffected: the server backend already provides `libpgport` so the macro resolves cleanly.

## Fix

Undefine the macro for the MEOS-standalone Windows build only, immediately after `<postgres.h>` has been processed. `<locale.h>` is pulled in transitively through `<postgres.h>` and still provides the libc `setlocale` prototype, so the existing call sites compile and link against libc as on Linux / macOS.

```c
#if defined(_WIN32) && defined(MEOS)
  #ifdef setlocale
    #undef setlocale
  #endif
#endif
```

## Cross-platform impact

- **Linux / macOS standalone MEOS**: no change (the macro is Windows-only).
- **In-extension MobilityDB**: no change (the `defined(MEOS)` guard is off in that build).
- **Windows MSYS2 standalone MEOS** (PR #934's new CI workflow): the link error goes away; `meos_strtod()` calls libc `setlocale` directly, which is the correct behaviour for a non-server context. No data-path change to LC_NUMERIC semantics — the save-switch-to-C-restore dance is identical, just with libc as the implementation under it.

## Test plan

- [x] Linux build still succeeds (`cmake -DMEOS=ON .. && make` on Ubuntu 24.04)
- [ ] Windows MSYS2 build (PR #934's `windows_msys2_meos.yml` job re-runs green once this lands or is rebased into the PR #934 branch)
- [x] No source code path is changed on Linux / macOS / in-extension builds; the `#undef` is gated by `_WIN32 && MEOS`

## Landing strategy

This fix is a one-line guard that can land either as a standalone PR or by rebasing it into PR #934's `meos/bootstrap-batch` branch. Standalone is faster to review; rebase into #934 keeps the "MEOS Windows bootstrap" story in a single PR.

### DIFF
--- a/.github/workflows/meos_locale.yml
+++ b/.github/workflows/meos_locale.yml
@@ -1,0 +1,111 @@
+name: MEOS locale safety
+
+# Bootstrap CI for issue #425 (Localization support for the MEOS library).
+#
+# This job:
+#   1. Builds the standalone MEOS C library on Linux.
+#   2. Installs a locale that uses ',' as the decimal separator
+#      (de_DE.UTF-8 — same shape as fr_FR, it_IT, es_ES, etc.).
+#   3. Sets LC_NUMERIC to that locale and runs a tiny C test that calls
+#      meos_initialize() and parses '[1.5@2001-01-01]' as a tfloat.
+#
+# With the meos_strtod() wrapper in place (postgres_types.c) the parse
+# succeeds and the round-tripped value contains "1.5". Without it, the
+# libc strtod() respects LC_NUMERIC and would parse '.' as a thousands
+# separator, producing 15 — the test would then fail. This pins MEOS's
+# locale-blindness contract: numeric I/O is always C-locale, regardless
+# of the calling process's LC_NUMERIC.
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/meos_locale.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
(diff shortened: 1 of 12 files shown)


estebanzimanyi:
Identical content to PR #934 (every file diff matches). The bootstrap batch already covers the locale + UTF-8 + Windows fix surface, including the postgres_types setlocale workaround. Closed as a duplicate.

PR #949, opened 2026-05-12
Make MEOS thread-safe for JVM/multi-threaded hosts (GEOS reentrant + JVM-safe error handler)

estebanzimanyi:
Routes every GEOS call through a thread-local GEOSContextHandle_t and uses the reentrant GEOSXxx_r API throughout. MEOS owns its handle in meos/src/temporal/meos.c (MEOS_GEOS_CONTEXT plus geos_initialize/geos_finalize/geos_get_context, wired into meos_initialize/meos_finalize alongside GSL and PROJ); vendored liblwgeom owns its handle in postgis/liblwgeom/lwgeom_geos.c (LWGEOM_GEOS_CTX plus lwgeom_geos_context/lwgeom_geos_finalize). MEOS spatial helpers in postgis_funcs.c, tgeo_spatialfuncs.c, and tpoint_spatialfuncs.c retrieve the context per thread; the meos_call_geos2 callback signature takes GEOSContextHandle_t as its first parameter. Adds meos_initialize_noexit_error_handler(), which installs an error path that logs to stderr and sets meos_errno without calling exit(), so a MEOS error inside a JVM does not tear down the host process. Validation: the new meos/test/threaded_geos_test runs 16 threads x 20000 iterations x 5 repeats (~19M GEOS calls) with zero errors and stable RSS; pg_regress full suite 136/136 pass; MobilitySpark BerlinMOD local[2] clean across Q1 to Q11, and local[4] Q2 measured at 44s (2.05x speedup vs local[2]). Supersedes #939.

### DIFF
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -380,6 +380,7 @@ typedef void (*error_handler_fn)(int, int, const char *);
 
 extern void meos_initialize_timezone(const char *name);
 extern void meos_initialize_error_handler(error_handler_fn err_handler);
+extern void meos_initialize_noexit_error_handler(void);
 extern void meos_finalize_timezone(void);
 extern void meos_finalize_projsrs(void);
 extern void meos_finalize_ways(void);
--- a/meos/include/meos_internal_geo.h
+++ b/meos/include/meos_internal_geo.h
@@ -40,6 +40,8 @@
 #include <json-c/json.h>
 /* PROJ */
 #include <proj.h>
+/* GEOS */
+#include <geos_c.h>
 /* PostgreSQL */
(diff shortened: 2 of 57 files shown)


estebanzimanyi:
The GEOS thread-safety core is re-derived as the single-concern #1166 (per-thread reentrant GEOS context, vendored/first-party commits split, ThreadSanitizer-verified, plus the geos_clip example and lwgeom_geos_errmsg race that this older base missed). The JVM-safe no-exit error handler + the inline-drop on the 446 C-SQL functions bundled here are unrelated to GEOS thread-safety and should land as their own PR. Branch retained.

PR #932, opened 2026-05-10
Add tnpoint extent() aggregate and asMFJSON I/O

estebanzimanyi:
Consolidates two single-commit `tnpoint` feature-addition PRs into one cleanly separated 2-commit PR per the ecosystem PR + commit consolidation policy (see reviewer guide).  Each commit is one feature; zero file overlap between the two.

- **`feat(tnpoint): extent() aggregate`** (was [PR #927](https://github.com/MobilityDB/MobilityDB/pull/927))
  Adds the `extent()` aggregate for `tnpoint` - combines instants/sequences into the spatiotemporal bounding box.  Mirrors the existing `extent()` definitions on `tgeompoint` and other temporal types.  Files: `mobilitydb/sql/npoint/095_tnpoint_aggfuncs.in.sql` + matching test `095_*` query/expected pair.

- **`feat(tnpoint): asMFJSON output + tnpointFromMFJSON input`** (was [PR #928](https://github.com/MobilityDB/MobilityDB/pull/928))
  Adds Movement Features JSON I/O for `tnpoint`.  Files: `meos/src/temporal/type_in.c` + `type_out.c` (parser/serialiser entry points), `mobilitydb/sql/npoint/083_tnpoint.in.sql` (SQL surface), and a new `085_tnpoint_mfjson.test.sql` regression with matching expected output.

The two changes were originally split because they touch disjoint file sets and each is small enough for one-shot review.  Folding them is purely a queue-optimisation: same domain (`tnpoint`), same author, both green CI, no shared dependency.  Reviewer can verify each commit independently.

## Closes

- Closes #927
- Closes #928

- [ ] `095_tnpoint_aggfuncs.test` passes locally
- [ ] `085_tnpoint_mfjson.test` passes locally
- [ ] Round-trip: `asMFJSON` then `tnpointFromMFJSON` recovers the original `tnpoint` value
- [ ] `extent(tnpoint)` produces an STBox covering all input instants

PR #952, opened 2026-05-12
Add 082_tnpoint_boundary regression test

estebanzimanyi:
Adds a regression test for boundary(tnpoint) across the four temporal subtypes. Split out of #932 per the one-change-per-PR policy.


estebanzimanyi:
Closing as iteration on #951. The 082_tnpoint_boundary regression test references tnpointFromMFJSON, which only exists in the MFJSON I/O PR. The test cannot pass CI without the function it exercises; folded into #951.

PR #933, opened 2026-05-10
Codacy excludes, macOS MEOS CI, and int64 ban

estebanzimanyi:
Rolling-topic CI cleanup PR per the ecosystem PR consolidation policy. Three genuinely independent CI tweaks with zero file overlap, each a single commit:

- **`ci(codacy): exclude vendored and optional paths from analysis`** (was [PR #918](https://github.com/MobilityDB/MobilityDB/pull/918))
  Adds `.codacy.yml` with exclude paths for vendored dependencies (PostGIS, h3-pg, etc.) and optional features.  Removes spurious Codacy noise on imported third-party code.

- **`ci(meos): add macOS standalone MEOS build CI`** (was [PR #930](https://github.com/MobilityDB/MobilityDB/pull/930))
  Adds `.github/workflows/meos_macos.yml` - builds the standalone MEOS library on macOS, surfaces ARM64 / Darwin-specific issues that the Linux-only CI misses.

- **`ci: ban int64_t in public headers + remove stale GEOS version guards`** (was [PR #910](https://github.com/MobilityDB/MobilityDB/pull/910))
  Adds `tools/scripts/check_int64_in_headers.sh` and wires it into `.github/workflows/check-code.yml` to ban `int64_t` (use PostgreSQL's `int64`) in public headers.  Removes stale `#if POSTGIS_VERSION_NUMBER >= …` guards now that the minimum PostGIS version covers all branches.  The 8 test-fixture updates reflect the rendering changes.

All three are CI-only, all green, all single-author.  Folding them is purely a queue-optimisation; reviewer can verify each commit independently.

## Closes

- Closes #918
- Closes #930
- Closes #910

- [ ] Codacy: vendored paths no longer flagged
- [ ] macOS MEOS workflow runs and reports build status
- [ ] `check_int64_in_headers.sh` exits 0 on master; introduces non-zero exit if `int64_t` is added to a public header
- [ ] All test fixtures unchanged in semantics - only reflect the GEOS-guard cleanup

PR #955, opened 2026-05-12
Add macOS standalone MEOS build CI workflow

estebanzimanyi:
Adds .github/workflows/meos_macos.yml to surface macOS-specific build issues on the standalone MEOS library that the Linux-only matrix misses. Split out of #933 per the one-change-per-PR policy.

### DIFF
--- a/.github/workflows/meos_macos.yml
+++ b/.github/workflows/meos_macos.yml
@@ -1,0 +1,82 @@
+name: Build standalone MEOS on macOS
+
+# Builds libmeos.dylib on macOS (Apple Silicon) without PostgreSQL.
+# This is the native library consumed by JMEOS / MobilitySpark on macOS.
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/meos_macos.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
+      - 'CMakeLists.txt'
+    branches-ignore:
+      - gh-pages
+  pull_request:
+    paths:
+      - '.github/workflows/meos_macos.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
+      - 'CMakeLists.txt'
+    branches-ignore:
+      - gh-pages
+
+jobs:
+  build-meos-macos:
+    name: meos-macos (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-14, macos-15]
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/meos_macos.yml
```diff
@@ -0,0 +1,80 @@
...
+    branch_ignore: gh-pages
+  pull_request:
+    paths:
+      - '.github/workflows/meos_macos.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'postgis/**'
+      - 'postgres/**'
+      - 'CMakeLists.txt'
+    branch_ignore: gh-pages
```
mschoema: The syntax is `branches-ignore`, not `branch_ignore`.

mschoema: Looking at the existing actions, we seem to have the issue there too. I suggest to also open another PR just fixing this in the other workflows.

.github/workflows/meos_macos.yml
```diff
@@ -0,0 +1,80 @@
+name: Build standalone MEOS on macOS
+
+# Builds libmeos.dylib on macOS (Apple Silicon and Intel) without PostgreSQL.
+# This is the native library consumed by JMEOS / MobilitySpark on macOS.
```
mschoema: The comment mentions Intel, but `macos-14` and `macos-15` are both apple silicon.

.github/workflows/meos_macos.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Configure (MEOS=ON, no PostgreSQL extension)
+        run: |
+          BREW_PREFIX="$(brew --prefix)"
+          cmake -S . -B build-meos \
+            -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX="$PWD/meos-install" \
+            -DCMAKE_PREFIX_PATH="$BREW_PREFIX" \
+            -DMEOS=ON \
+            -DCBUFFER=ON -DNPOINT=ON -DPOSE=OFF -DRGEO=OFF
```
mschoema: Why do we have `-DCBUFFER=ON -DNPOINT=ON` but `-DPOSE=OFF -DRGEO=OFF`? Are there bugs with pose and rgeo on macos?

mschoema (dismissed):

estebanzimanyi:
Addressed the four review points in 1e3a07db1:
- Corrected branches-ignore key in both push and pull_request triggers.
- Dropped the Intel mention in the header (macos-14 and macos-15 are both Apple Silicon).
- Enabled POSE and RGEO so the macOS configure exercises the same option surface as Linux.

Follow-up PR #1010 sweeps the same branch_ignore typo across the seven other workflow files.

PR #959, opened 2026-05-12
Bootstrap native Windows MEOS build

### LINKED ISSUE #513: Windows support for the MEOS library

estebanzimanyi:
Adds a non-blocking Windows CI workflow (continue-on-error, informational until cross-platform parity catches up) that builds the standalone MEOS library on Windows via MSYS2 UCRT64 with CMake and Ninja, together with the Windows-specific build and source fixes it depends on:

- define `int64` as `int64_t` so it stays 64-bit under LLP64, avoiding the TimestampTz truncation and pointer corruption a 32-bit `long` causes on Windows;
- re-list GEOS and PROJ after postgis so single-pass GNU ld resolves the reentrant GEOS, PROJ and geodesic symbols that libpostgis references;
- supply the `WIN32` macro to the C++ translation units so the Clipper2 and C sources agree on the platform typedefs;
- resolve the native timezone data directory on Windows (`SYSTEMTZDIR` / `MEOS_TZDATA_DIR`) rather than the unavailable `my_exec_path` share path;
- restore the pgwin32 `win32error` / `_dosmaperr` helper and the MEOS libc guards in the pgtypes port layer;
- bypass the setlocale-to-`pgwin32_setlocale` macro in `meos_strtod`, use `pg_strncasecmp` in the rgeo parser, and initialize the value-at-timestamp result Datums.

Closes #513.

### DIFF
--- a/.github/workflows/windows_msys2_meos.yml
+++ b/.github/workflows/windows_msys2_meos.yml
@@ -1,0 +1,110 @@
+name: Build MEOS on Windows (MSYS2/UCRT64)
+
+# Builds the standalone MEOS C library natively on Windows under MSYS2's
+# UCRT64 runtime, providing native Windows MEOS build coverage.
+#
+# What this job verifies:
+#   * cmake -DMEOS=ON resolves all dependencies under MSYS2 mingw-w64
+#     (json-c, geos, proj, gsl, tzdata).
+#   * The standalone libmeos.dll links cleanly.
+#   * A minimal smoke test (meos/test/temporal_test) builds and runs.
+#
+# Note: full MSVC support requires sweeping PGDLLEXPORT onto every public
+# extern declaration in meos.h / meos_*.h.  That is intentionally left as
+# a follow-up; MinGW-GCC under MSYS2 exports symbols by default
+# (--export-all-symbols) so this CI runs without that sweep.
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/windows_msys2_meos.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'pgtypes/**'
+      - 'postgis/**'
+      - 'CMakeLists.txt'
+    branch_ignore: gh-pages
+  pull_request:
+    paths:
+      - '.github/workflows/windows_msys2_meos.yml'
+      - 'cmake/**'
+      - 'meos/**'
+      - 'pgtypes/**'
+      - 'postgis/**'
+      - 'CMakeLists.txt'
+    branch_ignore: gh-pages
+
+jobs:
+  build-meos:
+    name: meos-windows
+    runs-on: windows-latest
+    continue-on-error: true
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup MSYS2 (UCRT64)
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: UCRT64
+          update: true
+          install: >-
+            mingw-w64-ucrt-x86_64-gcc
+            mingw-w64-ucrt-x86_64-cmake
+            mingw-w64-ucrt-x86_64-ninja
+            mingw-w64-ucrt-x86_64-json-c
+            mingw-w64-ucrt-x86_64-geos
+            mingw-w64-ucrt-x86_64-proj
+            mingw-w64-ucrt-x86_64-gsl
+            mingw-w64-ucrt-x86_64-tzdata
+
+      - name: Resolve native timezone data path
+        run: |
+          # cygpath -m converts the MSYS2 POSIX path to a Windows-native path
+          # (forward slashes, e.g. C:/msys64/ucrt64/share/zoneinfo) that fopen()
+          # inside the compiled DLL can open at runtime.
+          TZDATA_WIN=$(cygpath -m "$MSYSTEM_PREFIX/share/zoneinfo")
+          echo "MEOS_TZDATA_WIN=$TZDATA_WIN" >> "$GITHUB_ENV"
+
+      - name: Configure (MEOS=ON, no PostgreSQL extension)
+        run: |
+          # POSE pulls in RGEO via CMAKE_DEPENDENT_OPTION; RGEO stays off
+          # because rgeo's compilation on Windows depends on PostGIS
+          # build-tree generated headers that are not available there.
+          cmake -S . -B build-meos \
+            -G Ninja \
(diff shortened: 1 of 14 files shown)


estebanzimanyi:
Findings from a remote debug pass on the Windows smoke segfault. The crash is real and reproduces deterministically, but its root cause requires Windows-side debugging that cannot be done from a Linux session.

What we know:

The temporal_test.exe smoke walks tbool, tfloat, tint, tnumber, temporal_*, ttext accessors. It crashes immediately after the first call to ttext_start_value(), regardless of where in the ttext block that call lives.

Bisect: moving ttext_start_value(ttext1) to be the FIRST call in the ttext block (before ttext_end_value, ttext_max_value, ttext_min_value) shifts the crash with it. The ttext_start_value printf prints "A" cleanly, then the process dies at the next palloc-using call (ttext_end_value or its internals).

Pointer instrumentation: a side-by-side capture of ttext_min_value(ttext1) and ttext_start_value(ttext1) shows both return valid heap addresses with valid 5-byte varlena headers (first 4 bytes = 0x14, the postgres vl_len_ encoding for a 5-byte palloc). So both look like proper fresh allocations.

The crash happens after the ttext_start_value() printf, before any free() runs, and even when the returned pointer is leaked (not freed). So the destructive side effect is in the call itself, not in free().

Refactoring temporal_start_value to mirror temporal_min_value (tinstant_value_p view + explicit datum_copy at the end with MEOS_FLAGS_GET_BYVAL gate) was tested; it does not change the symptom. So the bug is not in the tinstant_value vs (tinstant_value_p + datum_copy) split.

Hypotheses to investigate on a Windows box:

(1) UCRT64 malloc heap state corruption from the specific palloc+free sequence preceding the ttext block (~hundreds of small allocations / frees across tbool/tfloat/tint/tnumber/temporal_*). HeapValidate before and after each ttext_*_value call would localise it.

(2) Some TLS / cache state set by tinstant_value_p on the first instant of the first sequence specifically. The macros TSEQUENCESET_SEQ_N(ttext1, 0) and TSEQUENCE_INST_N(seq, 0) on the FIRST instant are reached only by start_value; min_value also reaches them in its iteration but does so AFTER a longer accumulator setup.

(3) gflags +htc / drmemory / cdb on temporal_test.exe should pinpoint the line by surfacing the heap corruption at the moment it happens rather than at the next allocation.

What is currently in master of split/meos-windows-bootstrap: the original commit only (the bootstrap workflow + Windows source fixes). The experimental wip commits I pushed for instrumentation have been force-rolled back. The Windows smoke step is still continue-on-error in the workflow, so this does not block the PR.

estebanzimanyi:
Update: root cause found and fixed.

Set up MSYS2 UCRT64 + Dr. Memory + gdb on the Windows host (WSL2 underneath), reproduced the segfault deterministically, attached gdb, broke on ensure_not_null, and got a clean backtrace:

```
#0  meos_error errlevel=ERROR Null pointer not allowed
#1  ensure_not_null ptr=0x0 at temporal.c:140
#3  text2cstring txt=0x0 at postgres_types.c:2292
#4  text_out at type_out.c:105
#5  main at meos/test/temporal_test.c:1261
```

Line 1261 is the text_out(text_result) call immediately AFTER ttext_value_at_timestamptz, not the text_out after ttext_start_value. The previous remote-debug guess (that the crash was inside ttext_start_value) was wrong.

The bug is in tX_value_at_timestamptz (and its three siblings) in meos/src/temporal/temporal_restrict_meos.c:

```c
bool ttext_value_at_timestamptz(..., text **value) {
  ...
  Datum res;                                 // uninitialized
  bool result = temporal_value_at_timestamptz(..., &res);
  *value = DatumGetTextP(res);               // assigns even when result==false
  return result;
}
```

The underlying temporal_value_at_timestamptz only writes to its result out-param on the success path; when the timestamp falls outside the temporal value the local Datum res is never written. The four wrapper functions then unconditionally read it and propagate the uninitialized bytes into the callers via *value. On Linux that uninitialized Datum happens to contain a stale stack pointer that downstream code can dereference without crashing; on Windows MSYS2 UCRT64 it is zero, which DatumGetTextP turns into a NULL text*, and the next text_out call inside the smoke test hits VALIDATE_NOT_NULL → exit.

Fix pushed as 3cf049e75: initialize res to zero in all four variants. Datum res is then a deterministic NULL/0 on the miss path across every platform, matching the existing pattern in other accessors that compute on miss.

The fix alone does not turn the smoke test green in Windows CI because the test program also has fragility further down (temporal_at_timestamptz on a timestamp outside the range, etc.), which would need separate test-side defensive coding. The Windows job stays continue-on-error: true in the workflow as before. The MEOS source bug is fixed; the smoke test work is the second pass.

estebanzimanyi:
Windows CI green!

Final root cause: meos/include/postgres_ext_defs.in.h declared int64 as long int. Under Linux LP64 and macOS LP64 that is 64 bits; under Windows MinGW UCRT64 (LLP64) it is 32 bits. Every TimestampTz, Timestamp, TimeADT, TimeOffset returned across the libmeos DLL boundary on Windows was therefore silently truncated to 32 bits. The visible smoke-test symptom was the parsed string 2001-01-01 coming back as 2000-01-01 00:47:30.766848 (the low 32 bits of the actual microsecond count, sign-extended).

Switching the typedefs to stdint.h fixed-width int8_t through int64_t (and unsigned counterparts) is a no-op on Linux LP64 and macOS LP64 and unblocks 64-bit timestamps on Windows.

Combined with 3cf049e75 (initialize Datum res to zero in the four typed tX_value_at_timestamptz wrappers so out-params are deterministic on the miss path) this is sufficient: meos/test/temporal_test under MSYS2 UCRT64 now runs to exit code 0, 696 output lines, zero Null pointer not allowed, zero segfaults.

Two source fixes for the Windows beta:
- 3cf049e75 Initialize Datum res to zero in tX_value_at_timestamptz
- 12ead5889 Use stdint.h fixed-width types in postgres_ext_defs.in.h

Both should be uncontroversial cross-platform improvements; they make Linux LP64 and macOS LP64 behavior strict-aliased to what was already implicit, and they fix the Windows truncation.

Workflow: meos-windows is now a real gate (it was never marked continue-on-error in the workflow file despite the PR description note); subsequent regressions in Windows-only behavior will block PR merges.

PR #946, opened 2026-05-11
Consolidated trgeo lifting and rgeo asMFJSON pfree fixes

estebanzimanyi:
Consolidates two rgeo/trgeo PRs into one review surface:

- #849 - lifting infrastructure crashes on trgeo↔geometry cross-type comparisons (foundational; also clears the rgeo SKIP_TESTS list)
- #923 - pfree geo_as_geojson strings in asMFJSON output functions (closes #850)

## What's in the consolidation

| # | Commit | Origin | What |
|---|---|---|---|
| 1 | `75d715d4e` | #849 | Lifting infrastructure: clean dispatch for trgeo↔geometry cross-type comparisons; reactivate 162_tcbuffer_spatialrels; restore 124_trgeo_compops_tbl |
| 2 | `658342681` | #923 | rgeo asMFJSON output functions pfree the `geo_as_geojson` payload (closes #850) |

## Effect on SKIP_TESTS (beta blocker)

- `mobilitydb/test/rgeo/CMakeLists.txt` - SKIP_TESTS reduced to empty list. All 6 previously skipped trgeo tests (122_trgeo, 124_trgeo_compops*, 150_trgeo, 152_trgeo_compops*) now pass on `-DCBUFFER=ON -DPOSE=ON -DRGEO=ON -DH3=ON` against PG 17.8.
- `mobilitydb/test/cbuffer/CMakeLists.txt` - SKIP_TESTS reduced from 10 to 3 (#849 reactivated 162_tcbuffer_spatialrels*). The remaining 3 cbuffer skips (160 + 164×2) are addressed in a stacked cbuffer PR.

```
ctest -R 'rgeo|trgeo'
100% tests passed, 0 tests failed out of 15
```

## Closes

Supersedes #849, #923. The full trgeo feature batch (#820 trgeo parity batch 2, #858 geom-clip, #859 atGeometry, #860 tdistance, #903 MEOS unit-tests, #916 tdistance + dist2d fix, #866 rgeo refactoring) is a separate consolidation - those PRs add features rather than unblock beta.

PR #962, opened 2026-05-12
Add trgeo accessors, MF-JSON output, and geometry clip

estebanzimanyi:
Add the temporal rigid geometry MEOS unit-test suite with four accessor implementations, MF-JSON output, and the trgeo_geom_clip geometry-clip kernel and SQL wrapper, and document the input/output, constructor, conversion, and accessor surface in the temporal rigid geometries chapter of the manual.

PR #965, opened 2026-05-12
Add temporal rigid geometries DocBook chapter

estebanzimanyi:
Adds doc/temporal_rigid_geometries.xml as the canonical DocBook chapter for the trgeometry type. Split out of #946.


estebanzimanyi:
DocBook content distributed into the trgeo feature PRs (#962/#967/#964 + p2), each carrying its own section per the doc-with-code model.

PR #966, opened 2026-05-12
Add SeqSetGaps coverage in 122_trgeo

estebanzimanyi:
Adds SeqSetGaps coverage to the 122_trgeo regression test. Split out of #946.


estebanzimanyi:
Closing as superseded by #1001. The identical SeqSetGaps coverage pattern was applied to four test files in four separate PRs; combining three into one reduces reviewer overhead since the pattern is identical. The tpose variant (#991) stays separate because it has additional dependencies on the GeoPose feature stack.

PR #968, opened 2026-05-12
Add tdistance for trgeo by trgeo and trgeo by tgeompoint

estebanzimanyi:
Implements tdistance for the trgeo by trgeo and trgeo by tgeompoint shapes. Stacks on the trgeo accessors PR and the pose-distance fix; Draft until they merge. Split out of #946.


estebanzimanyi:
Subsumed by #964 (trgeo feature regrouping for review). Included verbatim in the grouped PR.

PR #967, opened 2026-05-12
Add trgeo restrictions and spatial relationships

estebanzimanyi:
Add the temporal rigid geometry restrictions atGeometry, minusGeometry, atStbox and minusStbox, the traversed area, and the ever, always and temporal spatial-relationship predicates against geometry, and document the traversed area, modification, and spatial restriction surface in the temporal rigid geometries chapter of the manual.

PR #969, opened 2026-05-12
Close trgeo temporal spatial-rel parity gap with 18 functions

estebanzimanyi:
Adds the 18 temporal spatial-relationship functions trgeo needed to reach parity with the other spatial temporal types. Stacks on the restrictions PR; Draft until it merges. Split out of #946.


estebanzimanyi:
Subsumed by #967 (trgeo feature regrouping for review). Included verbatim in the grouped PR.

PR #947, opened 2026-05-11
Clear cbuffer SKIP_TESTS and fix 160 LIMIT-determinism (stacked on #946)

estebanzimanyi:
Beta-blocker companion to #946 (trgeo). Clears the remaining cbuffer SKIP_TESTS list.

| Skip removed | Why it's safe to remove |
|---|---|
| `160_tcbuffer_distance_tbl` | Was flaky because eight `LIMIT 10` subqueries had no `ORDER BY`. Added `ORDER BY k LIMIT 10` (k is each table's PRIMARY KEY) - stable 10-row prefix across runs. |
| `164_tcbuffer_tempspatialrels` | Comment claimed "missing tcontains/econtains/ecovers SQL declarations for tcbuffer". Stale: declarations live in `mobilitydb/sql/cbuffer/164_tcbuffer_tempspatialrels.in.sql`. Test passes against current implementations. |
| `164_tcbuffer_tempspatialrels_tbl` | Same as above (table form). |

```
ctest -R 'cbuffer'
100% tests passed, 0 tests failed out of 19
```

## Stacking

Stacked on #946 - the empty SKIP_TESTS state of `mobilitydb/test/cbuffer/CMakeLists.txt` in this PR builds on the 162_tcbuffer_spatialrels reactivation #946 brought in (#849's lifting fix). Land #946 first; this PR auto-rebases to a single commit on top.

## Closes

Independent - no existing PR was tracking these specific 3 cbuffer skips. The cbuffer feature PRs (#872 missing spatialrels, #847 tDwithin/tIntersects SQL overloads, #864 tcbuffer minor gaps, parts of #929) remain a separate consolidation.

PR #972, opened 2026-05-12
Remove cbuffer SKIP_TESTS and fix 160 LIMIT 10 determinism

estebanzimanyi:
Clears the SKIP_TESTS markers in mobilitydb/test/cbuffer/CMakeLists.txt and fixes the LIMIT 10 ordering nondeterminism in 160_tcbuffer_distance.test.sql. Stacks on the lifting crash fix (#960); Draft until it merges. Split out of #947.


estebanzimanyi:
Subsumed by #1029 (tcbuffer distance regrouping) — the distance table test enablement is folded into the grouped PR.

PR #970, opened 2026-05-12
Add tcbuffer spatial relationships

estebanzimanyi:
Add the ever and always spatial relationships for circular buffers: eCovers, aCovers, eTouches, and aTouches over tcbuffer pairs and eContains and eCovers between a geometry and a tcbuffer, with STRICT and result-type fixes for tDwithin and tIntersects and the cross-type temporal comparison fix for a rigid geometry against a geometry.

PR #974, opened 2026-05-12
Add missing eCovers, aCovers, eTouches, aTouches and fix trgeo by geometry comparison crash

estebanzimanyi:
Adds the missing ever and always Covers and Touches functions on the cbuffer and trgeo surface and fixes the trgeo by geometry comparison crash. Stacks on the lifting fix (#960) and the cbuffer STRICT restype fixes; Draft until both merge. Split out of #947.


estebanzimanyi:
Subsumed by #970 (tcbuffer feature regrouping for review). The change is included verbatim in the grouped PR.

PR #944, opened 2026-05-11
Add temporal H3 cell index type (consolidated th3index)

estebanzimanyi:
Consolidates four scattered th3index PRs into a single review surface, per the "minimize PRs" policy:

- #807 - core `th3index` temporal H3 cell index type
- #938 - static-geometry → H3 cell set walker + polygon-coverage soundness (gap A: vertex + 1-ring expansion)
- #943 - trip-side densification (gap B: `tgeompoint_to_th3index` walker mirroring static-side Nyquist sampling); restores `T_H3INDEX` cases in `spatial_srid` / `spatial_flags`
- #893 - operational-notes DocBook chapter (when-to-use, H3 hazards, durability/storage)

The four PRs were logically a stack but were opened as siblings, fragmenting review across overlapping diffs. This consolidation linearises them into one branch with a commit-per-feature sequence.

## What's in the consolidation

| # | Commit | Origin | What it does |
|---|---|---|---|
| 1 | `fc10696e2` | #807 | Adds `tbigint` type + `th3index` temporal H3 cell index type |
| 2 | `45d3f253d` | #807 follow-up | Reclassifies `th3index` as tspatial+tgeodetic with geodetic STBox bbox |
| 3 | `8f2c043d5` | #938 | Static-geometry → H3 cell set public API (POINT/LINESTRING/POLYGON/MULTI*/GeomCollection) |
| 4 | `814c4aab9` + `3b3fca303` + `55fad793c` | #938 | pg_regress fixtures for `geoToH3Cell` / `geoToH3IndexSet` / `everIntersectsH3IndexSet_Th3Index`; renumber to 289 (post-th3index); correct includes |
| 5 | `0b4e4aaa8` | #938 | **Gap A fix** - sound polygon coverage via vertex + 1-ring expansion (was: `polygonToCells` `CONTAINMENT_CENTER` dropped overlapping-but-not-centred cells) |
| 6 | `002726f59` | #943 | **Gap B fix** - densify `tgeompoint_to_th3index`; restore `spatial_srid` / `spatial_flags` `T_H3INDEX` dispatch (lost as side effect of the parallel branches) |
| 7 | `944317056` | #893 | Operational-notes DocBook section: when-to-use, 5 H3-specific hazards with mitigations, durability/storage conventions |

## Soundness - the prefilter is now sound on both sides

The original sibling PRs left the H3 prefilter approximate (~81 % false-negative rate on BerlinMOD ch1). The two gaps are now both closed:

[...]

PR #977, opened 2026-05-12
Add static-geometry to H3 cell set public API

estebanzimanyi:
Adds the static-geometry to H3 cell set walker (POINT, LINESTRING, POLYGON, MULTI, GeomCollection) used to materialise an h3indexset for join prefiltering. Includes the pg_regress fixtures (geoToH3Cell, geoToH3IndexSet, everIntersectsH3IndexSet_Th3Index), the post-th3index test renumbering to 289, and the header includes correction. Stacks on the th3index core; Draft until that merges. Split out of #944.

PR #978, opened 2026-05-12
Make polygon to H3 coverage sound via vertex and 1-ring expansion

estebanzimanyi:
Fixes gap A in the static-geometry to H3 walker: polygonToCells with CONTAINMENT_CENTER dropped overlapping cells, so a join prefiltered against the result could miss matches. The fix expands the cover with vertex-touching cells and a 1-ring buffer, making the prefilter sound. Stacks on the static-geo walker; Draft until that merges. Split out of #944.


estebanzimanyi:
Closing as iteration on #977. The polygon-coverage soundness fix (Gap A) belongs with the static-geometry walker that introduced the unsound coverage; folded into #977.

PR #981, opened 2026-05-12
Refresh 289_h3_geo expected for restored h3index SRID dispatch

estebanzimanyi:
Refreshes the 289_h3_geo expected output to capture the restored h3index SRID dispatch path. Stacks on the static-geo walker; Draft until that merges. Split out of #944.


estebanzimanyi:
Closing as iteration on #977. The 289_h3_geo expected refresh is a follow-up to the static-geometry walker's test; folded into #977.

PR #983, opened 2026-05-12
Make 282 sequence-form query deterministic and refresh expected for CI precision

estebanzimanyi:
Adds an ORDER BY to the 282 sequence-form query so the result is deterministic across build configurations, and refreshes the expected output to match the CI-specific timestamp precision. Stacks on the prior th3index PRs; Draft until they merge. Split out of #944.


estebanzimanyi:
Closing as iteration on #977. The 282 deterministic ORDER BY plus the CI-precision expected refresh are test-stability follow-ups; folded into #977.

PR #976, opened 2026-05-12
Add tbigint type and th3index temporal H3 cell index type

estebanzimanyi:
Adds the tbigint scalar type and the th3index temporal type, a temporal indexer of H3 cells suitable for prefiltering spatial joins on a road or movement network.

The tbigint scalar is the storage representation; th3index is the same int64 storage exposed through an H3-specific API. The two are co-developed and co-validated: the BerlinMOD benchmark integration tests both surfaces together. Per the new policy's cohesion test (this change does not make sense without that change), they ship as one PR rather than two stacked Draft PRs. The earlier standalone tbigint PR #788 has been closed as superseded.

Includes the type catalog wiring, the geodetic STBox bbox classification, the MEOS build linking (link h3 OBJECTS, gate fmgr.h under MEOS-only build). Closes #807.

PR #985, opened 2026-05-12
Link h3 OBJECTS in MEOS build and gate fmgr.h under MEOS-only

estebanzimanyi:
Adjusts the MEOS-only build to link the h3 OBJECTS target into the standalone library and gates fmgr.h inclusion behind the MEOS flag so the build compiles on the MEOS-only path. Stacks on the full th3index chain; Draft until it merges. Split out of #944.


estebanzimanyi:
Closing as iteration on #976. The MEOS build wiring (link h3 OBJECTS, gate fmgr.h under MEOS-only) ships with the type itself; the changes have been folded into #976. Per the new policy, build wiring is not a separate change from the feature that requires it.

PR #945, opened 2026-05-11
Consolidated tpose feature batch: GeoPose v1.0, parity, drift, coverage

estebanzimanyi:
Consolidates the four pose-only PRs into a single review surface, per the "minimize PRs" policy. Each PR was already exactly one commit on master; the consolidation preserves the commit-per-feature structure.

- #819 - OGC GeoPose v1.0 standard support (workstreams #1 - #7 + #1.5)
- #865 - tpose spatial wiring (parity gap close via tgeompoint composition)
- #925 - pose quaternion drift tolerance + auto-renormalise
- #857 - pose SQL regression coverage for spatialfuncs and distance

## What's in the consolidation

| # | Commit | Origin | Theme | What it does |
|---|---|---|---|---|
| 1 | `36228c6c9` | #819 | foundation | OGC GeoPose v1.0 JSON I/O + temporal accessors and transforms (yaw/pitch/roll/speed/angularSpeed/normalise/applyPose) |
| 2 | `f13f05888` | #865 | parity | tpose spatial functions, analytics, tile via tgeompoint composition |
| 3 | `6736f658d` | #925 | correctness | widen quaternion drift tolerance + auto-renormalise on input (handles real IMU/AR-VR runtime input with `|q|=1±~1e-6`) |
| 4 | `3cbab182a` | #857 | coverage | SQL regression suite for tpose spatialfuncs and distance |

The four PRs were logically a stack - foundation → parity → correctness → coverage - but opened as siblings, fragmenting review. This linearises them.

## Conflict resolutions

- `mobilitydb/sql/pose/CMakeLists.txt` - both #819 and #865 claimed slot 115. #819's `115_geopose_frames` renamed to `120_geopose_frames` (#865's 115-119 are the tpose-band extensions, geopose is a distinct feature group). The rename is embedded in #819's commit so the history reads cleanly.
- `mobilitydb/test/pose/queries/100_pose.test.sql` + matching `expected/` - both #819 and #925 inserted at the same hunk. Unioned in the natural reading order: SRID frame transforms + applyPose (#819) → quaternion drift + canonicalization (#925).

- [x] Builds cleanly with `-DCBUFFER=ON -DPOSE=ON -DRGEO=ON -DH3=ON` against PG17
- [ ] CI green across the full matrix
- [ ] BerlinMOD regression unaffected (pose isn't on the bench critical path)

## Closes

Supersedes #819, #865, #925, #857. Closure pending review of this PR.

[...]

PR #988, opened 2026-05-12
Widen pose quaternion drift tolerance and auto-renormalise on input

estebanzimanyi:
Widens the quaternion drift tolerance accepted on pose input and auto-renormalises so a value with bounded floating-point drift is accepted rather than rejected. Stacks on the GeoPose foundation; Draft until that merges. Split out of #945.


estebanzimanyi:
Closing as iteration on #986. The quaternion drift tolerance widening and auto-renormalise on input fix the same surface that the GeoPose foundation introduces; folded into #986.

PR #987, opened 2026-05-12
Close tpose parity gap with spatial functions, analytics, and tile via tgeompoint composition

estebanzimanyi:
Adds the spatial functions, analytics, and tile surface for tpose by composing through tgeompoint, closing the parity gap with the other spatial temporal types. Stacks on the GeoPose foundation; Draft until that merges. Split out of #945.

PR #989, opened 2026-05-12
Add SQL coverage for tpose spatialfuncs and distance

estebanzimanyi:
Adds the missing SQL regression coverage for tpose spatial functions and distance. Split out of #945.


estebanzimanyi:
Folded into #987 as commit `6dc2e9540` (SQL coverage now ships in the same PR as the implementation, per the coverage-hygiene policy).

PR #990, opened 2026-05-12
Refresh tpose expected outputs preserving trailing whitespace and full FP precision

estebanzimanyi:
Refreshes the tpose expected outputs to match the current behaviour while preserving trailing whitespace and full floating-point precision (no rounding). Stacks on the tpose feature PRs and the SQL coverage; Draft until they merge. Split out of #945.


estebanzimanyi:
Folded into #986 (squashed into the GeoPose feature commit, since the refreshed expected outputs are for 100_pose.test.out and 103_pose_geopose.test.out — both #986 test files, not tpose-spatial-analytics files). Iteration signal per the iteration-PR policy.

PR #993, opened 2026-05-12
Prefix the SkipListType enum values for disambiguation

estebanzimanyi:
The skiplist mode discriminator enum values are SKIPLIST_TEMPORAL and SKIPLIST_KEYVALUE rather than the over-generic TEMPORAL and KEYVALUE, disambiguating them from other identifiers.

### DIFF
--- a/meos/include/meos_internal.h
+++ b/meos/include/meos_internal.h
@@ -733,13 +733,23 @@ struct SkipList
 };
 
 /**
- * @brief Enumeration for the relative position of a given element into a
- * skiplist
+ * @brief Skiplist mode discriminator
+ *
+ * SKIPLIST_TEMPORAL is the mode used by every production aggregate
+ * (temporal_aggfuncs.c, tgeo_aggfuncs.c, tnpoint_aggfuncs.c); elements are
+ * Temporal* values and merging routes through tsequence_tagg.
+ *
+ * SKIPLIST_KEYVALUE is the (key, value)-pair mode used by MEOS-direct
+ * streaming consumers (see meos/examples/ais_expand_skiplist.c). The
+ * supported pattern is search-then-splice-on-miss with caller-managed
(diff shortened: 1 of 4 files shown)


meos/include/meos_internal.h
```diff
@@ -733,13 +733,23 @@ struct SkipList
...
+ * SKIPLIST_TEMPORAL is the legacy mode used by every production aggregate
+ * (temporal_aggfuncs.c, tgeo_aggfuncs.c, tnpoint_aggfuncs.c); elements are
+ * Temporal* values and merging routes through tsequence_tagg.
+ *
+ * SKIPLIST_KEYVALUE is the (key, value)-pair mode used by MEOS-direct
+ * streaming consumers (see meos/examples/ais_expand_skiplist.c). The
+ * supported pattern is search-then-splice-on-miss with caller-managed
+ * in-place mutation via the user-supplied merge_fn. The batch-merge code
+ * path inside keyval_skiplist_merge is unvalidated and contains known
+ * correctness bugs (1-4 in doc/drafts/keyval_skiplist_continuation_plan.md).
```
mschoema: The reference to the markdown file should be removed as it is not in the repo.

mschoema (dismissed):
This PR theoretically contains 2 changes and should be split, but it's small enough so I'll accept it this time (after the requested change to the comment). Just a note for future PRs.

estebanzimanyi:
@mschoema Thanks. Addressed in 7262525d: dropped the reference to the markdown file that is not in the repo from meos_internal.h:747; kept the surrounding sentence that documents the batch-merge code path as unvalidated with known correctness bugs.

Noted on the split-into-two policy for future PRs.

meos/include/meos_internal.h
estebanzimanyi: Removed — the comment no longer references that markdown file; it ends at the known-correctness-bugs note.

PR #995, opened 2026-05-12
Demote LINEAR to STEP when result type does not support linear interpolation

estebanzimanyi:
Centralize the output-interpolation decision for tfunc_tsequence and tfunc_tsequence_base in a single helper lift_result_interp that consults temptype_supports_linear(restype). When the result type does not support LINEAR (e.g. STEP-only types like tbool, tint, ttext, th3index), a LINEAR input is demoted to STEP; otherwise the input's interpolation is carried through unchanged. The catalog predicate is the single source of truth for the result type's interpolation contract; the helper enforces it at the lift level so the contract holds without requiring every caller to keep lfinfo->reslinear consistent with the result type.

### DIFF
--- a/meos/src/temporal/lifting.c
+++ b/meos/src/temporal/lifting.c
@@ -181,11 +181,34 @@
 /* MEOS */
 #include <meos.h>
 #include <meos_internal.h>
+#include "temporal/meos_catalog.h"
 #include "temporal/temporal_restrict.h"
 #include "temporal/tsequence.h"
 #include "temporal/tsequenceset.h"
 #include "temporal/type_util.h"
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Closing as superseded by #1003. The caller-responsibility approach (every producer sets lfinfo.reslinear; tfunc_tsequence and tfunc_tsequence_base honor it directly) is cleaner than the centralized lift_result_interp helper introduced here, and #1003 already implements that path while also adding the tpfn_unary densification slot. The temptype_supports_linear predicate stays on master for its other 19 callers (default-interp selection in parsers, aggregates, and new-instant construction); only the lift-path coupling introduced by this PR is dropped.

PR #867, opened 2026-05-06
Close tpcpoint parity gap: SRID, centroid, trajectory, spatial rels, analytics, tile

estebanzimanyi:
Closes all remaining cross-type parity gaps for `tpcpoint` (temporal pgPointCloud point). Depends on #818 (main pgPointCloud PR, must merge first). All additions use the `::tgeompoint` projection cast - zero new C code beyond the SRID dispatch fix.

**Category 3 - Spatial functions** (`441_tpcpoint_spatialfuncs.in.sql`):
- `SRID(tpcpoint)` - via generic `Tspatial_srid`; `spatial_srid()` dispatches on `T_PCPOINT` and reads the SRID from the pcpoint schema via `meos_pc_schema(pcid)`.
- `centroid(tpcpoint) → tgeompoint` - SQL alias for `$1::tgeompoint`.
- `trajectory(tpcpoint) → geometry` - delegates to `trajectory(tgeompoint)`.

**Category 9 - Temporal spatial rels** (`443_tpcpoint_tempspatialrels.in.sql`):
- `tIntersects`, `tDisjoint`, `tDwithin` for `(geometry, tpcpoint)` and `(tpcpoint, geometry)`.

**Category 12 - Analytics** (`444_tpcpoint_analytics.in.sql`):
- `minDistSimplify`, `minTimeDeltaSimplify`, `maxDistSimplify`, `douglasPeuckerSimplify`.
- Simplify XY trajectory, extract surviving timestamps, restrict original `tpcpoint` - all per-point sensor channels (intensity, classification, …) preserved at surviving instants.

**Category 13 - Tile** (`445_tpcpoint_tile.in.sql`):
- `spaceBoxes`, `timeBoxes`, `spaceTimeBoxes`, `spaceSplit`, `spaceTimeSplit`.
- Double-cast `tpcpoint → tgeompoint → tgeometry`; reconstruct `tpcpoint` via `atTime($1, getTime(r.tgeo))`.
- New composite types `point_tpcpoint` / `point_time_tpcpoint`.

Part of the cross-type parity initiative - see `doc/contributing/cross_type_parity.md` for the full matrix. With this PR, `tpcpoint` reaches 100% parity on all applicable categories. Merge after #818.

- [ ] Merges cleanly on top of #818
- [ ] `SELECT SRID(t)` returns the schema-defined SRID
- [ ] `SELECT centroid(t)` returns a `tgeompoint`
[...]

PR #999, opened 2026-05-12
Add pgPointCloud temporal types tpcpoint and tpcpatch with per-point operators

estebanzimanyi:
Adds the pgPointCloud temporal types tpcpoint and tpcpatch as the foundation for temporal point-cloud workloads in MobilityDB, together with the per-point operator surface and the CI skip annotations for the cells that are not yet covered. Split out of #867.


estebanzimanyi:
Superseded by the policy-compliant split that keeps vendored third-party code separate from first-party code: #1028 vendors pgPointCloud v1.2.5 (with the matching .codacy.yml exclusion), #1048 adds the first-party tpcpoint/tpcpatch types and per-point operators forward-ported onto current master, and #1059 carries the point cloud documentation-prose cleanup. Together #1028 + #1048 + #1059 reproduce the entire content of this bundled PR with the required vendor/first-party separation, so this PR is closed in favor of that split.

PR #1002, opened 2026-05-12
Add cross-type parity methodology under doc/methodology/

estebanzimanyi:
Adds the cross-type parity methodology and its measured status report under `doc/methodology/`.

**Cross-type parity** means a function defined for one temporal type is expected on every other type in its family, with the same name and signature shape, so one query pattern works across types. There are two family references:

- **Point family** — reference `tgeompoint`; members `tgeogpoint`, `tnpoint`.
- **Extended-shape family** — reference `tgeometry`; members `trgeometry`, `tcbuffer`, `tpose`.

A member's expected surface is its reference's surface **minus** the reason-marked *intentional exclusions* (semantic and structural correct-absences that must never be implemented to chase a number).

Two files:
- `cross_type_parity.md` — the timeless methodology: families, function categories, the intentional-exclusion catalog, and the audit procedure.
- `cross_type_parity_report.md` — the **measured** status: a per-type parity table (with the excluded-operator accounting), the exclusions grouped by type, and the residual real gaps, produced by `tools/parity_audit/`.

The methodology is timeless; the measured report is a snapshot that updates when a parity-extending function family lands. Not a PR-queue catalog: this captures feature state, not queue state.

Placed under `doc/methodology/` (domain-specific parity invariants and audit procedure) rather than `doc/contributing/`. After the planned MEOS / MobilityDB repo split, these files migrate to MEOS unchanged, alongside the ecosystem-wide RFCs, as the single source of truth all bindings reference.

Cross-repo replication to MobilityDuck, JMEOS, MobilitySpark, PyMEOS, and meos-rs is deferred per the land-in-one-repo-first policy.


mschoema (changes requested):
I'm not against adding this, but I have 3 main comments:

1. We shouldn't already add things that are not committed yet. So th3index and tpcpoint should be removed from this PR and added in their respective PRs later (or in an independent doc PR per feature, which is probably better).
2. Audit Methodology seems to not provide much useful information to readers. Personally I'd remove it, but up to you.
3. I'm not a big fan of having file names in the tables. This might constraint things, and we'll most certainly forget to update this when we change files names/add new ones/etc.

estebanzimanyi:
@mschoema Thanks for the review. Addressed all three points in 6e560100:

1. Removed th3index and tpcpoint everywhere in this doc. They'll get their own doc section in the respective PRs that land their type.
2. Dropped the Audit Methodology section.
3. Removed both per-type file mapping tables. The doc now describes families and categories without naming individual SQL files, so renames and additions don't require touching it.

estebanzimanyi:
Thanks for the review. Addressed:
1. Dropped the not-yet-merged H3 reference — the geodetic space-binning note now just says it needs a spherical cell grid, no `th3index`/`geoToH3Cell`. (No `tpcpoint` references remain.)
2. The Audit Methodology section is no longer present.
3. Removed the hard-coded adapter file path from the table — it now reads as a SQL adapter over `pg_proc`.

estebanzimanyi:
Net-new methodology documentation, not part of the shipped pin; closing to reduce reviewer load.

PR #1004, opened 2026-05-13
Align bbox adjacency on strict set-theoretic share-a-boundary semantics

estebanzimanyi:
adjacent_span_span previously required the inclusive and exclusive bits on the touching endpoints to differ before reporting adjacency, while adjacent_stbox_stbox reported adjacency whenever any axis of the box intersection was degenerate, including the case where one input was itself degenerate on that axis and its value fell inside the other input's interior (a TInstant key sitting inside a tstzspan query, for example). Neither matched a clean topological definition, and the two predicates disagreed on cases that the th3index index test in #976 exercised. This PR replaces both with strict share-a-boundary: two closed regions are adjacent iff their closures meet on a codim-1 slab that lies on the boundary of both inputs and their interiors do not overlap. For spans this collapses to upper-equals-lower on either side regardless of inc/exc; for stboxes it requires at least one axis where box1.max equals box2.min or box2.max equals box1.min while no axis is fully disjoint. tbox adjacency rides on adjacent_span_span and inherits the same shape. Expected outputs for span, tbox, temporal, tgeo, tpoint, tcbuffer, tnpoint, tpose, and trgeo topology tests are refreshed; full local suite (166 tests) passes.


estebanzimanyi:
Exact duplicate of #1026, and the share-a-boundary (drop-the-bits) span predicate both carried is the continuous-type regression (it makes floatspan [1,5] -|- [5,9] return true). Master's adjacent_span_span is already correct (inc/exc-aware); #1026 is repurposed to document and pin that canonical rule. Closing as superseded.

PR #1007, opened 2026-05-13
Add minDistance(tgeompoint[], tgeompoint[]) exact set-set spatial distance

estebanzimanyi:
Adds an exact MEOS-native fused-aggregate min spatial distance between two arrays of temporal geos, equivalent to ST_Distance over the union of trajectories on each side but with each trip's STBox used as a sound lower-bound prefilter. Trip pairs are visited in ascending bbox-distance order and the iteration short-circuits once the running minimum is provably smaller than every remaining pair's lower bound. Per-pair distance still goes through liblwgeom's geom_distance2d / geom_distance3d so the answer is bit-identical to the aggregated form. Verified bit-for-bit equivalent to ST_Distance(ST_Collect, ST_Collect) on the BerlinMOD Q5 cross-join (100/100 licence pairs match). Includes smoke tests on crossing / parallel / multi-element / equivalence / empty / mixed-SRID cases.

### DIFF
--- a/doc/es/reference.xml
+++ b/doc/es/reference.xml
@@ -1175,6 +1175,10 @@
 				</listitem>
 
 				<listitem>
+					<para><link linkend="tgeo_minDistance"><varname>minDistance</varname></link>: Devuelve la distancia espacial mínima entre dos geometrías temporales sobre la unión de sus extensiones temporales</para>
+				</listitem>
+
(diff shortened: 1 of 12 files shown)


estebanzimanyi:
Merged to unblock current work on the ecosystem bump to 1.3

PR #1008, opened 2026-05-13
Address cppcheck-derived code-scanning alerts in one sweep

estebanzimanyi:
Closes the open cppcheck-derived code-scanning alerts visible on every open PR by fixing the underlying issues on master. Promotes read-only pointer locals and parameters to pointer-to-const across MEOS source and headers, drops unread or shadowing variable assignments, removes an unused struct typedef in tspatial_transform_meos, restores the declaration/definition argument-order parity on numspan_tstzspan_to_tbox, merges two duplicate-condition branches in tpointseq_decouple_iter, simplifies a trivially-true return in tgeoseq_ev_geo_func, fixes an off-by-one in tsequence_segm_duration_iter so the segment carrying the upper-inclusive bound is the last one, renames the second local i in AddToMEOSPROJSRSCache, and extends the cppcheck workflow with explicit basename suppressions for the vendored lwin_wkt flex and bison output so their findings stop leaking through to the SARIF upload.


estebanzimanyi:
This PR addresses every open CppCheck code-scanning alert.

**First-party (98 alerts) — fixed in code, no suppressions:**
- 44 × constVariablePointer
- 21 × constParameterPointer
- 9 × unreadVariable
- 8 × shadowVariable
- 5 × knownConditionTrueFalse
- 4 × unusedStructMember
- 2 × shadowFunction
- 2 × funcArgOrderDifferent
- 2 × duplicateCondition
- 1 × constVariable

Each is resolved by a genuine source fix (const-correctness, unused-variable removal, shadow renames) — 71 const-correctness edits, zero inline `cppcheck-suppress`. Every alert file is covered, and the build is green across PostgreSQL 16/17/18 on Ubuntu and macOS.

**Vendored (30 alerts) — provenance-excluded:** the lwin_wkt lexer/parser alerts are suppressed in `cppcheck.yml` alongside the existing postgis / pgtypes / h3-pg exclusions, since these are verbatim third-party sources whose style we do not own.

estebanzimanyi:
Stale cppcheck sweep: forked far behind master. Its diff reverts current master code — e.g. tnumber_mathfuncs now uses the correct DatumGetFloat8, and this branch reintroduces Float8GetDatum — and master is already cppcheck-critical-clean. Closing as stale churn to keep the open set lean and avoid a revert when composing the pin.

PR #1012, opened 2026-05-14
Add eCovers, aCovers, and tCovers (geometry, tgeompoint) overloads

estebanzimanyi:
For a temporal point the value at every instant is a point, and ST_Covers and ST_Intersects are equivalent on a (polygon, point) pair, so the new Covers SQL overloads on tgeompoint and tgeogpoint dispatch to the same C kernels as the existing Intersects overloads. Closes the gap surfaced when calling eCovers, aCovers, or tCovers with a geometry and a tgeompoint at the SQL level; previously the call failed to resolve against the parent tgeometry overload because the tgeompoint to tgeometry cast rejects linear interpolation.

### DIFF
--- a/mobilitydb/sql/geo/070_tpoint_spatialrels.in.sql
+++ b/mobilitydb/sql/geo/070_tpoint_spatialrels.in.sql
@@ -178,6 +178,37 @@ CREATE FUNCTION aDisjoint(tgeogpoint, tgeogpoint)
   LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
 /*****************************************************************************
+ * eCovers, aCovers
+ *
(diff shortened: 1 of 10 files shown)


estebanzimanyi:
Closing per reviewer feedback. The change was technically equivalent (Covers ≡ Intersects on a point), but conceptually wrong: Covers only carries distinct semantics from Intersects when the moving value is a shape with extent. Since arbitrary geometries cannot be interpolated between snapshots, tgeometry is the only temporal type where Covers meaningfully differs from Intersects, and the tgeompoint→tgeometry cast correctly rejects linear interpolation for that reason. Users wanting Covers semantics on a tgeompoint should use eIntersects/aIntersects/tIntersects directly.

PR #1014, opened 2026-05-14
Fix spatialrel_tgeo_geo always-quantifier on collection trajectories

estebanzimanyi:
The collection-element loop in spatialrel_tgeo_geo broke on the first true element regardless of quantifier, which is correct for ever semantics but wrong for always semantics: a TSEQUENCESET whose trajectory is a geometry collection of several linestrings would be reported as always covered, always contained, or always within when only one composing linestring satisfied the predicate. Adds an ever parameter that selects break-on-first-true for ever and break-on-first-false for always, with a vacuously-true identity for the empty collection case, and threads the right value through the six call sites in ea_contains, ea_covers, ea_disjoint, ea_intersects, and ea_dwithin. Updates the expected counts in 070_tgeo_spatialrels_tbl and 070_tpoint_spatialrels_tbl, which previously captured the inflated true counts; cross-verified the new aDwithin count on tbl_geom_point against ST_Covers(ST_Buffer(g,d), trajectory(temp)).

### DIFF
--- a/meos/src/geo/tgeo_spatialrels.c
+++ b/meos/src/geo/tgeo_spatialrels.c
@@ -295,21 +295,24 @@ geo_dwithin_fn_geo(int16 flags1, uint8_t flags2)
  * @brief Generic spatial relationship for the trajectory or traversed area
  * of a temporal geo and a geometry
  * @details The function computes the function passed as parameter with the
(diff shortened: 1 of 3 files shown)


estebanzimanyi:

PR #1015, opened 2026-05-14
Add bbox prefilter to ea_dwithin_tgeo_geo

estebanzimanyi:
ea_dwithin_tgeo_geo previously called geom_buffer(gs, dist) for every join row in the ALWAYS branch and entered the per-segment iteration for every join row in the EVER branch, even when the temporal value's bbox lay entirely outside the geometry's dist-expanded bbox. Adds an STBox prefilter that mirrors the pattern already used by the sibling ea_intersects_tpoint_geo and ea_touches_tgeo_geo at lines 1228 and 1330 of the same file: temp's bbox must overlap (EVER) or be contained in (ALWAYS) the geom's bbox expanded by dist; otherwise the predicate is short-circuited to false. The prefilter is gated to planar inputs because the geodetic STBox is in degrees while dist is in metres. Verified on the AIS tcbuffer bench: aDwithin(geom, Trip, 100) drops from 20.1 s to 224 ms (90 times faster) over a 10000-pair cross-join, matching the eDwithin cost on the same query. Stacks on #1014; the bench measurement assumes #1014's always-quantifier correctness fix is in place.

### DIFF
--- a/meos/src/geo/tgeo_spatialrels.c
+++ b/meos/src/geo/tgeo_spatialrels.c
@@ -295,21 +295,24 @@ geo_dwithin_fn_geo(int16 flags1, uint8_t flags2)
  * @brief Generic spatial relationship for the trajectory or traversed area
  * of a temporal geo and a geometry
  * @details The function computes the function passed as parameter with the
- * trajectory or the traversed area of the temporal geometry and the geometry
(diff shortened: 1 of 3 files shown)


estebanzimanyi:

PR #1016, opened 2026-05-14
Lift point-only guard on tdwithin_tgeo_geo for non-point geometries

estebanzimanyi:
tdwithin_tgeo_geo previously rejected non-point geometries on the static side with "Only point geometries accepted", because the turning-point function tpointsegm_tdwithin_turnpt reads both sides as POINT2D_P. The lifting infrastructure (tdwithin_tlinearseq_base_iter) also has a structural 2-turning-point-per-segment limit that makes a fully analytical moving-point-vs-polygon kernel a substantive refactor of its own. For non-point geometries this routes the call through tinterrel_tgeo_geo on a d-expanded polygonal buffer of the geometry, which is the same polygonal approximation that ea_dwithin_tgeo_geo's always-quantifier branch already uses internally via geom_buffer. Adds positive coverage for Linestring, Polygon, and MultiPolygon static side, and drops the no-longer-erroring linestring entries from the Errors section of 072_tpoint_tempspatialrels. Stacks on #1014 and #1015. The polygonal-buffer approximation is an upgrade target for a later PR that would rework the turnpt infrastructure to support N turning points per moving segment.

### DIFF
--- a/meos/src/geo/tgeo_spatialrels.c
+++ b/meos/src/geo/tgeo_spatialrels.c
@@ -295,21 +295,24 @@ geo_dwithin_fn_geo(int16 flags1, uint8_t flags2)
  * @brief Generic spatial relationship for the trajectory or traversed area
  * of a temporal geo and a geometry
  * @details The function computes the function passed as parameter with the
- * trajectory or the traversed area of the temporal geometry and the geometry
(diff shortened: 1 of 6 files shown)


estebanzimanyi:

PR #1019, opened 2026-05-14
Enable tDwithin (geometry, tcbuffer, d) and (tcbuffer, geometry, d) SQL bindings

estebanzimanyi:
The MEOS kernel tdwithin_tcbuffer_geo (meos/src/cbuffer/tcbuffer_tempspatialrels.c) and the PG wrapper Tdwithin_geo_tcbuffer (mobilitydb/src/cbuffer/tcbuffer_tempspatialrels.c) are fully implemented. The matching SQL CREATE FUNCTION statements in 164_tcbuffer_tempspatialrels.in.sql were left commented out. tinterrel_tcbuffer_geo already accepts arbitrary geometry types so the overload works for any input geometry shape.


estebanzimanyi:
Subsumed by #970. Enabling the tDwithin(geometry, tcbuffer) and tDwithin(tcbuffer, geometry) SQL bindings (IMMUTABLE STRICT, backed by Tdwithin_geo_tcbuffer / Tdwithin_tcbuffer_geo) is already delivered there as part of the STRICT and restype fixes; this branch adds no further C kernel or surface.

PR #1023, opened 2026-05-15
Avoid materialising the traversed area in tcbuffer nad and shortestLine

estebanzimanyi:
The nearest approach distance and shortest line between a temporal circular buffer and a geometry, circular buffer or spatiotemporal box built the entire traversed area (a collection of per-segment swept capsules) and ran one GEOS call against it on every invocation, so a cross-join rebuilt the same trip's traversed area for every right-hand row and that dominated the wall time. Since the distance to a collection is the minimum of the distances to its members and the shortest line is the one to the nearest member, the result is now computed by scanning the per-segment swept capsules with a running minimum, skipping any segment whose bounding box (the centre expanded by the radius, which contains the swept capsule of a linearly moving, linearly resized disk) is provably farther than the current best so it is never built. The per-segment swept capsules are the exact geometries the previous code measured, so the numeric result is unchanged; this was verified across instant, linear, step and sequence-set inputs against point, overlapping-polygon, crossing-line and far-multipolygon geometries, and the 160_tcbuffer_distance regression suite passes unchanged. The pre-existing 158_tcbuffer_topops failure on master is unrelated to this change (it exercises the -|- adjacency operator, not distance) and is addressed separately.

### DIFF
--- a/meos/src/cbuffer/tcbuffer_distance.c
+++ b/meos/src/cbuffer/tcbuffer_distance.c
@@ -34,6 +34,7 @@
 
 /* C */
 #include <float.h>
+#include <math.h>
 /* MEOS */
 #include <meos.h>
 #include <meos_internal.h>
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded by the GEOS-free analytic nearest-approach and shortest-line kernel for tcbuffer and geometry in #1029 and #1030. Those rewrite the same `tcbuffer_distance.c` entry points and compute the result analytically, without scanning the per-segment traversed-area capsules this branch relies on.

PR #1025, opened 2026-05-15
Expose Edge struct and edge-rtree helpers via geo/tpoint_geom_clip.h

estebanzimanyi:
The geometry-edge enumeration, the per-edge R-tree builder and the ray-casting point-in-polygon test in tpoint_geom_clip.c were file-local. This declares the Edge struct, EdgeType, geom_extract_edges, build_edge_rtree and point_in_polygon in a new geo/tpoint_geom_clip.h and removes the duplicate typedefs from the source; it is a visibility-only change. Note that tpoint_geom_clip.c is currently dormant (not listed in meos/src/geo/CMakeLists.txt and not API-current), so this header is preparatory: a separate change must port that file to the current MeosArray and R-tree APIs and add it to the build before the declared symbols are actually linkable. It is the first step toward the GEOS-free analytic capsule-vs-polygon distance path.

### DIFF
--- a/meos/include/geo/tpoint_geom_clip.h
+++ b/meos/include/geo/tpoint_geom_clip.h
@@ -1,0 +1,84 @@
+/*****************************************************************************
+ *
+ * This MobilityDB code is provided under The PostgreSQL License.
+ * Copyright (c) 2016-2025, Université libre de Bruxelles and MobilityDB
+ * contributors
+ *
+ * MobilityDB includes portions of PostGIS version 3 source code released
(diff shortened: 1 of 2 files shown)


estebanzimanyi:
Superseded: the analytic nearest-approach distance kernel in #1029 is self-contained (inline lwgeom edge extraction + ray-cast point-in-polygon) and needs neither the dormant tpoint_geom_clip.c, its R-tree subsystem (which is not linked into the MobilityDB extension), nor this header. Closing as unnecessary for that work.

Issue #821, opened 2026-04-29
Inconsistency: set_spanset canonicalizes adjacent spans, but SQL regression tests expect non-canonical output


estebanzimanyi:
## Summary

`set_spanset()` in `meos/src/temporal/spanset.c` calls `spanset_make_free(spans, s->count, NORMALIZE, ORDER_NO)` with the `NORMALIZE` flag, which merges adjacent single-value spans created from a Set into one wider span. The SQL regression tests, however, expect the non-normalized form.

## Reproducer

\`\`\`sql
SELECT spanset(dateset '{2000-01-01, 2000-01-02}');
-- MEOS / MobilityDuck (current): {[2000-01-01, 2000-01-03)}
-- MobilityDB regression expectation: {[2000-01-01, 2000-01-02), [2000-01-02, 2000-01-03)}

SELECT dateset '{2000-01-01, 2000-01-02}'::datespanset;
-- Same divergence as above.
\`\`\`

## Source

\`\`\`c
// meos/src/temporal/spanset.c, line ~417
SpanSet *
set_spanset(const Set *s)
{
  ...
  for (int i = 0; i < s->count; i++)
  {
    Datum value = SET_VAL_N(s, i);
    span_set(value, value, true, true, s->basetype, spantype, &spans[i]);
  }
  return spanset_make_free(spans, s->count, NORMALIZE, ORDER_NO);
//                                          ^^^^^^^^^
}
\`\`\`

Compare with `span_to_spanset` (line ~449) which uses `NORMALIZE_NO`.

## Possible resolutions

1. **Switch the flag to `NORMALIZE_NO`** to match SQL test expectations and the related `span_to_spanset` behavior. One-line change, but other callers of `set_spanset` may rely on canonicalization — needs an audit.
2. **Update the regression tests** to expect the canonicalized output. MEOS's current behavior arguably produces a more compact and computationally cheaper representation.
3. **Add a parameter** `set_spanset_ext(s, bool normalize)` and have the SQL wrapper pass `false` to preserve current SQL semantics, while internal MEOS callers can opt into either.

This was surfaced while building parity tests for MobilityDuck; the test in question lives at `test/sql/parity/007_spanset.test`. Tagging @mschoema for triage on intended semantics.

estebanzimanyi:
Apologies — I went back and re-checked the MobilityDB regression file `mobilitydb/test/temporal/expected/007_spanset.test.out` (line 93 in current `master`):

[...]

PR #1026, opened 2026-05-15
Pin the canonical span adjacency edge cases (tests)

estebanzimanyi:
Adds discriminating regression tests for span adjacency (`-|-`) in `005_span_ops` so any change back to a drop-the-bits predicate fails loudly: the four float inc/exc combinations at a shared boundary and the discrete consecutive/overlap cases. The canonical rule itself is documented on `adjacent_span_span` in the comment-only #1121; this PR is the automated guard. Note: the regression is already caught by the existing suite (`003_span`'s `floatspan '[3.5,5.5]' -|- 5.5` flips under drop-the-bits) — these cases make the four-way matrix explicit. Refs #821.

### DIFF
--- a/mobilitydb/test/temporal/expected/005_span_ops.test.out
+++ b/mobilitydb/test/temporal/expected/005_span_ops.test.out
@@ -52,6 +52,42 @@ SELECT floatspan '[1, 3]' -|- floatspan '[1, 3]';
  f
 (1 row)
 
+SELECT floatspan '[1, 5]' -|- floatspan '[5, 9]';
+ ?column? 
+----------
+ f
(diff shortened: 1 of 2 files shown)


estebanzimanyi:
Superseded by #1123 (adopt canonical set-theoretic share-a-boundary adjacency, issue #821). Under share-a-boundary (B), adjacency does not consult the inclusive/exclusive bits, so the four float inc/exc cases this PR pinned all become `t` — the discriminating matrix loses its purpose. #1123 adopts B on master and regenerates the topology guards. Closing; a focused B-regression case can be re-added later if wanted.

PR #1027, opened 2026-05-15
Vendor h3-pg v4.2.3 under h3-pg/

estebanzimanyi:
Imports the third-party h3-pg PostgreSQL extension and bundled libh3 verbatim under h3-pg/, at upstream tag v4.2.3 (postgis/h3-pg commit a26630b8353d441e6bc8065c0a8dcaa3d89ef87b), together with the scripts/h3pg_import extraction tooling and the .codacy.yml exclusion of the vendored tree from static analysis, mirroring the existing postgis / postgres / pgtypes exclusions since we do not own h3-pg's style. Reviewable by provenance against upstream.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -19,12 +19,26 @@ engines:
       - "postgis/**"
       - "postgres/**"
       - "pgtypes/**"
+      - "h3-pg/**"
     config:
       languages:
         - "c"
+# Exclusions:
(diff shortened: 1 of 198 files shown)


estebanzimanyi:
Superseded by #1164 (the temporal H3 family PR vendors h3-pg in its own commit). Every line this PR contributes is present in the consolidated deliverable (verified against the assembled firm-pin tree). Branch retained.

PR #1029, opened 2026-05-15
Add tcbuffer distance functions

estebanzimanyi:
Add the distance surface for circular buffers: the analytic nearest-approach distance and shortest line against a geometry with edge-rtree and bounding-box pruning of the swept area, the touch-relation prefilter, the minDistance aggregate with a closed-form segment-pair kernel and an STBox prune, and the STBox prefilter for adwithin.

PR #1030, opened 2026-05-16
Compute the shortest line of a tcbuffer and a geometry analytically

estebanzimanyi:
shortestLine between a temporal circular buffer and a geometry built the whole traversed area and called GEOS on every invocation. It now reuses the closed-form nearest-approach minimisation and additionally tracks the witness, the parameter on the swept-capsule segment that attains the minimum and the nearest point on the geometry edge, and builds the connecting line from those two points with lwgeom, with no GEOS call. A swept-capsule centre inside a polygon yields a zero-length line at that interior point, mirroring the nearest-approach interior short-circuit, and curved or unsupported geometry falls back to the exact traversed-area shortest line. The connecting line's length equals the analytic nearest-approach distance and its geometry-side endpoint lies on the geometry, verified across instant, linear, step, discrete and sequence-set inputs over point, line, polygon, polygon-with-hole and multipolygon geometries, with matching coverage added to 160_tcbuffer_distance. Stacks on the analytic nearest-approach distance change; the combined diff shows both until that one merges.


estebanzimanyi:

PR #1031, opened 2026-05-16
Prefilter the tcbuffer touch relations with the exact nearest-approach distance

estebanzimanyi:
Touching requires the temporal circular buffer and the geometry to be at distance zero, so the exact analytic nearest-approach distance is computed once per pair before the per-segment touch evaluation: a strictly positive distance means the two are disjoint and cannot touch under either the ever or the always quantifier, returning 0 without building any swept area or calling GEOS, while a zero distance falls through to the existing exact per-segment path so genuine boundary contacts are unaffected. eTouches and aTouches in both argument orders return identical results with and without the prefilter across instant, linear, step, discrete and sequence-set inputs over disjoint, near-miss, boundary-touching, interior-overlapping and containing geometries. Stacks on the analytic nearest-approach distance and shortest line; the combined diff shows them until those merge.


estebanzimanyi:

PR #1033, opened 2026-05-16
Add MEST RTree selectivity example

estebanzimanyi:
Add a standalone MEOS example that demonstrates the multi-entry RTree selectivity gain over the single-box index for wiggly, high-extent temporal points. The program generates a deterministic set of zig-zag random-walk tgeompoint trips whose per-trip minimum bounding box is loose, builds one single-box index with rtree_insert_temporal and several multi-entry indexes with rtree_insert_temporal_split for a sweep of maxboxes, and queries each with rtree_search_temporal and rtree_search_temporal_dedup using small probe trips. It compares the candidate sets against an exact ground truth computed independently of the index from the finest per-segment boxes, prints a table of candidate counts and false positives per configuration, and asserts the MEST invariants: no false negatives, dedup uniqueness, candidate count no larger than the single-box index and tightening as maxboxes grows, and maxboxes no greater than one reproducing the single-box candidate set. The summary doubles as the benchmark for choosing a default maxboxes. This is stacked on the MEST API in #1032 and contains that branch's commit until it merges.


estebanzimanyi:
Subsumed by #1032 (MEST regrouping). The selectivity example is included in the grouped PR.

PR #1034, opened 2026-05-16
Vendor Apache Arrow C Data Interface

estebanzimanyi:
Add the Apache Arrow C Data Interface ABI header (ArrowSchema and ArrowArray plus their release callbacks) verbatim from the Apache Arrow project under a dedicated top-level arrow/ directory, mirroring the existing vendored trees. The header is the stable cross-language ABI specification, reproduced so MEOS can produce and consume Arrow columnar data without linking the Apache Arrow C++ library; it contains no first-party code and is reviewed by provenance against upstream. It is exposed to the build as a SYSTEM include directory and installed alongside the MEOS public headers, and arrow/ is excluded from Codacy like the other vendored sources since we do not own its style guidelines.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -19,12 +19,18 @@ engines:
       - "postgis/**"
       - "postgres/**"
       - "pgtypes/**"
+      - "arrow/**"
     config:
       languages:
         - "c"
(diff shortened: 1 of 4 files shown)


estebanzimanyi:

PR #1035, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal float

estebanzimanyi:
Add meos_temporal_to_arrow and meos_temporal_from_arrow, converting a Temporal value to and from the Apache Arrow C Data Interface without linking libarrow, so any Arrow engine can read MEOS temporal data and the conversion is the shared kernel for vectorised TemporalParquet. The Arrow form is the nested Struct{subtype,interp,flags,srid,seqs:List<Struct{lower_inc,upper_inc,insts:List<Struct{t,v}>}>} contract; from_arrow dispatches on the carried subtype and interpolation. All tfloat subtypes (instant, sequence, sequence set) across discrete, step, and linear interpolation round-trip with value identity; any non-float value leaf raises MEOS_ERR_FEATURE_NOT_SUPPORTED rather than sampling or approximating. Includes a standalone round-trip example covering nine shapes. Stacked on #1034 (Vendor Apache Arrow C Data Interface); since the base must be master on the upstream repo, that vendor commit appears in this diff until #1034 merges.


estebanzimanyi:

PR #1036, opened 2026-05-16
Emit SIZEOF_LONG_LONG in the rendered pg_config.h

estebanzimanyi:
postgres/port/pg_bitutils.h selects __builtin_clzll/ctzll through SIZEOF_LONG_LONG when SIZEOF_LONG is not 8, but the narrowed pg_config.h.in template never emitted that macro, so on ILP32 targets such as wasm32-emscripten it expanded to 0 and the file hit its "cannot find integer type of the same size as uint64_t" #error, breaking the DuckDB-Wasm builds in the downstream MobilityDuck CI. On LP64 hosts the SIZEOF_LONG == 8 branch wins first, which is why this was never observed before. This renders SIZEOF_LONG_LONG from the value ConfigurePgConfig.cmake already computes, mirroring what PostgreSQL's own configure produces; verified that the configured pg_config.h gains the define on a native build with zero change to the LP64 path.

### DIFF
--- a/cmake/ConfigurePgConfig.cmake
+++ b/cmake/ConfigurePgConfig.cmake
@@ -25,6 +25,13 @@ check_type_size("long"            SIZEOF_LONG             BUILTIN_TYPES_ONLY)
 check_type_size("size_t"          SIZEOF_SIZE_T)
 check_type_size("long long int"   SIZEOF_LONG_LONG_INT    BUILTIN_TYPES_ONLY)
 
+# postgres/port/pg_bitutils.h selects __builtin_clzll/ctzll via
+# SIZEOF_LONG_LONG when SIZEOF_LONG != 8 (e.g. wasm32-emscripten and other
+# ILP32 targets). Mirror PG's own configure, which emits SIZEOF_LONG_LONG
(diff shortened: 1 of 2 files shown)


estebanzimanyi:
Closing as subsumed by #751 (the pgtypes vendoring). This PR targeted `postgres/pg_config.h.in` + `cmake/ConfigurePgConfig.cmake`, but #751 relocated the pg_config machinery into `pgtypes/`, where the same outcome is already achieved on master:
- `pgtypes/CMakeLists.txt:36` — `check_type_size("long long" SIZEOF_LONG_LONG)` defines the variable;
- `pgtypes/pg_config.h.in:23` — `#define SIZEOF_LONG_LONG @SIZEOF_LONG_LONG@`;
- `pgtypes/CMakeLists.txt:93` — `configure_file(pg_config.h.in pg_config.h @ONLY)` substitutes it.

So `pg_bitutils.h`'s `SIZEOF_LONG == 8 / SIZEOF_LONG_LONG == 8` selection works on ILP32/wasm targets without this change. Nothing left to merge.

PR #1037, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal integer

estebanzimanyi:
Extend the Arrow C Data Interface conversion to temporal integer by dispatching the value leaf on the base type: temporal float decomposes to a Float64 column and temporal integer to an Int32 column, with the rest of the nested schema unchanged. All temporal integer subtypes (instant, sequence, sequence set) and bounds round-trip losslessly; the reader gates on the value-leaf format so a non-numeric column still routes to MEOS_ERR_FEATURE_NOT_SUPPORTED rather than being misread. The SQL surface gains arrowRoundtrip(tint), the pg_regress test covers every temporal integer subtype alongside the existing float cases, and the manual is updated in English and Spanish. Stacked on #1035; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1035 merges.


estebanzimanyi:

PR #1038, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal boolean

estebanzimanyi:
Extend the Arrow C Data Interface conversion to temporal boolean: the value leaf now also dispatches to a bit-packed Boolean column ("b", one bit per instant in the Arrow boolean layout) alongside the Float64 and Int32 columns, with the rest of the nested schema unchanged. All temporal boolean subtypes (instant, sequence, sequence set) and bounds round-trip losslessly; the reader still gates on the value-leaf format so an unrecognised column routes to MEOS_ERR_FEATURE_NOT_SUPPORTED rather than being misread. The SQL surface gains arrowRoundtrip(tbool), the pg_regress test covers every temporal boolean subtype alongside the existing float and integer cases, and the manual is updated in English and Spanish. Stacked on #1037; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1037 merges.


estebanzimanyi:

PR #1039, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal text

estebanzimanyi:
Extend the Arrow C Data Interface conversion to temporal text. The value leaf now also dispatches to a variable-length Utf8 column ("u": int32 offsets plus a byte buffer, a three-buffer Arrow array node) alongside the fixed-width Float64, Int32 and bit-packed Boolean columns, with the rest of the nested schema unchanged. All temporal text subtypes (instant, sequence, sequence set) and bounds round-trip losslessly; the reader still gates on the value-leaf format so an unrecognised column routes to MEOS_ERR_FEATURE_NOT_SUPPORTED rather than being misread. This completes the scalar tier. The SQL surface gains arrowRoundtrip(ttext), the pg_regress test covers every temporal text subtype with values of differing length to exercise the offsets, and the manual is updated in English and Spanish. Stacked on #1038; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1038 merges.


estebanzimanyi:

PR #1040, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal points

estebanzimanyi:
Extend the conversion to temporal geometry and geography points. The value leaf now also dispatches to a Struct of parallel Float64 x, y and (when 3D) z columns; the struct child count encodes the dimension and the geometry/geography distinction is recovered from the flags word carried verbatim. The previously reserved top-level srid slot now carries the point SRID and a non-zero srid is accepted only for the point value leaf; scalar leaves still require srid zero, so existing scalar Arrow data reads unchanged and the change is backward compatible. Reconstruction uses geopoint_make, defined in both the MEOS and the MobilityDB extension builds, so the point split constructors that exist only in the MEOS build are not referenced. All temporal point subtypes and bounds round-trip losslessly including the SRID and the Z dimension. The SQL surface gains arrowRoundtrip for tgeompoint and tgeogpoint, declared in the geo input-output module so the point types exist at load time; the pg_regress test covers 2D, 3D, a non-zero SRID, geometry and geography across every subtype, and the manual is updated in English and Spanish. Stacked on #1039; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1039 merges.


estebanzimanyi:

PR #1042, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal poses

estebanzimanyi:
Extend the conversion to temporal poses. The Struct value leaf gains two more shapes: a 2D pose decomposes to parallel Float64 x, y and theta columns; a 3D pose to x, y, z and the unit quaternion W, X, Y, Z. The reader disambiguates the decomposed extended types by the third struct child name and the child count: r is a circular buffer, theta a 2D pose, a seven-child struct a 3D pose, z a 3D point, otherwise a point; the srid slot carries the pose SRID like the other spatial types. Reconstruction uses pose_make_2d and pose_make_3d. As with the circular buffer, every pose reference is guarded by #if POSE because temporal_arrow.c is in the always-built temporal module while the pose subsystem is behind the POSE build option, and a pose Arrow value presented to a build without POSE routes to MEOS_ERR_FEATURE_NOT_SUPPORTED rather than being misread. All temporal pose subtypes and bounds round-trip losslessly including the SRID, the 2D rotation and the 3D quaternion. The SQL surface gains arrowRoundtrip(tpose) in the pose module, the pg_regress test covers 2D and 3D across every subtype with a non-zero SRID, and the English manual is updated; the pose manual has no Spanish mirror in the tree. Stacked on #1041; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1041 merges.


estebanzimanyi:

PR #1043, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal network points

estebanzimanyi:
Extend the conversion to temporal network points, completing the decomposed type tier. The Struct value leaf gains its first mixed-type shape: a network point decomposes to an Int64 rid column and a Float64 pos column; the reader recognises it by the first struct child being named rid. The network point value carries no spatial reference of its own, so the srid slot stays zero for it. Reconstruction uses npoint_make, which requires the network to be loaded, exactly as parsing a network point literal does; the arrow form carries the route id and the position fraction losslessly. As with the other optional subsystems, every network point reference is guarded by #if NPOINT because temporal_arrow.c is in the always-built temporal module, and a network point Arrow value presented to a build without NPOINT routes to MEOS_ERR_FEATURE_NOT_SUPPORTED rather than being misread. The coverage workflow built only with -DCBUFFER -DPOSE -DRGEO, so the same commit adds -DNPOINT=ON to the pgversion.yml coverage step; without it the new code would be uncovered. The SQL surface gains arrowRoundtrip(tnpoint) in the network point module, the pg_regress test covers every subtype across several routes under the network fixture, and the manual is updated in English and Spanish. Stacked on #1042; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1042 merges.


estebanzimanyi:

PR #1044, opened 2026-05-16
Add tbigint temporal big integer type

estebanzimanyi:
Add tbigint, a temporal big integer base type (temporal int64), parallel to tint, threaded through the temporal/number/span/set/box/aggregate/distance infrastructure with its SQL surface, pg_regress coverage and manual entries. This is the standalone first-party part of the bundled #976 commit that originally added both tbigint and the temporal H3 cell index together; carved out (Option B, #976 left intact, exact-reproduction gate) so the general temporal big integer type lands on its own merits with no dependency on the vendored libh3/h3-pg subtree. Rebased onto #1050 so the canonical 50-per-type SQL renumber of the npoint, cbuffer, and rgeo directories is inherited from #1050 rather than bundled here; the generated mobilitydb--1.4.0.sql is byte-identical before and after the rebase, the only tree change being the removal of stale unreferenced old-numbered SQL files that the renames in #1050 already supersede. The temporal H3 cell index (th3index, a tbigint with an H3-specific API) and its vendored dependency (#1027) follow as a separate stacked change; a few inert H3 build and documentation scaffolding hunks are deliberately left untouched here and reconciled there. No h3 reference remains in the compiled code or the SQL surface. Merge order is #1050, then this.


estebanzimanyi:
Superseded by #1163, the clean re-derived temporal big integer type (4 reviewer-walkable commits, full regression green). Branch retained.

PR #1046, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal geometries and geographies

estebanzimanyi:
Open the opaque type tier with general temporal geometry and geography. Unlike the decomposed types an arbitrary geometry has no fixed column shape, so the value leaf is an opaque LargeBinary carrying each instant's extended well-known binary verbatim; the reader recovers the geometry-versus-geography distinction from the carried flags word since the binary does not encode it, and the spatial reference identifier and dimensionality travel inside the extended binary so the top-level srid slot stays zero for this leaf as for the scalar leaves. The serialization mirrors the canonical geo_as_ewkb and geo_from_wkb_state paths, but the public wrappers are not linked into the MobilityDB extension build so the bodies are reproduced over the liblwgeom primitives available in both builds and the reconstruction uses geom_serialize or geog_serialize so the geodetic flag is set, exactly as the temporal binary reader does. General geometry and geography are part of the always-built geo module, so no build flag guard and no coverage workflow change are needed unlike the optional extended subsystems. The SQL surface gains arrowRoundtrip(tgeometry) and arrowRoundtrip(tgeography) in the geo module, the central temporal Arrow pg_regress test covers linestrings, polygons, a non-zero SRID, three dimensions, mixed geometry sequences and every subtype, and the manual is updated in English and Spanish. Stacked on #1043; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1043 merges.


estebanzimanyi:

PR #1047, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal rigid geometries

estebanzimanyi:
Complete the opaque type tier with temporal rigid geometry. A rigid geometry is a fixed reference geometry moved by a temporal pose, so the value leaf is a Struct whose leading ref child is the LargeBinary extended well-known binary of the shared reference geometry, followed by the pose fields (x, y, theta for a 2D pose, or x, y, z and the quaternion W, X, Y, Z for a 3D pose); the leading child name ref discriminates it on read. The reference geometry is constant across the value so it is serialized once and the bytes repeated per instant to keep the per-instant leaf uniform with the published contract. Reconstruction mirrors the canonical path: the pose fields rebuild a temporal pose which is then wrapped with the reference geometry through geo_tposeinst_to_trgeo, geo_tposeseq_to_trgeo and geo_tposeseqset_to_trgeo, the same composition the textual and binary readers use; the reference binary reuses the build-portable liblwgeom helpers introduced for general geometry and geom_serialize keeps the reference a geometry. Every rigid geometry reference is guarded by if RGEO because temporal_arrow.c is in the always-built temporal module, a rigid geometry Arrow value presented to a build without RGEO routes to MEOS_ERR_FEATURE_NOT_SUPPORTED, and the build without RGEO compiles clean; the coverage workflow already builds with RGEO so no workflow change is needed. The SQL surface gains arrowRoundtrip(trgeometry) and a pg_regress test covers the reference geometry, the SRID and every subtype. The rigid geometry manual chapter is authored separately in the trgeo reference PRs #967 and #969 which restructure doc/temporal_spatial_p1.xml and p2.xml; the arrowRoundtrip listitem belongs in that reference section and is added once those land, with the C documentation carrying it meanwhile. Stacked on #1046; since the base must be master on the upstream repo, that branch's commits appear in this diff until #1046 merges.


estebanzimanyi:

PR #1049, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal big integer

estebanzimanyi:
Extend the conversion to temporal big integer. A big integer decomposes to the Int64 "l" value leaf, exactly as temporal integer uses Int32 "i": the leaf dispatch gains an Int64 case, the value buffer is sized for int64, the walk reads it with DatumGetInt64, and the reader maps the "l" format and a zero srid to T_TBIGINT. The base type is by-value so no other change is needed, and big integer is a core type so there is no build-flag guard and no coverage workflow change, as for temporal integer and float. The SQL surface gains arrowRoundtrip(tbigint) next to the other scalar types, the central temporal Arrow pg_regress test covers every subtype including a value beyond the 32-bit range to exercise the full width, and the manual is updated in English and Spanish. The test uses a large non-sentinel value rather than the 64-bit maximum because a temporal big integer whose value is exactly the 64-bit maximum crashes plain tbigint input and output independently of this conversion, a pre-existing defect in the big integer type itself (the type-maximum collides with the span and bounding-box infinity sentinel); reproducing it needs no Arrow code and it is reported separately rather than worked around here. Stacked on #1047 for the Arrow kernel and on #1044 for the tbigint type; since a fork pull request must target master upstream, both appear in this diff until they merge.


estebanzimanyi:

PR #1052, opened 2026-05-16
Add Arrow C Data Interface conversion for temporal H3 index

estebanzimanyi:
Wire the temporal H3 cell index into the Arrow C Data Interface round-trip kernel. An H3 cell index is a 64-bit value, so it decomposes to the unsigned 64-bit value leaf "L", distinct from the signed 64-bit leaf "l" of temporal big integer; only the carried Arrow format and the reconstructed temporal type differ, the raw 64 bits move through the shared big-integer value path. Adds the arrowRoundtrip(th3index) SQL function, the pg_regress coverage over every subtype and interpolation, and the English and Spanish manual entries. This stacks on the reconciled type lineage: the canonical SQL renumber, the standalone temporal big integer type, the vendored h3-pg subtree, the temporal H3 cell index type, and the temporal big integer Arrow conversion; those ancestors appear in the diff until they merge in order ahead of it.


estebanzimanyi:

PR #1053, opened 2026-05-17
Add the Spanish translation of the circular buffers chapter

estebanzimanyi:
Adds the Spanish translation of the circular buffers chapter, which until now existed only in the English manual. The new doc/es/temporal_circular_buffers.xml mirrors the English doc/temporal_circular_buffers.xml structure exactly, keeping every sect1, sect2, listitem, xml:id, function signature, SQL example and link byte-identical and translating only the human prose using the terminology and conventions of the existing Spanish chapters; it includes the recently added MF-JSON listitems tcbufferFromMFJSON and asMFJSON(tcbuffer). The chapter is wired into the Spanish book in doc/es/mobilitydb-manual.xml at the position mirroring the English book, and the parallel circular buffer reference-index entries plus the cbuffer type-table row are mirrored into doc/es/reference.xml with identifiers and links identical to the English reference and prose translated. This is stacked on PR #1051 (feat/tcbuffer-mfjson) because the English chapter's MF-JSON content lives there until #1051 merges; the base is master and the #1051 commit shows in this diff until #1051 lands, after which the diff reduces to the Spanish files only. The Spanish book builds with the same dblatex toolchain the documentation workflow uses and all changed files are well-formed XML with all xincludes and entities resolving.

### DIFF
--- a/doc/es/mobilitydb-manual.xml
+++ b/doc/es/mobilitydb-manual.xml
@@ -23,6 +23,7 @@
 <!ENTITY temporal_types_analytics SYSTEM "temporal_types_analytics.xml">
 <!ENTITY temporal_types_aggregation SYSTEM "temporal_types_aggregation.xml">
 <!ENTITY temporal_network_points SYSTEM "temporal_network_points.xml">
+<!ENTITY temporal_circular_buffers SYSTEM "temporal_circular_buffers.xml">
 <!ENTITY reference SYSTEM "reference.xml">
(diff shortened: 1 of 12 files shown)


estebanzimanyi:
Superseded by #1157 (its Spanish circular-buffers chapter is folded into the cbuffer family PR). Every line this PR contributes is present in the consolidated deliverable (verified against the assembled firm-pin tree). Branch retained.

PR #1028, opened 2026-05-15
Vendor pgPointCloud v1.2.5 under pointcloud-pg/

estebanzimanyi:
Imports the third-party pgPointCloud PostgreSQL extension verbatim under pointcloud-pg/, at upstream tag v1.2.5 (pgpointcloud/pointcloud commit ca4c54eff6e88cbc5c71a9589f4b349b1e0ad88d), together with the .codacy.yml exclusion of the vendored tree from static analysis, mirroring the existing postgis / postgres / pgtypes / h3-pg exclusions since we do not own pgPointCloud's style. Reviewable by provenance against upstream.

PR #1054, opened 2026-05-17
Add Arrow C Data Interface conversion for temporal point cloud

estebanzimanyi:
Wire tpcpoint and tpcpatch into the Arrow C Data Interface conversion as an opaque LargeBinary value leaf in the same opaque tier as the general temporal geometry and geography. Each instant's pgPointCloud value is carried verbatim as its serialized varlena body, the same canonical encoding the temporal WKB uses (pcpoint_to_wkb_buf / pcvarlena_from_wkb_state); the schema id is resolved out-of-band against the pointcloud_formats catalog so the srid slot stays zero, and the leaf name discriminates a point cloud value from the geometry leaf. Every point cloud symbol is guarded under the POINTCLOUD build flag with a reader reject when it is off, mirroring the rigid-geometry guard pattern; the POINTCLOUD-off build compiles clean and the non-POINTCLOUD arrow regression is byte-exact. Adds arrowRoundtrip(tpcpoint) and arrowRoundtrip(tpcpatch), a POINTCLOUD-gated per-subsystem pg_regress test, and the English manual entries. This is a cross-fork stack on the Arrow stack tip (#1047) plus the vendored pgPointCloud (#1028) and the first-party tpcpoint/tpcpatch types (#1048); those layers show in the aggregate diff until they merge and carry no content change here beyond the faithful .codacy.yml and pgversion.yml unions.


estebanzimanyi:

PR #1056, opened 2026-05-17
Add MF-JSON input and output for the temporal network point type

estebanzimanyi:
Temporal network points previously had no MF-JSON serialization, so they could not round-trip through the Moving Features JSON representation that the other spatiotemporal types support. This adds a parser and emitter for the temporal network point type guarded by NPOINT, using the typestring MovingNetworkPoint and encoding each value as an object with an integer route member and a numeric position member, mirroring the route and getPosition accessors. A network point is a network-relative reference rather than a coordinate, so there is no coordinate or Z/geodetic handling. The type is wired through the input value dispatch, typestring validation and mapping, the output emitter and the three value-emit sites, and the bounding-box emission for the BBOX and CRS options. It adds the tnpoint_from_mfjson MEOS wrapper, the tnpointFromMFJSON and asMFJSON(tnpoint) SQL functions, the English and Spanish documentation in the shared spatial chapter, and pg_regress coverage with golden output captured from a local NPOINT build.

### DIFF
--- a/doc/es/temporal_spatial_p1.xml
+++ b/doc/es/temporal_spatial_p1.xml
@@ -131,7 +131,7 @@ SELECT asEWKT(ARRAY[geometry 'SRID=5676;Point(0 0)', 'SRID=5676;Point(1 1)']);
 					<listitem><para>1: JSON_C_TO_STRING_SPACED</para></listitem>
 					<listitem><para>2: JSON_C_TO_STRING_PRETTY</para></listitem>
 				</itemizedlist>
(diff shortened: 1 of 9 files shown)


estebanzimanyi:
Closing as redundant. The tnpoint MF-JSON parity is already fully covered by the pre-existing open #932 re-split: #951 (parser/emitter/SQL + pg_regress, CI green, identical {"route","position"} schema) and #953 (DocBook in doc/temporal_network_points.xml, which already documents asMFJSON(tnpoint)/tnpointFromMFJSON). This PR was opened without first enumerating existing PRs and duplicates that work; #951 + #953 are the canonical line. No content here supersedes them.

PR #1057, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export against an external consumer

estebanzimanyi:
The MEOS Arrow export was self-checked only by the MEOS reader and pg_regress round-trips, never by a zero-MEOS Arrow consumer. This vendors nanoarrow (the dependency-free Apache-2.0 Arrow C Data Interface implementation, 0.8.0 amalgamation, under nanoarrow/, provenance-reviewed, not linked into libmeos) and adds a standalone validator meos/examples/temporal_arrow_validate.c that hands the temporal circular buffer export to nanoarrow for full specification validation over the whole nested Struct/List tree; tcbuffer exercises the fully decomposed nested-Struct value leaf and is the representative probe for the decomposed tier. A new job in meos.yml builds with -DCBUFFER=ON and runs it in CI. The first run found a real malformed-export bug: the schema struct children arrays inst_kids and seq_kids were stack-local arrays whose addresses were stored into the produced ArrowSchema, so the schema dangled once meos_temporal_to_arrow returned and an external consumer walking it dereferenced freed stack memory; they are now arena-allocated, mirroring the already-correct array-side inst_a_kids/seq_a_kids and the schema-side top_sc in the same function. After the fix nanoarrow full validation passes on every shape and the full pg_regress arrow suite plus the MEOS round-trip example are byte-exact unchanged. This branches off the temporal circular buffer Arrow PR #1041 (109e93a26); that PR's commits appear in this diff until it merges. .codacy.yml excludes the vendored nanoarrow/ tree in the same commit.


estebanzimanyi:

PR #1058, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export for the opaque and rigid geometry types

estebanzimanyi:
Branched off the assembled arrow stack so the opaque LargeBinary value leaf (temporal geometry and geography, both core) and the rigid geometry Struct{ref:LargeBinary, pose} value leaf are present. Carries PR #1057's canonical schema-children arena fix verbatim (its home is #1057; duplicated here only so these leaves can be validated on the full stack, and rebases away once #1057 and the stack land). Vendors nanoarrow 0.8.0 (provenance reviewed, not linked into libmeos), extends meos/examples/temporal_arrow_validate.c to drive tgeometry, tgeography and trgeo and run nanoarrow full validation plus a self round-trip over the whole nested tree, and adds a meos.yml job built with -DCBUFFER -DPOSE -DRGEO that runs it. nanoarrow full-validate passes on all 22 shapes (8 tgeometry, 4 tgeography, 6 trgeo, 4 circular buffer regression-guard). tpcpoint and tpcpatch are out of scope here (the pgPointCloud build environment); their opaque leaf shares the same fixed code path but is not independently exercised by this change. The arrow stack and #1057 commits appear in this diff until they merge.


estebanzimanyi:
Subsumed by #1070 (Arrow validation and demo regrouping). Included verbatim.

PR #1059, opened 2026-05-17
Clean up the point cloud documentation prose

estebanzimanyi:
Carve out the point cloud documentation prose cleanup that follows the type commit in #999 and was deliberately excluded from the policy-compliant split into #1028 (vendored pgPointCloud v1.2.5) and #1048 (first-party tpcpoint/tpcpatch). It rewrites the contrib/pdal/README.md, contrib/pcl/README.md, and the doc/temporal_pointcloud.xml DocBook chapter to describe the supported surface as it stands rather than its verification state at a point in time, replaces the Status and Implementation roadmap sections with capability descriptions, and removes the inline em-dashes, with no code, SQL, test, or vendored change. Because the point cloud documentation files only exist on the forward-ported types branch, this stacks on #1048; since a fork pull request must target master upstream, the vendored tree from #1028 and the first-party types from #1048 appear in this diff until they merge. There is no Spanish mirror for the point cloud chapter so no doc/es change applies. With #1028, #1048, and this change in place, the bundled #999 is fully superseded and can be closed.


estebanzimanyi:
Superseded by #1165, which adds the temporal point cloud types tpcpoint and tpcpatch together with their documentation in the same change. The DocBook chapter doc/temporal_pointcloud.xml and the contrib/pcl/README.md and contrib/pdal/README.md prose in #1165 already match this cleanup, and the chapter additionally documents the Arrow round-trip conversions, so the point cloud documentation surface is complete there.

PR #1061, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export for the pose type

estebanzimanyi:
Extends the nanoarrow external-consumer conformance probe to the temporal pose, the remaining decomposed-Struct value-leaf type the first probe validated only through the circular buffer representative. The validator now builds 2D poses (the Struct{x,y,theta} three-Float64 field set) and 3D poses (the Struct{x,y,z,W,X,Y,Z} seven-Float64 field set with the orientation quaternion) across instant, sequence with inclusive and exclusive bounds, single instant, negative theta, and sequence set, hands each export to nanoarrow ArrowArrayViewValidate at the FULL level, and reports the resolved value-leaf child count so the 2D versus 3D field set is visible in the verdict; the circular buffer, geometry, geography and rigid geometry cases stay as the regression guard, and the conformance job already builds the pose subsystem so it exercises the new cases with its name and comments updated to name poses. There is no kernel change: meos_temporal_to_arrow is unmodified and the full pg_regress arrow suite is byte-exact in this config, with nanoarrow reporting OVERALL PASS for tpose 2D and tpose 3D on the unmodified export. This stacks on the opaque-tier probe branch (validate-arrow-opaque-tier-session) and inherits the schema-children arena fix from the first external-consumer probe, so that branch's commits and that fix appear in the diff until they merge; the inherited tpc_aggfuncs.c POINTCLOUD build-leg red on the integration base is a separate session's concern and is neither introduced nor addressed here.


estebanzimanyi:

PR #1062, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export for the point cloud types

estebanzimanyi:
Extends the nanoarrow external-consumer conformance probe to the temporal point cloud point and patch types, the last value-leaf tier the opaque-tier proof deliberately scoped out due to pgPointCloud build friction. temporal_arrow_validate.c gains per-type tpcpoint and tpcpatch cases (instant, single-instant sequence, discrete sequence, step sequence, sequence set) built the canonical standalone way (pcpoint_hex_in for points, an in-process pgPointCloud patch for patches, the pcid 1 schema registered exactly as tpcbox_rtree.c does) and run through ArrowArrayViewValidate at the FULL level plus the self round-trip, with all twenty-two existing opaque and rigid geometry cases kept as the regression guard. Two pre-existing pgPointCloud build defects that blocked any standalone consumer are fixed alongside by mirroring the canonical solutions: meos/CMakeLists.txt links libpc.a into libmeos.so and installs meos_pointcloud.h under POINTCLOUD (the pointcloud OBJECT library cannot propagate its private link through TARGET_OBJECTS, identical to the duplicate-link block already in mobilitydb/CMakeLists.txt), and meos_initialize installs the bundled libpc.a handlers under POINTCLOUD so a standalone process is symmetric with the PG backend's mobilitydb_init pc_set_handlers, without which pc_schema_from_xml dereferences a NULL allocator and crashes. The Arrow kernel temporal_arrow.c is unchanged and the meos.yml conformance job builds pgPointCloud from the vendored source and MEOS with POINTCLOUD. Stacked on #1058 which carries the #1057 schema-children arena fix verbatim; its and the stack's commits appear in the diff until they merge.


estebanzimanyi:

PR #1063, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export for the network point type

estebanzimanyi:
Extend the standalone nanoarrow conformance probe to the temporal network point, whose Arrow value leaf is a decomposed mixed Struct{rid:Int64, pos:Float64}. The probe builds representative tnpoint values through the canonical tnpoint_in textual form across instant, single-instant sequence, discrete, linear sequence with inclusive and exclusive bounds (boundary positions 0 and 1), step sequence with inclusive and exclusive bounds, and sequence set over routes that exist in the installed ways1000.csv, hands each MEOS-produced ArrowSchema/ArrowArray to nanoarrow, and runs ArrowArrayViewValidate at the FULL level over the whole nested tree plus the MEOS self round-trip; all inherited circular buffer, geometry, geography and rigid geometry cases are kept unchanged as a regression guard. The conformance job is built with -DNPOINT=on so the ways route data the standalone build reads is installed, and its name and comments mention network points. This is stacked on PR #1058 and inherits its #1057 schema-children arena fix and the vendored nanoarrow and validator harness (their commits show in the diff until they merge); the network point decomposed Struct{rid,pos} export is validated independently here, not claimed from the circular buffer or the shared fix. No kernel change was needed: nanoarrow FULL-validate passes on the unmodified export. The known inherited tpc_aggfuncs.c POINTCLOUD coverage-leg red carried by the #1058 base is a separate session's merge-sequence concern, not introduced or addressed by this PR.


estebanzimanyi:

PR #1064, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export for the big integer type

estebanzimanyi:
Stacked on PR #1058 so the conformance harness, the vendored nanoarrow 0.8.0 amalgamation and PR #1057's canonical schema-children arena fix are inherited verbatim (their homes are #1057 and #1058; they rebase away once those land). The temporal big integer scalar Int64 "l" value leaf was never independently validated by the earlier probes, which predate the big integer type, so it is validated here on its own evidence per the no-unverified-parity rule, not inferred from the temporal integer, the other scalars or the shared fix. The validator gains validate_tbigint covering instant, single-instant sequence, exclusive-bound step sequence, discrete set, step sequence and sequence set shapes, the parsed-view leaf locator is widened to recognise the scalar Int64 leaf, and nanoarrow ArrowArrayViewValidate runs at the FULL level plus a self round-trip over the whole nested tree; all existing #1058 cases are kept as the regression guard. Only non-sentinel int64 values are used (largest 9000000000000000000); the type maximum 9223372036854775807 is a documented backend-crashing value and is deliberately never exercised, exactly as the canonical big integer pg_regress coverage avoids it. The big integer is unguarded core, so the conformance job needs no extra build flag; only its name and comments are updated. nanoarrow full-validate passes on all shapes with no export change. The known inherited tpc_aggfuncs.c POINTCLOUD-leg red on the #1058 base is a separate session's merge-sequence concern, not this PR's.


estebanzimanyi:

PR #1065, opened 2026-05-17
Validate the MEOS Arrow C Data Interface export for the H3 index type

estebanzimanyi:
Stacks on the opaque and rigid geometry conformance branch (PR #1058) and inherits its canonical schema-children arena fix from PR #1057; the #1058 commits appear in this diff until that PR merges. Adds a validate_th3index probe to temporal_arrow_validate.c that builds representative temporal H3 index values (instant, single-instant sequence, discrete, step sequence including an exclusive upper bound, sequence set) from canonical valid H3 cell indices and hands each export to nanoarrow for full Arrow C Data Interface validation, focusing on the scalar UInt64 L value leaf that the earlier probes never exercised; nanoarrow resolves it as a UInt64 leaf and FULL-validates it, and all existing opaque and rigid geometry cases are kept as a regression guard. Extends the conformance job to build -DH3=on, locating libh3 with the same recipe as the pgversion.yml coverage leg, and mirrors the existing meos build wiring into linking and installing the H3 subsystem so the standalone MEOS build can compile, link and install meos_h3.h for the validator. The known inherited tpc_aggfuncs.c POINTCLOUD-coverage-leg red is a separate session's reconciliation concern, not introduced or addressed here; the MEOS-only conformance build does not compile that file.


estebanzimanyi:

PR #1066, opened 2026-05-17
Rename and export the trgeometry I/O functions for cross-type API uniformity

estebanzimanyi:
The public MEOS C I/O functions for the rigid-geometry type used the short trgeo_ family prefix, unlike every sibling spatial temporal type which exports t<type>_in/_out (tcbuffer_in, tpose_in, ...). This renames trgeo_out to trgeometry_out and promotes the previously inline, unexported trgeo_in and trgeo_from_mfjson to exported trgeometry_in and trgeometry_from_mfjson declared in meos_rgeo.h, closing the export gap so rigid geometry has the same external C I/O surface as its siblings. The single in-tree caller is updated in lockstep. Internal helpers, the module directory, and the SQL API (already trgeometry*) are intentionally untouched; this is the first of a few small incremental external-API uniformization steps.


estebanzimanyi:
Superseded by #1137, which landed the complete trgeo→trgeometry public-API rename in a single atomic change (kernel, wrappers, SQL AS-clauses, and @csqlfn/@sqlfn/@sqlop tags). Closing as redundant.

PR #1067, opened 2026-05-17
Rename the trgeometry accessor, transformation, restriction and constructor functions for cross-type API uniformity

estebanzimanyi:
Second incremental step of the meos_rgeo.h external C API uniformization (trgeo_* to trgeometry_*), following the I/O-trio step in #1066. It renames the accessor, transformation, restriction and conversion functions plus the trgeoinst_make and geo_tpose_to_trgeo constructors; the constructor form follows the tnpointinst_make precedent, where a single concrete type takes the full type token (the tgeoinst_make short stem is a shared-geo-family exception that does not apply to the rigid-geometry type). The 161 call sites are updated lockstep across the meos and mobilitydb trees so the build links. The SQL surface is already trgeometry* and is unchanged; internal helpers and the rgeo directory are deliberately out of scope. trgeo_to_tpoint also appears in #993 and reconciles trivially at merge; this PR is independent of #1066 (non-overlapping hunks in the shared files).

PR #1069, opened 2026-05-17
Rename the trgeometry distance and comparison functions for cross-type API uniformity

estebanzimanyi:
Final incremental step of the meos_rgeo.h external C API uniformization (trgeo_* to trgeometry_*), following the I/O-trio step in #1066 and the accessor/transformation/restriction/conversion/constructor step in #1067. It renames the distance functions (tdistance, nad, nai, shortestline) and the ever/always plus temporal comparison functions (teq, tne). The 107 call sites are updated lockstep across 5 files in the meos and mobilitydb trees so the build links. The SQL surface is already trgeometry* and is unchanged; internal helpers and the rgeo directory are deliberately out of scope. This PR is independent of #1066 and #1067 (disjoint symbol sets, non-overlapping hunks).


estebanzimanyi:
Superseded by #1137, which landed the complete trgeo→trgeometry public-API rename in a single atomic change (kernel, wrappers, SQL AS-clauses, and @csqlfn/@sqlfn/@sqlop tags). Closing as redundant.

PR #1070, opened 2026-05-17
Validate the Arrow export for all temporal types and demonstrate zero-MEOS Parquet consumption

estebanzimanyi:
Demonstrates zero-MEOS consumption of the MEOS Arrow C Data Interface export end to end: the temporal Arrow to Parquet producer, consumer, and bridge with a reproducible demo entrypoint, plus set/span and temporal Arrow validators checked against the vendored nanoarrow at the FULL validation level.

Stacked on the Arrow export #1071 (which provides `meos_{set,span,spanset,stbox,tbox,temporal}_to_arrow`). This PR adds only the demo-specific surface: the `nanoarrow/` vendored validator and the `meos/examples/` Arrow/Parquet examples.

Each example carries its copy-pastable `gcc` command in the header (the validators add `nanoarrow/nanoarrow.c` to the compile line and build from the repository root with `-I.`), matching the `01-09` MEOS example convention. The temporal validator gates its optional-family checks behind `#if CBUFFER` / `#if RGEO` (doc comments included), so it compiles against whatever families the installed libmeos was built with; the temporal geometry and geography checks are unconditional core coverage. The vendored nanoarrow is excluded from Codacy.


estebanzimanyi:

PR #1071, opened 2026-05-17
Add Arrow C Data Interface export for set, span, and box types

estebanzimanyi:
Export the finite-subset closure types (set, span, span set) and the number and spatiotemporal boxes (tbox, stbox) to the Arrow C Data Interface, with the matching from-Arrow import and round-trip.

The export is gated on the `ARROW` build option: the conversion lives in the `ARROW`-gated `meos/src/arrow/` object library, the public surface is declared in `meos_arrow.h`, and the `arrowRoundtrip` SQL functions, their PostgreSQL wrappers, and the regression tests are all guarded by `ARROW`. The round-trip example uses the MEOS API only.

Stacked on #1041 (the temporal-type Arrow export, which establishes the `meos/src/arrow/` directory and the `#if ARROW` SQL preprocessor), and transitively on #1144 (the vendored Arrow C Data Interface) and #1193 (geodetic poses).

The independent nanoarrow full-conformance validation and the zero-MEOS Parquet consumption demo are not part of this export PR; they are the deliverable of #1070, which stacks on this one.


estebanzimanyi:

PR #1072, opened 2026-05-17
Add Arrow C Data Interface conversion for temporal and spatiotemporal boxes

estebanzimanyi:
Complete the closed-algebra Arrow interchange for the product-domain finite-subset closure types tbox and stbox, the finite rectangular subsets of the time x value and the space x time domains. A box is literally a product of spans, so meos_tbox_to_arrow/meos_stbox_to_arrow compose verbatim the exact shared span Struct builder introduced for set, span and spanset in #1071 (the aw_span_struct/aw_span_from_struct pair), giving one span schema shared between a span as a value and a span as a box component; the rest mirrors the canonical TBox/STBox C structs (the int16 flags word carried verbatim as the authoritative discriminator, the int32 srid slot, flag-gated optional dimensions encoded as omitted children exactly as the temporal point leaf encodes 2D versus 3D by child count, and the spatial extent as raw min/max doubles with no per-axis inclusivity since an stbox carries no spatial inclusivity). Adds the PG wrappers and SQL arrowRoundtrip(tbox)/arrowRoundtrip(stbox), pg_regress tests with captured expected output, the English and Spanish manual entries, and the nanoarrow conformance validator and CI job extension. This is the optional additive MEOS-ARROW columnar encoding only; it does not touch the ratified v1 MEOS-WKB plus native-scalar sidecar substrate or the parallel WKB plus sidecar lane, and is Option A (an explicit typed box in the optional encoding), not the RFC-track sidecar-unify. Stacked on #1071 (set, span and spanset Arrow); its commits and ancestors show in the diff until it merges.


estebanzimanyi:
Subsumed by #1071 (Arrow set/span/box export regrouping). Included verbatim.

PR #1075, opened 2026-05-18
Add portable bare-name function aliases for all operators

estebanzimanyi:
## Portable bare-name function aliases for MobilityDB operators

The ecosystem spans three engines — MobilityDB (PostgreSQL), MobilityDuck (DuckDB),
MobilitySpark (Spark). A query can't run unchanged on all three because MobilityDB's
operators (`<<#`, `<<`, `&&`, `?=`, …) are PostgreSQL custom operators that DuckDB and
Spark SQL can't register. The portable dialect uses only **named functions**, so every
operator needs a callable bare-name form.

`tools/portable_aliases/generate.py` parses every `CREATE OPERATOR` and emits a
`CREATE FUNCTION` alias under the portable bare name **reusing the operator's own
backing C symbol** — **purely additive** (no new C, no overhead, behaviour identical).
**1046 aliases** across the temporal, geo, cbuffer, npoint, pose, rgeo and json
families.

### Positional operators

`<<#`→`before`, `#>>`→`after`, `<<`→`left`, `>>`→`right`, `<<|`→`below`, `|>>`→`above`,
`<</`→`front`, `/>>`→`back` (+ `over*`). **Dimension-aware by construction** (data-driven
scan): `tcbuffer` (2D) gets left/right/below/above + before/after but **no** front/back;
`tjsonb` (time-only, bbox `tstzspan`) gets **only** before/after; `jsonbset` (set) gets
left/right.

### Comparison operators

The temporal (`#`), ever (`?`) and always (`%`) families map to the canonical camelCase
prefix (`t`/`e`/`a`) + `{Eq,Ne,Lt,Le,Gt,Ge}` suffix — `#=`→`tEq`, `?=`→`eEq`, `%=`→`aEq`,
`?<`→`eLt`, … — matching the **json family, which already exposes these names natively**.
Added **alongside** the existing `ever_eq`/`always_eq` names (additive; a family that
already names its function canonically, like json, is left untouched).

### Provable 100% coverage

Every `CREATE OPERATOR` symbol is **classified** (never excluded): positional &
comparison → aliased; topological (`&&`/`@>`/`<@`/`-|-`/`~=` and the tjsonb
time-containment `#@>`/`<@#`) and distance → named directly by their bare backing
functions; the temporalized jsonb access operators (`->`, `->>`, `#>`, `#>>`, `#-`, `?`,
`?&`, `?|`) → portable via their own named backing functions (the json `*_opr`
operator-support convention), resolved by the backing function rather than the shared
symbol. `generate.py --check` = **0 unclassified / 0 unresolved / 0 collisions / 0
invented** across 1046 aliases; `verify_equivalence.sql` proves at runtime each alias
shares the operator's exact `pg_proc` implementation.

### Wiring

Generated files land per family as `<NNN>_portable_aliases.in.sql`; the
`append_portable_aliases()` CMake macro registers them (a new family needs no manifest
edit), feature-gated per family.


estebanzimanyi:
Superseded: the portable-alias generator and the generated bare-name aliases are now on master (delivered together with the portable-aliases CI gate). Closing as redundant to keep the open set lean.

PR #1076, opened 2026-05-18
Fix duplicate SQL file numeric prefix in rgeo

estebanzimanyi:
`mobilitydb/sql/rgeo/` had two files sharing the numeric prefix **133** — `133_trgeo_distance` and `133_trgeo_vclip` — violating the unique-ordered numbering the extension uses to sequence its SQL (the top-level `list(SORT)` orders all `.sql` by filename prefix).

The two are independent: `133_trgeo_vclip` defines `v_clip_*` functions and does not reference the distance functions, nor vice-versa. `133_trgeo_vclip` is therefore renumbered to **`132_trgeo_vclip`** (free slot, keeps it adjacent and correctly ordered), and the `rgeo/CMakeLists.txt` `LOCAL_FILES` list updated accordingly. CMake configure verified.

Pure rename + list reorder; no SQL content change.


estebanzimanyi:
Superseded by #1161, the per-family consolidation that delivers the complete temporal rigid geometry surface (every function this PR adds is present there). Closing as part of the clean-basis re-derive; the branch is retained.

estebanzimanyi:

estebanzimanyi:
Superseded by #1161 (temporal rigid geometry). Multi-dimension audit verified that every function, symbol, fix, and doc/test line this PR contributes is present in the consolidated tree (residuals are only test-golden formatting, or surface delivered under the canonical renamed form, or the superseded cross_type approach replaced by the materialize-per-instant solution). Branch retained for reference.

PR #1079, opened 2026-05-18
Uniformize MEOS public-API output signatures

estebanzimanyi:
## Summary

Uniformizes the output-parameter shape of the MEOS public C API so that
binding generators (PyMEOS, MobilityDuck, Spark, …) see one consistent
convention instead of per-function special cases. Three signature
irregularity classes are driven to zero across the public headers
(`meos.h`, `meos_geo.h`, `meos_cbuffer.h`, `meos_npoint.h`,
`meos_pose.h`, `meos_rgeo.h`, `meos_internal.h`, `meos_internal_geo.h`):

| Class | Functions | Change |
|---|---|---|
| out-param-naming | 41 | the lone output parameter is renamed `result` |
| array-return-shape | 21 | a trailing `int *count` is added to array-returning functions |
| multi-out | 10 | multiple outputs are bundled into one returned struct |

Result: 2536 of 2666 public functions are uniform; the remaining
stateful (78) and plumbing (52) functions are intentional, by-design
exclusions and are left untouched.

## Commits (one per class)

1. **Uniformize lone output parameters to the name `result`** — pure
   parameter renaming, no ABI or call-site impact.
2. **Add a trailing `int *count` to array-returning functions** — every
   in-tree caller in `meos` and `mobilitydb` is updated to pass and use
   the count.
3. **Return a struct from multi-output functions** — new typed result
   structs (`IntSplit`, `FloatSplit`, `TimeSplit`, `IntTimeSplit`,
   `FloatTimeSplit`, `DatumSplit`, `SpaceSplit`, `SpaceTimeSplit`,
   `MvtGeom`, `MinBoundingCircle`); the empty struct doubles as the
   former NULL/false error sentinel.

## Validation

- MEOS library and the PostgreSQL extension build cleanly
  (`Built target MobilityDB-1.4`).
- All in-tree callers updated; behaviour is preserved by construction
  (renames are cosmetic; counts equal the previously recomputed values;
  structs carry the same data and error sentinel).


estebanzimanyi:
Superseded by **#1101** — same content + the 9-callsite test-file fix in `meos/test/setspan_test.c` that was missing here (added trailing `&count` arg to each `*_values` / `*_spans` / `spanset_spanarr` call). See #1101's body for the per-line table. Closing in favor of #1101.

Issue #1087, opened 2026-05-19
Geodetic gaps in temporal spatial relationships: ttouches unsupported + continuous tdwithin turning point is planar


estebanzimanyi:
Surfaced while making the temporal spatial relationships geodetic-consistent (tintersects/tdisjoint = tdwithin(0), tdwithin mirroring tdistance — see the accompanying PR). Two genuine, *pre-existing* geodetic gaps remain, neither introduced by that PR:

### 1. `ttouches` is unsupported for geodetic coordinates
`ea_touches_tgeo_geo` / the touches temporal path guard with `ensure_not_geodetic(temp->flags)` → `MeosInvalidArgValueError: Only planar coordinates supported`. Unlike `tintersects`/`tdisjoint` (which are definitionally `tdwithin(·,·,0)` and now route through the geodetic distance kernel), `ttouches` is a genuine DE-9IM topological predicate (boundary∩ ∧ interiors-disjoint), implemented via planar GEOS `relate`. A geodetic definition/implementation is a real enhancement (sphere topology), not a guard removal. Consumers (PyMEOS etc.) currently document it as planar-only.

### 2. Continuous geodetic `tdwithin`/`tintersects` turning point is planar
`tpointsegm_tdwithin_turnpt` solves the **planar** segment quadratic for where the within-distance predicate changes. For a geodetic sequence it can report the predicate true over an entire segment when the geodesic distance is only briefly within the threshold (e.g. `[POINT(1 1)@t1, POINT(2 2)@t2]` within 2 m of POINT(1 1): pointwise/discrete is correctly {true,false} since the geodesic distance at t2 is ~156 876 m, but the continuous result is all-true). This affects the continuous geodetic path generally, including the temporal-temporal geodetic relationships whose test expectations are correspondingly planar-approximate. A geodesic-exact turning point has no closed form like the planar quadratic and warrants deliberate numerical design.

Both are tracked as enhancements; the consistency/architecture fix (one geodetic distance kernel; dwithin/intersects/disjoint derived from it) ships separately and closes the erroring/inconsistent cases.

estebanzimanyi:
[...]

PR #1088, opened 2026-05-19
Make the temporal spatial relationships geodetic-consistent

estebanzimanyi:
See commit message. Closes the geodetic-spatial-relationship irregularity (tintersects/tdisjoint are tdwithin(0); tdwithin mirrors tdistance). Planar behaviour unchanged; geodetic instant/discrete intersects/disjoint/dwithin now correct instead of erroring. Deeper pre-existing gaps (geodetic ttouches; continuous geodetic turning-point precision) tracked in #1087. Verified against the composed integration surface: full PyMEOS suite 4492 passed.


estebanzimanyi:

PR #1090, opened 2026-05-20
Fix geog_in WKT fall-through NULL dereference

### LINKED ISSUE #1089: geog_in: missing braces around WKT-parse failure branch causes NULL dereference when meos_error returns
### Bug

In `meos/src/geo/postgis_funcs.c` (around line 3732), the WKT branch of `geog_in` is missing braces:

```c
/* WKT then. */
else
{
  if ( lwgeom_parse_wkt(&lwg_parser_result, (char *) str, 
      LW_PARSER_CHECK_ALL) == LW_FAILURE )
    PG_PARSER_ERROR(lwg_parser_result);
    lwgeom = lwg_parser_result.geom;
}
```

`PG_PARSER_ERROR` (defined just above, line 63) expands to `do { meos_error(ERROR, ...); } while(0);` -- it does **not** `return`. Because of the missing braces, only the macro is the `if`-body; the assignment `lwgeom = lwg_parser_result.geom;` runs **unconditionally**. On parse failure, `lwg_parser_result.geom` is `NULL` (set by `lwgeom_parser_result_init`), so the subsequent `if (lwgeom->srid == SRID_UNKNOWN || ...)` at line 3739 dereferences `NULL`.

### Why it usually doesn't crash in standalone MEOS

In standalone, no custom `MEOS_ERROR_HANDLER` is installed → `meos_error(ERROR, ...)` falls into the `fprintf(stderr, ...)` + `exit(EXIT_FAILURE)` branch at `meos/src/temporal/error.c:220` and the process exits before the NULL deref. So the bug is hidden.

### Why it surfaces in MobilityDuck (DuckDB extension)

MobilityDuck loads MEOS via `meos_initialize()` from within a long-lived DuckDB process. When `meos_error(ERROR, ...)` is invoked from inside a query worker, the `exit(EXIT_FAILURE)` in the default handler aborts the entire DuckDB process; depending on atexit/cleanup interactions (threads, stdio buffering, libc destructors), this can present to the user as a SIGSEGV rather than a clean exit. MobilityDuck's `src/geo/stbox_functions.cpp:23-26,1146` documents working around exactly this class of failure (*"avoids MEOS geog_in/geog_area which can SIGSEGV in this extension"*) with a permanent C++ spherical-lon/lat sidestep for stbox area.

### Fix

Add the missing braces *and* make `PG_PARSER_ERROR` defensively `return NULL` after `meos_error` so a future custom handler that does not abort will not fall through:

```c
/* Modified version of PG_PARSER_ERROR */
#if MEOS
#define PG_PARSER_ERROR(lwg_parser_result) \
  do { \
    meos_error(ERROR, MEOS_ERR_INTERNAL_TYPE_ERROR, \
      "%s", lwg_parser_result.message); \
    return NULL; \
  } while(0);
#else
  #include <lwgeom_pg.h>
#endif
```

and / or rewrite the WKT branch in `geog_in` with explicit braces:

```c
else
{
  if (lwgeom_parse_wkt(&lwg_parser_result, (char *) str,
                       LW_PARSER_CHECK_ALL) == LW_FAILURE)
    PG_PARSER_ERROR(lwg_parser_result);
  lwgeom = lwg_parser_result.geom;
  if (! lwgeom)
    return NULL;
}
```

(The `if (! lwgeom)` is belt-and-suspenders in case a future change to `PG_PARSER_ERROR` does not return.)

### Audit of other `PG_PARSER_ERROR` callsites

A grep for `PG_PARSER_ERROR(` should be run; every callsite that uses it without braces is potentially affected by the same fall-through. (Filed companion analysis on MobilityDuck#157 for the second related crash, `geog_area` / `stbox_area` on geodetic 3D / PolyhedralSurface, which is a *separate* MEOS bug deeper in the lwgeom polyhedral-area path -- this issue tracks only `geog_in`.)

estebanzimanyi:
See commit message for the diagnosis + A/B test result (unpatched: SIGSEGV exit 139; patched: PASS). Closes #1089. Addresses the geog_in half of MobilityDuck#157; geog_area / stbox_area geodetic 3D crash remains as a separate, tracked MEOS-C bug. Verified: libmeos + the MobilityDB PostgreSQL extension both build clean (the #1088 lesson).

### DIFF
--- a/meos/src/geo/postgis_funcs.c
+++ b/meos/src/geo/postgis_funcs.c
@@ -1299,8 +1299,12 @@ geo_geo_n(const GSERIALIZED *gs, int n)
 
   LWCOLLECTION *coll = lwgeom_as_lwcollection(lwgeom);
   if (! coll)
+  {
     meos_error(ERROR, MEOS_ERR_INTERNAL_ERROR,
       "Unable to handle type %d in ST_GeometryN", lwgeom->type);
+    lwgeom_free(lwgeom);
+    return NULL;
+  }
 
   /* Handle out-of-range index value */
   n -= 1;
@@ -3300,9 +3304,20 @@ geo_from_text(const char *wkt, int32_t srid)
 
   if (lwgeom_parse_wkt(&lwg_parser_result, (char *) wkt,
       LW_PARSER_CHECK_ALL) == LW_FAILURE )
+  {
     PG_PARSER_ERROR(lwg_parser_result);
+    return NULL;
+  }
 
   lwgeom = lwg_parser_result.geom;
+  /* Same defensive guard as in geog_in: handle the case where a
+   * non-exiting MEOS_ERROR_HANDLER lets execution fall through. */
+  if (! lwgeom)
+  {
+    meos_error(ERROR, MEOS_ERR_TEXT_INPUT,
+      "parse error - invalid geometry");
+    return NULL;
+  }
 
   if ( lwgeom->srid != SRID_UNKNOWN )
   {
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
## Reviewer's quickstart — ~5 minutes

**What this PR does in one sentence:** adds missing `{}` braces (and a `return NULL`) around three `meos_error(ERROR, ...)` calls in `postgis_funcs.c` whose calling site assumes the error aborts, but where the error handler may return — leading to NULL-deref crashes on the fall-through path. Originally surfaced by MobilityDuck #157.

**Files to read (1 file, ~30 lines of context):**
- `meos/src/geo/postgis_funcs.c` — three small braced-block fixes around lines 1300, 3303, 3733 (geo_geometry_n / geo_from_text / geog_in). Each fix is the same shape: `if (cond) { meos_error(...); lwgeom_free(...); return NULL; }`.

**Smoke verification (no env required):**
```bash
git checkout fix/geog-in-wkt-parse-fallthrough
cd build && cmake -DMEOS=ON -DCMAKE_BUILD_TYPE=Release .. && make -j
# Expected: green; the change is localised, all other code paths are untouched.
```

**Behavioural verification (env'd):**
```sql
-- Before this PR: SIGSEGV with a custom non-exit error handler installed
-- After this PR: clean error return
SELECT geog_in('not a valid wkt');
```

**Why it's safe to merge:** purely defensive (adds a `return`); existing happy-path semantics unchanged; covered by the new Wave-2 ratchet (#1096) so the bug class can't silently grow.

estebanzimanyi:

PR #1092, opened 2026-05-20
Make ea_*_tgeo_geo (ever/always) geodetic-consistent

estebanzimanyi:
Companion to #1088 at the ever/always scalar layer. `adwithin_tgeo_geo` is currently unconditionally broken for any geodetic input (returns -1 with "Operation on mixed planar and geodetic coordinates" from the planar geom_buffer path); the EVER variants mostly worked by accident on instant/discrete because spatialrel_tgeo_geo's per-instant path lets the geodetic-aware datum_geog_dwithin selector resolve correctly. This commit routes the geodetic branch through the same lifted temporal-base machinery that tdistance_tgeo_geo uses (`eafunc_temporal_base` + `geo_dwithin_fn_geo`), and defines ea_intersects/ea_disjoint geodetic as ea_dwithin(0)/INVERT mirroring #1088 and MEOS's own static `geog_intersects = geog_dwithin(0.0)` convention. `eafunc_temporal_base` lives in `lifting.c` (not `*_meos.c`) so this fix is extension-safe (verified `nm -D --undefined-only libMobilityDB-*.so` clean). A/B-verified locally: 0 regressions, 2 strictly-broken paths (adwithin always Instant/Discrete) unblocked. The continuous-segment turning-point precision is unchanged (pre-existing #1087 item; tpfn_base = NULL for now).

### DIFF
--- a/meos/src/geo/tgeo_spatialrels.c
+++ b/meos/src/geo/tgeo_spatialrels.c
@@ -935,6 +935,16 @@ ea_disjoint_tgeo_geo(const Temporal *temp, const GSERIALIZED *gs, bool ever)
   /* Ensure the validity of the arguments */
   if (! ensure_valid_tgeo_geo(temp, gs) || gserialized_is_empty(gs))
     return -1;
+
+  /* tdisjoint is the negation of tintersects = tdwithin with zero
+   * distance, with the ever/always quantifier swapped (cf. static
+   * geog_disjoint = ! geog_dwithin(., ., 0.0); #1088 temporal-result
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded by **#1100** — same content, plus the expected-output update in `mobilitydb/test/geo/expected/070_tpoint_spatialrels_tbl.test.out` for the 4 query counts that shifted with the planar→spheroid semantics. See #1100's body for the count table. Closing in favor of #1100.

Issue #1091, opened 2026-05-20
Two MEOS design questions surfaced during the SIGSEGV diagnosis: epsilon-for-geog-intersects, error-handler default


estebanzimanyi:
Two design-judgment items are resolved on the user's fork integration branch as part of the 15-PR MEOS-1.4 bump bundle. The bundle is OPEN, MERGEABLE, and CI-green and awaits the reviewer merge cadence.

## (A) Tolerance for geodetic intersects/disjoint

MEOS has two definitions of "are these two geographies intersecting?":

* `geog_intersects(gs1, gs2, use_spheroid)` (`meos/src/geo/postgis_funcs.c`):

  ```c
  return geog_dwithin(gs1, gs2, PGIS_FP_TOLERANCE, use_spheroid);
  ```

* `eintersects_tgeo_geo(temp, gs)` (`meos/src/geo/tspatial_rels.c`):

  ```c
  return eintersects_tgeo_geo_alike(temp, gs, false, 0.0);
  ```

**Resolution**: the predicate epsilon is `0.0`. `PGIS_FP_TOLERANCE` (`1e-12`) is `lwgeom`'s internal solver-convergence epsilon for `geog_distance_geographic_for_dwithin` and is not a distance threshold for the predicate itself. Delivered by **PR #1094**.

## (B) `meos_error` exit-on-fatal semantics

The default error handler, when no application handler is installed, calls `exit(EXIT_FAILURE)` on any `MEOS_ERROR(...)` of `errlevel >= ERROR`. The behaviour is appropriate for a standalone tool but surprising in an embedded library context (a DuckDB extension, a Python binding, a long-running service).

**Resolution**: `meos_error(int errlevel, int errcode, …)` is contract-documented as "may or may not return" — every caller checks the return value (or sets its return sentinel ahead of the call). Embedders supply their own handler via `meos_initialize_error_handler` that throws or longjmp's instead of `exit`-ing; the default handler is unchanged but no longer load-bearing for embedded use.

Implemented across four waves:

| Wave | Scope | PR |
|---|---|---|
| 1 | Contract documentation in `meos/include/meos.h` + `error.c` | #1095 |
| 2 | Static-analysis CI ratchet — `clang-tidy` rule flags any new call site not honouring the contract | #1096 |
| 3 | Per-site fix of 6 fall-through sites identified by the baseline | #1097 |
[...]

PR #1094, opened 2026-05-20
Use 0.0 as the geographic intersects/disjoint predicate epsilon

estebanzimanyi:
Closes the Q1 half of #1091. Aligns `datum_geog_intersects` / `datum_geog_disjoint` (`tgeo_spatialrels.c`) with PostGIS's `geography_intersects` and MEOS's static `geog_intersects` wrapper, both of which already use `geog_dwithin(., ., 0.0, …)`. Drops the unjustified 10 µm slack from the temporal-dispatch path. lwgeom's internal `PGIS_FP_TOLERANCE` (1e-12) solver-convergence ε stays unchanged — they sit at different layers. Both targets build clean (the #1088 lesson).


estebanzimanyi:

PR #1095, opened 2026-05-20
Document the meos_error return-or-not contract

estebanzimanyi:
Document on `meos.h` and `error.c` that `meos_error(int errlevel, ...)` has an undefined return-or-not contract: depending on the installed error handler it may return control to the caller, so a caller must follow the call with an immediate control transfer (return with a sentinel, goto, break, exit, or continue) and never fall through. Documentation only; no behavioural change.

### DIFF
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -350,6 +350,25 @@ typedef enum
 
 } errorCode;
 
+/**
+ * Report an error at the given level/code with a printf-style message.
+ *
+ * @note Return-or-not contract is **undefined**: this function MAY
+ * return control to the caller, depending on the installed error
+ * handler. The default handler `exit(EXIT_FAILURE)`s on `ERROR` (safe
+ * for one-shot CLI use), but any custom handler an embedder installs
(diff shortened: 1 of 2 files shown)


estebanzimanyi:
## Reviewer's quickstart — ~3 minutes

**What this PR does in one sentence:** documents that `meos_error(int errlevel = ERROR, ...)` has an *undefined* return-or-not contract — depending on the installed error handler it may return control to the caller — and that callers MUST therefore treat it as a no-return-guaranteed call.

**Files to read (2 files, ~30 lines):**
- `meos/include/meos.h` — adds a doc-comment immediately before the existing `extern void meos_error(...)` declaration.
- `meos/src/temporal/error.c` — adds a brief reference in the function definition itself.

**Smoke verification:** docs-only; the only build step is a re-pass through doxygen / clang's `-Wdocumentation`. No runtime change.

**Why it's safe to merge:** zero runtime behaviour change. Just makes explicit what's been implicit since the embedder-handler mechanism was added.

Issue #1093, opened 2026-05-20
Hardening audit: meos_error(ERROR, ...) fall-through pattern across MEOS


estebanzimanyi:
Tracking the wider audit class that surfaced #1089 / PR #1090. Scope: every `meos_error(ERROR, ...)` callsite in MEOS where execution may *return* to the caller (any non-exiting custom `MEOS_ERROR_HANDLER`), and the next reachable statement dereferences a value that the error path was supposed to abort on. Class is identical to #1089: the bug is invisible in standalone (default handler exits) but SIGSEGVs in long-lived embedders (MobilityDuck, PyMEOS, etc.).

### Scan result

A systematic scan of `meos/src/**.c` (480 `meos_error(ERROR, ...)` callsites total) flagged 70 sites whose next non-blank, non-comment statement was not an immediate `return`/`goto`/`break`/`exit`/`}`. Of those I triaged in depth:

* **3 confirmed bugs — already fixed in PR #1090** (`geog_in`, `geo_from_text`, `geo_geometry_n` in `meos/src/geo/postgis_funcs.c`).
* **Many safe-by-design**: `pfree(str); return false;` patterns; `DATE_NOEND(result)` / `TIMESTAMP_NOEND(result)` sentinel-then-caller-detects patterns; `retval = false; break; ... return retval;` switch-default patterns. The static scan flags them as suspicious because the next line isn't `return`, but the logic correctly anticipates `meos_error` returning.
* **4 additional libmeos-side suspects** worth a deeper look (not fixed in #1090 to keep that PR focused):

  1. `meos/src/geo/tspatial_transform_meos.c:430` — laea projection: `meos_error(ERROR, "Unknown yzone encountered!")` then unconditional `snprintf(strs.proj4text, MAX_PROJ_LEN, "...lat_0=%g +lon_0=%g...", lat_0, lon_0);`. If the error branch was reached, `lon_0` is uninitialised. Undefined behaviour, not necessarily a NULL deref. `*_meos.c` -> libmeos-only (not in the PG extension build).
  2. `meos/src/geo/tspatial_transform_meos.c:557` — `if (! pjstrs_has_entry(&from_strs)) meos_error(ERROR, "got NULL for SRID (%d)", srid_from);` then continues to use `from_strs` for the projection. PROJ initialisation can fault on invalid PjStrs. `*_meos.c` -> libmeos-only.
[...]

PR #1097, opened 2026-05-20
fix: honor meos_error(ERROR, ...) return-or-not contract in 6 fall-through sites

estebanzimanyi:
**Wave 3** of #1091.

These six sites all called \`meos_error(ERROR, ...)\` and then fell through to code that assumes the error aborted execution. Per the contract documented in \`meos/include/meos.h\` (#1095), callers MUST treat \`meos_error(ERROR, ...)\` as having an *undefined* return-or-not contract — when an embedder installs a non-exiting handler, execution continues and the fall-through path runs.

### Site-by-site

| File | Line (master) | Bug class | Fix |
|---|---|---|---|
| \`tspatial_transform_meos.c\` | 430 | Silent wrong-projstring (lon_0 defaults to 0 for unknown LAEA yzone) | Free partial alloc + return empty PjStrs |
| \`tspatial_transform_meos.c\` | 557, 561 | Falls through to \`lwproj_from_str\` with empty PjStrs | Free + return NULL |
| \`trgeo_distance.c\` | 139 (\`append_cfp_elem\`) | OOB write: size field already doubled, but buffer still old size | Restore size + return |
| \`trgeo_distance.c\` | 189 (\`append_tdist_elem\`) | Same pattern as above | Same fix |
| \`trgeo_distance.c\` | 1716 (closest-features loop, \`edge<->edge\`) | Reads uninitialised \`state\` (impl is commented-out) | \`return NULL\` matching surrounding idiom |
| \`trgeo_vclip.c\` | 323 (\`v_clip_tpoly_point\`) | Computes distance from unconverged feature state on cycle detect | \`return MEOS_DISJOINT\` |

### Builds & tests

\`\`\`text
libmeos build (cmake -DMEOS=ON):  green
PG-extension build (cmake):       green
\`\`\`

Each fix adds \`return ...;\` immediately after the \`meos_error\` call, so the static-analysis ratchet (#1096) sees these sites as resolved on its next baseline regeneration.

### Position in the #1091 plan

- ✅ **Wave 1** — Documentation (#1095)
- ✅ **Wave 2** — Static-analysis CI gate (#1096)
- ✅ **Wave 3** — Per-site fixes (**this PR**)
- ⬜ **Wave 4** — Close-read of remaining unexamined audit sites
- ⬜ **Wave 5** *(gated on Wave 4)* — Optional default-handler change

### Refs

\`Refs: #1089, #1090, #1091, #1093, #1095, #1096\`

### DIFF
--- a/meos/src/geo/tspatial_transform_meos.c
+++ b/meos/src/geo/tspatial_transform_meos.c
@@ -425,8 +425,17 @@ GetProjStrings(int32_t srid)
       else if (yzone == 0 || yzone == 5)
         lon_0 = 90.0 * (xzone - 2) + 45.0;
       else
+      {
+        /* See doc-comment on meos_error in meos/include/meos.h: handler is
+         * not guaranteed to abort. Bail explicitly so we don't silently
+         * emit a wrong-but-syntactically-valid projstring with lon_0=0
+         * for an unknown yzone. */
+        pfree(strs.proj4text);
+        strs.proj4text = NULL;
         meos_error(ERROR, MEOS_ERR_INTERNAL_ERROR,
           "Unknown yzone encountered!");
+        return strs;
+      }
 
       snprintf(strs.proj4text, MAX_PROJ_LEN,
         "+proj=laea +ellps=WGS84 +datum=WGS84 +lat_0=%g +lon_0=%g +units=m +no_defs",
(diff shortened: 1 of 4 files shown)


estebanzimanyi:
## Reviewer's quickstart — ~5 minutes

**What this PR does in one sentence:** fixes 6 fall-through `meos_error(ERROR, ...)` sites discovered by inspecting the audit list of #1093 — same bug class as #1090, different files.

**Files to read (3 files):**
| File | Bug class | What was fixed |
|---|---|---|
| `meos/src/geo/tspatial_transform_meos.c` | Silent wrong-projstring + falls-through with empty PjStrs | LAEA unknown-yzone, `GetProjStringsSPI` ×2 |
| `meos/src/rgeo/trgeo_distance.c` | OOB write (size already doubled, but buffer still old size) + uninitialised state read | `append_cfp_elem`, `append_tdist_elem`, "Can't happen" edge↔edge branch |
| `meos/src/rgeo/trgeo_vclip.c` | Computes distance from cycle-detected/unconverged feature state | `v_clip_tpoly_point` |

**Each fix is the same shape:** add braces, add cleanup, add explicit `return`. Aggregate diff: +54 / −5 lines.

**Smoke verification:** local build green (libmeos + PG extension). The CI baseline (#1096) sees these sites as resolved on its next regen.

**Why it's safe to merge:** purely defensive (adds `return`s); existing happy-path semantics unchanged; behaviour change only on the previously-undefined fall-through path.

Issue #22, opened 2020-09-14
Do not allow in-source builds


cvvergara:
**Is your feature request related to a problem? Please describe.**
https://hsf-training.github.io/hsf-training-cmake-webpage/02-building/index.html

> Never do an “in-source” build - that is, run cmake . from the source directory. It will pollute your source directory with build outputs, CMake configuration files, and will disable out-of-source builds. A few packages do not allow the source directory to even sit inside the build directory; if that is the case, you need to change the relative path .. accordingly.

Just to clarify, you can point CMake at either the source directory from the build directory, or at an existing build directory from anywhere.

Currently it allows in source builds:
```
cmake .
-- The C compiler identification is GNU 8.4.0
-- The CXX compiler identification is GNU 8.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
...
```
the source directory gets polluted
```
git status
...
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        CMakeCache.txt
        CMakeFiles/
        CTestTestfile.cmake
        DartConfiguration.tcl
        Makefile
        cmake_install.cmake
```

**Describe the solution you'd like**
```
cmake .
CMake Error at CMakeLists.txt:4 (message):
  In-source builds not allowed.

          Please make a new directory (called a build directory) and run CMake from there.
          You may need to remove 'CMakeCache.txt' and 'CMakeFiles/'.


-- Configuring incomplete, errors occurred!
```

cvvergara:
This can be closed PR #23

PR #1114, opened 2026-05-24
Generalize spatiotemporal box-tiling to tpose and tcbuffer

estebanzimanyi:
`stboxes`, `splitNStboxes` and `splitEachNStboxes` worked only on the geo types. The per-instant box computation is routed through the existing generic `tspatialinst_set_stbox` dispatcher and the `tgeo_type_all` assertions are relaxed to `tspatial_type`, so the same functions cover tpose and tcbuffer without renaming the symbols.

Segment boxes are unions of per-instant boxes, which is exact for both: linear interpolation of center+radius (tcbuffer) and position (tpose) stays within the union of the endpoint boxes, and tcbuffer boxes are radius-expanded as required.

Binds `stboxes`, `splitNStboxes` and `splitEachNStboxes` for tpose and tcbuffer.

Part of the cross-type parity completion; integrated build+test evidence = accumulate (fork PR #22).


estebanzimanyi:
Superseded: the spatiotemporal box-tiling is now carved into the per-type PRs — tcbuffer (stboxes/splitNStboxes/splitEachNStboxes) folded into #1157, and tpose (119_tpose_tile, byte-identical) folded into #1160. No content is lost; closing as fully distributed.

PR #820, opened 2026-04-28
Add trgeo position operators and extent aggregate

estebanzimanyi:
Register the front, back, overfront and overback position operators between stbox and trgeometry, trgeometry and stbox, and trgeometry and trgeometry, together with the extent aggregate over trgeometry, wiring the rigid geometry aggregate SQL into the build.

PR #1115, opened 2026-05-24
Add atValues/minusValues for trgeometry (restrict by pose)

estebanzimanyi:
A trgeometry's base value is a pose, so it is restricted by pose: strip to the temporal pose, apply the value restriction, and re-attach the reference geometry.

Fixes `trgeometry_restrict_value`/`values`, which restricted the un-stripped trgeometry and then re-wrapped it (double-wrap / VALIDATE_TPOSE failure) -- the reason the SQL bindings were disabled. The same strip-rewrap dispatch is added to the generic `temporal_restrict_value`/`values` for direct callers, and the obsolete geometry basetype special-case is dropped.

Binds `atValues`/`minusValues` for (trgeometry, pose) and (trgeometry, poseset).

Coordinate merge order with the open trgeo PR #820.

Part of the cross-type parity completion; integrated build+test evidence = accumulate (fork PR #22).


estebanzimanyi:
Re-opening as a draft stacked on the rename PR #1067, which provides geo_tpose_to_trgeometry/trgeometry_to_tpose. The `test install` failure here was an undefined-symbol error because atValues calls those renamed converters, defined by #1067 (not master). Rebased onto #1067 and verified via a fresh CREATE EXTENSION load test.

PR #1119, opened 2026-05-24
Fix wrong turning point in temporal number multiplication and division

estebanzimanyi:
`tfloat_arithop_turnpt` in `meos/src/temporal/tnumber_mathfuncs.c` uses `Float8GetDatum` (converts `double` → `Datum`) to extract the four endpoint values `x1..x4` from their `Datum` arguments, where `DatumGetFloat8` (extracts `double` ← `Datum`) is required. The vertex computation receives garbage values, the turning point is silently discarded, and the linearly-interpolated product or quotient segment omits the extremum that would otherwise be materialized in the piecewise-linear representation.

Fix uses `DatumGetFloat8`, matching the `Datum`→`double` idiom used throughout the codebase. A pg_regress test covers the zero-crossing case where the incorrect representation is unambiguously wrong.


mschoema:
I'm fine with this PR, but maybe we should add a test in `026_tnumber_mathfuncs.test.sql` that goes thought this code path, as it doesn't seem like that was the case before.

mschoema:
This PR message should also be reworked to follow the guidelines we discussed to have simple postgres-style commit messages without headers or similar things. If this info is intersting for the PR, it can be added in subsequent messages, but the first one should contain the future commit message only.

estebanzimanyi:
Done — two tests are added to `026_tnumber_mathfuncs.test.sql` (and the matching golden updated). The first covers the zero-crossing case `[1@t0, -1@t2]² → [1@t0, 0@t1, 1@t2]` where the incorrect straight-line result is unambiguous; the second (already present as a coverage note) covers the general turning-point case.

estebanzimanyi:
Updated — the PR description is now a single commit-message paragraph with no section headers.

PR #1122, opened 2026-05-25
Initialize upper_inc in the tgeo at-stbox restrict kernel

estebanzimanyi:
`tpointseq_linear_at_stbox_xyz` (meos/src/geo/tgeo_restrict.c) declares `bool upper_inc;` and assigns it on the first statement of its `for (i = 1; i < seq->count; i++)` loop. When `seq->count <= 1` the loop body never runs, so `upper_inc` is read **uninitialised** further down — undefined behaviour.

Initialize it to `false` at the declaration (matching the adjacent `lower_inc`). One-line, behaviour-preserving for the common path (the loop overwrites it before use); only removes the UB on the empty-loop edge.

Surfaced by an integration-coverage audit (`tools/integration_audit/audit.py`) as a fix that existed only on a local integration branch and in no PR — i.e. a crash-orphan that would be lost on accumulate regeneration. This un-strands it.


estebanzimanyi:

PR #1124, opened 2026-05-26
Define the declared-but-unimplemented trgeo accessor functions

estebanzimanyi:
## Problem

Five functions are declared in `meos/include/meos_rgeo.h` but have **no implementation** anywhere in `libmeos`:

```
trgeo_points   trgeo_rotation   trgeo_segments   trgeo_traversed_area   nad_stbox_trgeo
```

They are phantom declarations — any consumer that wraps them (e.g. a language binding generated from the headers) compiles but then fails to **link** with `undefined symbol`. Surfaced while building a WASM binding against the public MEOS surface.

## Fix

Implement the five in the MEOS C library as thin wrappers over functions **already present on master** — no header or declaration changes, so the public API surface is unchanged:

| Function | Implementation |
|---|---|
| `trgeo_points` | `tpose_points(trgeo_to_tpose(temp))` |
| `trgeo_rotation` | `tpose_rotation(trgeo_to_tpose(temp))` |
| `trgeo_segments` | inter-instant rigid-geometry segments via `trgeo_to_tpose` + `temporal_segments` + `geo_tposeseq_to_trgeo` |
| `trgeo_traversed_area` | per-instant `geom_apply_pose` over the reference geometry, collected with `geo_collect_garray` (+ optional `geom_unary_union`) |
| `nad_stbox_trgeo` | argument-swap wrapper over the existing `nad_trgeo_stbox` |

Defined under the current `trgeo_*` names (consistent with the rest of `meos_rgeo.h`); the eventual `trgeo`→`trgeometry` rename will rename these along with everything else.

## Verification

- `libmeos` and `libMobilityDB-1.4` both build clean (CBUFFER/NPOINT/POSE/RGEO on).
- `nm -D libmeos.so` now lists all five as defined (`T`) symbols.

### DIFF
--- a/meos/src/rgeo/trgeo.c
+++ b/meos/src/rgeo/trgeo.c
@@ -41,6 +41,8 @@
 #include <liblwgeom_internal.h>
 /* MEOS */
 #include <meos.h>
+#include <meos_geo.h>
+#include <meos_pose.h>
 #include <meos_rgeo.h>
 #include <meos_internal.h>
 #include "temporal/lifting.h"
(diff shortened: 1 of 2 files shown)


estebanzimanyi:
Subsumed by the trgeometry split chain. trgeo_points, trgeo_rotation and trgeo_segments are implemented in #962 (alongside their MEOS unit-test coverage), and the symmetric nad_stbox_trgeo is defined in #968 next to nad_trgeo_stbox, so every accessor this branch declares is now defined on the trgeometry surface.

PR #1125, opened 2026-05-27
Prune far geometry edges in the analytic tcbuffer nearest approach

estebanzimanyi:
The analytic nearest approach distance of a temporal circular buffer and a geometry scans every geometry edge for every swept-capsule segment, which dominates the cost when a trajectory passes near a complex coastal polygon. This adds a coarse branch-and-bound prune against the geometry bounding box (a lower bound on a segment's nearest approach), runs it ahead of the point-in-polygon ray cast so far segments skip that too, and groups the geometry edges into a Morton-ordered bucket bounding-volume hierarchy used by both the distance loop and the ray cast so a segment only visits nearby edges. Results are unchanged; only provably-farther work is skipped. On one day of Danish AIS at 100 vessels by 100 protected areas the nearest approach to the noise footprint drops from 119.7 s to 47.2 s, verified bit-identical. Builds on the analytic kernel in #1029.


estebanzimanyi:

PR #1126, opened 2026-05-28
Reactivate the tpcpoint GiST index regression test

estebanzimanyi:
The 437_tpc_gist_tbl expected output carried stale result counts. The sequential-scan and GiST-index counts agree at every query, so the index is consistent with the table scan and the drift was in the golden file rather than the index. Regenerate the expected output and remove the pointcloud SKIP_TESTS list.


estebanzimanyi:
Superseded by #1165, the temporal point cloud family consolidation (pcpoint/pcpatch base + tpcpoint/tpcpatch + vendored pgPointCloud, full regression green including table tests). Branch retained.

PR #1129, opened 2026-05-29
Treat implicit function declarations as build errors

estebanzimanyi:
Add -Werror=implicit-function-declaration and -Werror=implicit-int to the GCC flags. An implicit declaration silently assumes an int return, truncating any 64-bit pointer the callee returns and turning a missing header into a runtime crash instead of a compile-time failure; promoting it to an error catches that class of bug at build time across the MEOS kernel and all extension targets. The full tree, including the vendored liblwgeom, h3-pg and pgcommon sources, builds clean under the stricter flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,6 +205,10 @@ add_compile_definitions(_USE_MATH_DEFINES)
 if(CMAKE_COMPILER_IS_GNUCC)
   # MobilityDB compiler flags
   add_definitions(-Wall -Wextra -std=gnu1x -Wunused-parameter -Wno-strict-aliasing)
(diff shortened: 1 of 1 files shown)


mschoema:
I'm not specifically against this, but should we always have it on, or just during testing (maybe in our github actions for example, or in debug mode)?

estebanzimanyi:
> I'm not specifically against this, but should we always have it on, or just during testing (maybe in our github actions for example, or in debug mode)?

I have hit this problem so many times and spend so many hours debugging to find ot that this is simply a missing declaration that I would enforced it for all the ecosystem. A missing function or a missing parameter in MEOS is an error that cannot be propagated especially to the other tools of the ecosystem.

PR #1130, opened 2026-05-29
Make GEOS initialization thread-safe

estebanzimanyi:
Guard the single global GEOS context with a pthread_once initializer (meos_initialize_geos) and call it at the MEOS init path and before each GEOS-using geo entry point, replacing the per-operation initGEOS/finishGEOS pairs that tore the context down after every call. Concurrent MEOS worker threads no longer race on initGEOS, and repeated GEOS operations within a process reuse the live context. The vendored liblwgeom change is isolated in its own commit from the first-party MEOS wiring, and POSIX threads are linked via find_package(Threads) for the pthread_once guard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,11 @@ int main() {
 # Other dependencies
 #--------------------------------
 
+# POSIX threads — needed for pthread_once in meos_initialize_geos()
+# On Linux ≥ glibc 2.34 pthreads are in libc; on older systems and MinGW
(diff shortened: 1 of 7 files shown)


estebanzimanyi:
Superseded by #1166, which makes GEOS operations (not only initialization) thread-safe via a per-thread reentrant context. pthread_once global init only makes initialization safe; concurrent GEOS operations still share one global handle and race. Branch retained.

PR #1132, opened 2026-05-29
Add spatial and spatiotemporal tiling for temporal rigid geometries

estebanzimanyi:
spaceBoxes, timeBoxes and spaceTimeBoxes return the spatiotemporal boxes of a trgeometry split with respect to a spatial, temporal or spatiotemporal grid, computed on its centroid trajectory. The kernel functions trgeo_space_boxes and trgeo_space_time_boxes are part of the public MEOS surface so the whole ecosystem inherits them, reusing the temporal geometry tiling kernel. The MEOS kernel, MobilityDB SQL wrappers, the rigid-geometry manual chapter and pg_regress coverage travel together.


estebanzimanyi:
Subsumed by #1128.

PR #1134, opened 2026-05-29
Add motion metrics for temporal rigid geometries

estebanzimanyi:
length, cumulativeLength, speed and twCentroid report the motion of a trgeometry along its centroid trajectory. The kernel functions trgeo_length, trgeo_cumulative_length, trgeo_speed and trgeo_twcentroid are part of the public MEOS surface so the whole ecosystem inherits them, reusing the temporal point length, derivative and time-weighted centroid kernels. The MEOS kernel, MobilityDB SQL wrappers, the rigid-geometry manual chapter and pg_regress coverage travel together.


estebanzimanyi:
Subsumed by #1131.

PR #1138, opened 2026-05-30
Honor the meos_error return contract in the stbox_space_time_tile SRID check

estebanzimanyi:
In stbox_space_time_tile the equality of the point and space-origin SRIDs is enforced with ensure_same_srid, whose boolean result must be checked so that an embedder installing a non-exiting meos_error handler does not continue into tile construction with mismatched SRIDs. The check returns NULL on mismatch, matching the two sibling SRID guards already present in the same function.


meos/src/geo/tgeo_tile.c
mschoema, line 853: We should also use the same structure here:

```
if (! ensure_has_Z_geo(sorigin))
    return NULL;
```

mschoema (changes requested):

meos/src/geo/tgeo_tile.c
estebanzimanyi: Done — both ensure_has_Z_geo call sites now check the return value and return NULL on failure.

estebanzimanyi: Done — both call sites in `stbox_space_time_tile` and `tgeo_space_time_tiles` now check the return value and return `NULL` on failure.

PR #1143, opened 2026-05-30
Default the traversedArea unary union argument to false for temporal circular buffers

estebanzimanyi:
The traversedArea function for tcbuffer takes a Boolean argument that, when true, applies a spatial union to the per-segment geometries to return a single geometry. Its default of false matches the traversedArea functions for tgeometry and tgeography, which leave the per-segment geometries unmerged. The DocBook signature for the temporal circular buffers chapter records the argument and its default.


estebanzimanyi:

estebanzimanyi:

estebanzimanyi:

PR #1144, opened 2026-05-31
Vendor the Apache Arrow C Data Interface

estebanzimanyi:
Vendor the Apache Arrow C Data Interface ABI header under arrow/, with its static-analysis exclusion. Provenance import of the upstream Apache-2.0 header; reviewed against the upstream source, not line by line.


estebanzimanyi:

PR #1150, opened 2026-06-01
Return the within-distance interval for the temporal tDwithin of circular buffers

estebanzimanyi:
The temporal tDwithin of two circular buffers used the turning-point function that returns the instants at which the distance equals the bound, which the distance lifting needs to split the temporal distance. For tDwithin this made the result true only at the crossing instant rather than over the whole period during which the buffers are within the distance. Add a dedicated turning-point function that returns the within-distance sub-interval, derived from the crossings and the within status at the segment endpoints, and use it for tDwithin(tcbuffer, cbuffer) and tDwithin(tcbuffer, tcbuffer). The distance and intersection paths keep using the crossing turning point.


estebanzimanyi:

estebanzimanyi:

estebanzimanyi:

PR #1154, opened 2026-06-02
Free fresh by-reference results in tinstant_tagg

estebanzimanyi:
tinstant_tagg copies the per-pair aggregate value into the result instant. Fresh by-reference results are released afterwards; callbacks that return one of their inputs verbatim (such as datum_min_text) leave the pointer owned by the input instants, so the free is guarded against the two input values.

### DIFF
--- a/meos/src/temporal/temporal_aggfuncs.c
+++ b/meos/src/temporal/temporal_aggfuncs.c
@@ -399,10 +399,26 @@ tinstant_tagg(TInstant **instants1, int count1, TInstant **instants2,
     {
       if (func)
       {
-        result[count++] = tinstant_make(func(tinstant_value_p(inst1),
-          tinstant_value_p(inst2)), inst1->temptype, inst1->t);
+        Datum v1 = tinstant_value_p(inst1);
+        Datum v2 = tinstant_value_p(inst2);
+        Datum value = func(v1, v2);
+        result[count++] = tinstant_make(value, inst1->temptype, inst1->t);
         if (tofree)
           tofree1[nfree1++] = result[count - 1];
+        /* tinstant_make copies value into the result instant, so when func
+         * returns a freshly allocated by-reference value that allocation is
+         * orphaned and leaks unless it is freed here. The by-reference sum
+         * callbacks (e.g. datum_sum_double2, which calls double2_add) are this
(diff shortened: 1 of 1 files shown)


meos/src/temporal/temporal_aggfuncs.c
```diff
@@ -399,10 +399,17 @@ tinstant_tagg(TInstant **instants1, int count1, TInstant **instants2,
...
+        result[count++] = tinstant_make(value, inst1->temptype, inst1->t);
         if (tofree)
           tofree1[nfree1++] = result[count - 1];
+        /* Free freshly-allocated by-reference results. Callbacks such as
+         * datum_min_text return one of their inputs verbatim; those pointers
+         * are still owned by the input instants and must not be freed here. */
+        if (! basetype_byvalue(temptype_basetype(inst1->temptype)) &&
+            DatumGetPointer(value) != DatumGetPointer(tinstant_value_p(inst1)) &&
+            DatumGetPointer(value) != DatumGetPointer(tinstant_value_p(inst2)))
+          pfree(DatumGetPointer(value));
```
mschoema: Maybe take out the `tinstant_value_p(inst1)` and `tinstant_value_p(inst2)` calls so they are not done twice (in the `func` call, and in the `if` below).

mschoema (commented):
Is this the only function that has this problem? If not, we should probably all fix them in the same PR.

meos/src/temporal/temporal_aggfuncs.c
estebanzimanyi: Done — v1 and v2 are now pre-computed before the func call and reused in the equality checks.

estebanzimanyi: Done — `v1` and `v2` are now pre-computed before the `func` call and reused in the pointer comparison below.

estebanzimanyi:
The double-call pattern that needs a fix occurs only in this loop: `func()` uses the pointer and the `DatumGetPointer` comparison below it reuses the same pointer without pre-computing. The other `tinstant_value_p` calls in the file (line 562 of the sync-merge loop, and lines 1382/1405 in the `double2` path) each use the result only once, so there is no redundant evaluation there.

PR #1156, opened 2026-06-03
Return the instant count from tsequence_insts_p and tsequenceset_insts_p

estebanzimanyi:
The internal accessors tsequence_insts_p and tsequenceset_insts_p write the number of instants of the returned pointer array through an output parameter, letting callers obtain the array and its length in a single call.


estebanzimanyi:
The tsequence_insts_p/tsequenceset_insts_p count out-parameter is carried by #1157: the temporal circular buffer is its only consumer among the type families, so the accessor change rides with that family rather than as a standalone core change.

PR #1162, opened 2026-06-03
Complete the temporal geo spatial relationship, distance, analytics, and tiling surface

estebanzimanyi:
Add the temporal geo family surface for tgeompoint, tgeogpoint, tgeometry, and tgeography: geodetic-consistent ever/always and temporal spatial relationships, distance and nearest-approach, analytics (similarity and simplification), restrictions, and spatiotemporal tiling. The spatial accessors return their results through value structs (MvtGeom, SpaceSplit, SpaceTimeSplit) and the set and cluster accessors return their element counts through an output parameter. The tnumberinst_distance prototype is declared in meos_internal.h. Stacks on the lifted ever/temporal relationship fix in fix/eintersects-tcontseq-crossing-value; that branch's lifting.c change appears here until it lands.


estebanzimanyi:
Rebased onto current master to resolve the conflict in `meos/src/geo/tgeo_tile.c`. The struct-return signatures from this branch are preserved; the `@csqlfn` tags added by #1190 are picked up from master.

PR #1167, opened 2026-06-04
Implement the temporal distance between a temporal rigid geometry and a temporal point

estebanzimanyi:
tdistance_trgeometry_tpoint reduces to the existing static-target rigid-geometry distance solver by working in the moving point's translating frame: the inputs are synchronized, an adjusted temporal pose is built whose translation is the rigid geometry's translation minus the point's position with the rotation unchanged, and the result is the distance of that relative rigid geometry to the origin. The reduction is exact (validated against a densified oracle to 1.78e-15 at the operator's breakpoints), and it makes nad_trgeometry_tpoint, the nearest-approach instant, and the reversed-argument form return exact results. The change also corrects the missing denominator parentheses in the V-clip closest-feature solver, the interpolation-type argument and posed-geometry materialization in the rigid-geometry distance workers, and the operand passed to the nad_trgeometry_tpoint validity check.


estebanzimanyi:
The temporal rigid geometry distance surface, including tdistance(trgeometry, tgeompoint) and the V-clip closest-feature solver fix, is delivered by the rgeo family PR #1161 with the full 161_trgeo_distance and 161_trgeo_vclip test coverage.

PR #1170, opened 2026-06-04
Seed the random generator once so the test data generation is reproducible

estebanzimanyi:
Set a fixed seed at the top of the data generation extension so that every create_test_tables_* call in the same session produces reproducible tables for a given PostgreSQL version.

### DIFF
--- a/mobilitydb/datagen/header.in.sql
+++ b/mobilitydb/datagen/header.in.sql
@@ -38,3 +38,7 @@
 
 -- complain if script is sourced in psql, rather than via CREATE EXTENSION
 \echo Use "CREATE EXTENSION mobilitydb_datagen" to load this file. \quit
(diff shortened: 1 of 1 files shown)


mschoema:
I don't think this should be added. The `CREATE EXTENSION` command doesnt actually create any tables, it just defines the functions. For reproducibility, the user should just run `setseed` as needed before the `create_test_tables_ *` calls. This should be the user's responsibility, I don't think we should add `setseed` commands in the extension itself.

estebanzimanyi:
Agreed. The CI test suite loads committed xz pg_dumps, so the datagen extension is not in the reproducibility-critical path. Closing this PR; users who want reproducible table generation can call setseed before the create_test_tables_* functions.

PR #1185, opened 2026-06-05
Disable floating-point contraction for cross-platform deterministic results

estebanzimanyi:
Geometric computations produce identical floating-point results across compilers and optimization levels by disabling fused multiply-add contraction.


mschoema:
Not sure I like this. Most other projects don't do it. And FMA is actually an optimization (both in performance and accuracy), so it's probably not ideal to disable it when its available.

estebanzimanyi:
You are right. The trig functions applied here (sin/cos/tan per instant, no extrema detection) do not involve any fused multiply-add patterns, so a global `-ffp-contract=off` flag would not change their output and is an unnecessary hammer on the whole build. Closing this PR.

PR #1189, opened 2026-06-06
Provide a single temporal-first instantiation for the commutative bbox predicates

estebanzimanyi:
The topological predicates same, overlaps and adjacent between a temporal value and a bounding box (tbox, stbox) or a span (numeric span, tstzspan) are commutative, so one temporal-first typed function covers both argument orders: same_tspatial_stbox, same_tnumber_tbox, same_tnumber_numspan, same_temporal_tstzspan and the overlaps and adjacent counterparts. The PostgreSQL operators ~=, && and -|- expose both argument directions through the generic Boxop dispatchers together with the operator COMMUTATOR, which is what GiST and SP-GiST index access uses; these typed functions are MEOS binding API outside the index path, so index access on either side is unaffected.

### DIFF
--- a/meos/include/meos.h
+++ b/meos/include/meos.h
@@ -1628,14 +1628,11 @@ extern TBox *tnumber_tboxes(const Temporal *temp, int *count);
 
 /* Topological functions for temporal types */
 
-extern bool adjacent_numspan_tnumber(const Span *s, const Temporal *temp);
-extern bool adjacent_tbox_tnumber(const TBox *box, const Temporal *temp);
 extern bool adjacent_temporal_temporal(const Temporal *temp1, const Temporal *temp2);
 extern bool adjacent_temporal_tstzspan(const Temporal *temp, const Span *s);
 extern bool adjacent_tnumber_numspan(const Temporal *temp, const Span *s);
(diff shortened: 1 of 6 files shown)


estebanzimanyi:
Reverting per ecosystem decision. This symmetric-operator dedup removes the 12 commutative-reverse C wrappers (overlaps_/same_/adjacent_ × {tstzspan_temporal, stbox_tspatial, numspan_tnumber, tbox_tnumber}). Sound for PostgreSQL (operators use commutators → one C direction suffices), but it BREAKS function-based bindings (MobilityDuck and every non-PG binding) which must register both calling orders as distinct functions — e.g. overlaps(stbox,tgeompoint) AND overlaps(tgeompoint,stbox) — so the reverse wrapper is a referenced symbol; removing it makes MobilityDuck's 15g build fail (undeclared symbol).

Per prefer-proper-meos-export / close-gaps-in-meos-c-not-sql-composition, the 12 thin 'return forward(b,a)' reverse shims belong in MEOS-C (the single source of truth) so every binding gets them for free. master already carries all 12; closing this keeps them, and the next pin is re-derived without this PR — restoring them with no binding-side change.

PR #1193, opened 2026-06-06
Support geodetic poses and temporal poses

estebanzimanyi:
A pose and a temporal pose carry a geodetic flag like every spatial type, with the planar and geodetic subtypes distinguished by a flag exactly as the stbox type distinguishes STBOX and GEODSTBOX. The pose constructors take the flag, pose_make_point sources it from the position point geometry or geography nature, the GEODPOSE and GeodPose keyword marks it in the canonical and WKT representations, and the WKB flags byte carries it, so planar and geodetic poses and temporal poses round-trip through text and binary.


estebanzimanyi:
Folded into #1160, the single independent temporal-pose PR (base + geodetic + temporal-point conversion).

PR #1194, opened 2026-06-06
Convert a geodetic temporal pose to a temporal geography point

estebanzimanyi:
The conversion of a temporal pose to a temporal point yields a temporal geometry point for a planar pose and a temporal geography point for a geodetic pose, with the result type derived from the pose geodetic flag. A single Tpose_to_tpoint function backs both the tgeompoint(tpose) and tgeogpoint(tpose) casts. Stacks on the geodetic pose support (#1193).


estebanzimanyi:
Folded into #1160, the single independent temporal-pose PR (base + geodetic + temporal-point conversion).

PR #1195, opened 2026-06-07
Index the hex-decode table with the unsigned byte value

estebanzimanyi:
bytes_from_hexbytes indexes the 256-entry hex2char lookup with a hex character cast to int from a plain char. On platforms where char is signed a byte with the high bit set casts to a negative index and reads out of bounds, so a hex-WKB string carrying such a byte is decoded through undefined behavior instead of being rejected cleanly. Cast the character to uint8_t so the index is the unsigned byte value on every platform.


mschoema (approved):
Fine with me, but maybe worth trying to send the same fix to PostGIS too.

PR #1203, opened 2026-06-13
Expose base-type operations in the public MEOS API

estebanzimanyi:
## Summary

The public MEOS API must provide **all operations over the base types**, so that every binding generated from the MEOS catalog (JMEOS, PyMEOS, MEOS.NET, …) can reach them under a clean standalone name. This PR declares the comparison / arithmetic / math / conversion operations on `int`, `float8`, `Interval`, `DateADT` and `TimestampTz` whose implementations already live in the vendored `pgtypes` library, and adds the one implementation that was missing.

Functions exposed:

| Operation | Functions |
|---|---|
| comparison | `int32_cmp`, `int64_cmp` |
| math | `float8_exp`, `float8_ln`, `float8_log10` |
| interval arithmetic | `add_interval_interval`, `mul_interval_double` |
| date / timestamptz arithmetic | `minus_date_date`, `minus_date_int`, `minus_timestamptz_interval`, `minus_timestamptz_timestamptz` |
| conversion | `date_to_timestamp`, `date_to_timestamptz` |

`mul_interval_double` was the only function without an implementation; its body (PG 17.2 `interval_mul`) is added to `postgres_types.c`. The other twelve are already implemented in `pgtypes/` and only needed a public declaration.

## Changes
- `meos/include/postgres_ext_defs.in.h` — 13 public declarations.
- `meos/src/temporal/postgres_types.c` — `mul_interval_double` implementation + `utils/timestamp_def.h` include (for `DAYS_PER_MONTH`).

## Verification
`libmeos.so` builds clean and `nm -D` exports all 13 symbols.

> **Stacked on #751** (`pg18_types_linked`): the declarations rely on the `pgtypes` library introduced there. Review after #751.


estebanzimanyi:
Note on CI: the single failing **Build** job (`Compile and run meos/test`, ~1 min) fails with `cc1: fatal error: bool_test.c: No such file or directory` — `.github/workflows/meos.yml` compiles `bool_test.c`, which does not exist on this branch. This is **inherited from the #751 base**, where the identical job fails the same way; this PR does not touch `meos/test`. The build matrix (PG 16/17/18 on ubuntu-24.04 + macOS) all pass, and `nm -D libmeos.so` exports all thirteen new symbols. The red resolves when #751 adds `bool_test.c` / merges.

estebanzimanyi:
Subsumed by current master. The base-type operations this exposes are already public on master via the installed `pg_int.h` / `pg_float.h` / `pg_interval.h` / `pg_date.h` / `pg_timestamp.h` headers (plus `postgres_ext_defs`), and master already carries the `postgres/`→`pgtypes/` vendoring this branch proposed (it has since diverged, so the diff would revert that work). The one genuine gap — `mul_interval_double` (master has `div_interval_float8` but no multiply) — is being added as a small focused PR. Closing.

estebanzimanyi:
Correction: the interval multiply is in fact already present on master as `mul_interval_float8(interval, float8)` and `mul_float8_interval(float8, interval)` (`pgtypes/utils/timestamp.c:3461`/`:3573`, public in `pgtypes/pg_interval.h:76-77`, covered in `meos/test/interval_test.c`). So there is no remaining gap — this PR is fully subsumed by master. No follow-up needed.

PR #1204, opened 2026-06-14
Add the portable aliases for the H3 family

estebanzimanyi:
Gives every H3 operator its catalog-portable bare/camelCase function name, the same way the other type families get theirs from the portable-alias generator (#1075). In particular the `?=` prefilter operator gets **`everEq(h3indexset, th3index)`** (plus the `th3index`/`h3index` comparison forms), which the canonical BerlinMOD suite uses as the index-less spatial prefilter on DuckDB/Spark — so the suite names it without any hand alias.

- Registers the `h3` group in `tools/portable_aliases/generate.py` (`GROUP_PREFIX`); its `298` prefix places the file after the H3 operator defs.
- Commits the generated `mobilitydb/sql/h3/298_portable_aliases.in.sql` (pure generator output).
- `h3/CMakeLists.txt` does **not** hand-list the file — it is auto-registered by the `append_portable_aliases()` macro #1075 adds, so onboarding a family needs no manifest edit.

**Stacks on #1164** (the H3 type family) **and #1075** (the alias generator + auto-include macro). Until those land, their commits are visible in this diff; the net-new content is the `GROUP_PREFIX` line, the `298` file, and the `append_portable_aliases()` call. Merge #1075 and #1164 first, then this.


estebanzimanyi:
Folded into #1164 (clean/h3). The H3 portable bare-name aliases (now generated into `mobilitydb/sql/h3/298_portable_aliases.in.sql` via `tools/portable_aliases/generate.py`), the canonical operator dialect, the `@sqlfn` empty-parens form, and the `@csqlfn` chain for `geoToH3Cell` are all carried in #1164 directly. Verified H3-only build + full pg_regress green (169/169). Closing so each family PR is self-contained for the pin fold.

PR #1207, opened 2026-06-14
Normalize the remaining CRLF source files to LF

estebanzimanyi:
#1184 standardized the MEOS/MobilityDB sources on LF line endings but missed a handful of files. This converts the stragglers to LF:

- the doxygen group headers for every type family (cbuffer/geo/npoint/pose/rgeo/temporal, both public and internal),
- `meos/src/geo/tpoint_geom_clip.c`,
- the `npoint` and `setspan` MEOS tests,
- the geo datagen script.

**Line endings only** — every file is byte-identical modulo the stripped `\r` (verified). No code or doc content changes.


estebanzimanyi:
Content landed in master via #1206 (this was its stacked base; merging #1206 brought the CRLF→LF normalization). Closing as redundant.

PR #1209, opened 2026-06-14
Make the vendored pgtypes MEOS includes resolvable in the standalone build

estebanzimanyi:
The pgtypes sources include the MEOS public headers (`meos.h`, `meos_error.h`, `meos_tls.h`) via a hardcoded depth-2 relative path `../../meos/include/<h>`. The pgtypes sources live at depths 1..4 under `pgtypes/`, so that path only resolves for the depth-2 files; the **standalone pgtypes build** (the minimal build used by MobilityDuck, MobilitySpark/JMEOS and MobilityNebula — which does not leak `meos/include` onto the pgtypes compile the way the full MEOS/extension builds do) breaks on the others:

```
pgtypes/postgres.h:74: fatal error: meos_error.h: No such file or directory
```

**Fix:** expose `${CMAKE_SOURCE_DIR}/meos/include` on the pgtypes target's include directories (depth-independent) and drop the fragile `../../` prefix. The PG extension is unaffected (it uses the real PostgreSQL headers, not the vendored `pgtypes/`).

Stacks on #751 (which introduces `pgtypes/`). Verified: clean-room compile of a depth-3 source fails without the include dir and resolves with it; standalone `-DMEOS=ON` and the PG extension both build.


estebanzimanyi:
Absorbed into #751: this vendored-pgtypes standalone include fix targets #751's own pgtypes sources and cannot precede #751, so it now lands as part of #751 (which builds standalone MEOS green with it folded in). Closing as absorbed.

PR #1210, opened 2026-06-14
Add the temporal CARTO QUADBIN cell-index type (tquadbin)

estebanzimanyi:
Adds the static `quadbin`/`quadbinset` base types and the temporal **`tquadbin`** CARTO QUADBIN cell index — a `uint64` carried as `int8` (on-disk identical to `tbigint`), with input/output, accessors, operators, the geo cell-index integration, conversions to/from `tbigint` and `tgeompoint`, SQL, regress tests (295–305, all green), docs, and a MEOS smoke example.

It also exposes `ensure_temporal_isof_type()` (moves it out of the `#if MEOS` guard) so the `tbigint`↔`tquadbin` conversion validators link in the PG extension — the same way the sibling `th3index` cell index uses it.

**Stacked on #1163** (the `tbigint` base type it converts to and shares storage with) — the diff shows #1163's commits until it merges; the net-new content is the quadbin family.

Extracted as a clean topical PR from the integration branch (the prior fork-internal #26 is superseded). Verified locally: builds green with `-DQUADBIN=ON`, 100% of the quadbin regress suite passes, catalog/tbigint/geo regression is unaffected, cppcheck (project-mode) clean.


estebanzimanyi:
Heads-up on the MeosType enum slots: to stay collision-free against the other in-flight type PRs, quadbin now reserves its ecosystem-wide final values — **T_QUADBIN=78, T_QUADBINSET=79, T_TQUADBIN=80** — sitting after tbigint(63), th3index(64-66), pgpointcloud(67-73), tjsonb(74-77). On this PR's #1163 base the 64-77 slots are simply unused (those types ship on their own PRs); the enum already carries reserved gaps and the Oid cache auto-sizes from NUM_MEOS_TYPES, so the build/tests are unaffected (100% quadbin suite green).

PR #1212, opened 2026-06-14
Validate the WKB type code in the span/set/spanset/temporal readers

estebanzimanyi:
**Crash / robustness defect.** The WKB readers (`span_from_wkb_state`, `set_from_wkb_state`, `spanset_from_wkb_state`, `temporal_from_wkb_state` in `type_in.c`) read a 2-byte type code and **trusted it** — assigning it into the parse state and looking up its base type. A valid-hex buffer of a *different* family (e.g. feeding a `tstzspan` hex to `temporal_from_hexwkb`) was parsed as the wrong structure, reading over the buffer and **SIGSEGV**-ing the host process instead of returning NULL.

Repro (confirmed via JMEOS + the Spark BerlinMOD probe):
```
p    = temporal_from_hexwkb("…")        // a tint
hex  = span_as_hexwkb(temporal_to_tstzspan(p), …)   // a tstzspan WKB
temporal_from_hexwkb(hex)               // ← SIGSEGV, not NULL
```
Real trigger: the canonical BerlinMOD NxN — `overlaps(timeSpan(t1.trip), timeSpan(t2.trip))` hands a `tstzspan` WKB to the temporal reader.

**Fix:** one guard per reader — after reading the code, verify it belongs to the expected family via the existing `span_type`/`set_type`/`spanset_type`/`temporal_type` classifiers; otherwise raise `MEOS_ERR_WKB_INPUT` and return NULL (graceful null under the no-exit handler) **before** any base-type lookup or structure parse. Fixes it at the source so every untyped-hex consumer (Spark/Duck/PyMEOS/Nebula) is protected without a per-binding workaround. Complements #1172 (the hex-WKB type *discriminator*); this is the defensive read path.

### DIFF
--- a/meos/src/temporal/type_in.c
+++ b/meos/src/temporal/type_in.c
@@ -2083,7 +2083,18 @@ type_from_wkb(const uint8_t *wkb, size_t size, MeosType type)
   if (span_type(type))
   {
     Span *span = palloc(sizeof(Span));
+    /* Unlike the pointer-returning readers, span_from_wkb_state returns a
+     * Span by value, so on a foreign type code it can only return a zeroed
+     * struct (non-NULL once wrapped). Honour the same NULL contract as the
+     * other families: reset the error state, parse, and fail if the reader
+     * raised an error (the type-code guard in span_from_wkb_state). */
+    meos_errno_reset();
     *span = span_from_wkb_state(&s);
+    if (meos_errno() != 0)
+    {
+      pfree(span);
+      return (Datum) 0;
+    }
     return PointerGetDatum(span);
   }
   if (spanset_type(type))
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Completed the span family (relayed gap). `span_from_wkb_state` returns a `Span` by value, so on a rejected type code it could only yield a zeroed struct that `span_from_hexwkb`/`span_from_wkb` wrapped into a **non-NULL** pointer — unlike the set/spanset/temporal readers which return NULL. Added commit `Honour the NULL contract in the span hex-WKB reader`: the span branch of `type_from_wkb` now resets the error state, parses, and returns NULL when the type-code guard fired.

Runtime-verified (conforming noexit handler): `span_from_hexwkb(tint_hex)` → NULL (errno=24 MEOS_ERR_WKB_INPUT), while a valid intspan still round-trips equal. This lets type-erased hex-WKB callers (Spark/Duck/PyMEOS/Nebula) drop their per-family type-byte wrappers.

PR #1214, opened 2026-06-15
Add pin_preflight.sh as a repo tool with full-tree/test, pointcloud, and dup-main guards

estebanzimanyi:
`tools/pin_preflight.sh` was **pin-only** (the one artifact not derived from a PR). This promotes it to a tracked repo tool and closes the two gate gaps that let real breaks ship green:

- **Full MEOS tree incl. test executables** (`-DBUILD_TESTING=ON`, default target). The existing MEOS gates build `--target meos -DBUILD_TESTING=OFF` = library only, so `meos/test/*` is never compiled — that is exactly why #803's duplicate-`main` `npoint_test.c` passed preflight yet reddened every binding CI in pins 15a/15c/15d.
- **Static `one main() per meos test`** ratchet for the same class (fast, runs in `--static-only`).
- **POINTCLOUD build leg** (builds `libpc.a` first) so the pointcloud surface — incl. the tpcpoint W124 funcs — is actually gated; it was a false-green at 15c. Loud SKIP on envs without the pgPointCloud toolchain.

No functional code change; tooling only. `bash -n` clean.

### DIFF
--- a/tools/pin_preflight.sh
+++ b/tools/pin_preflight.sh
@@ -1,0 +1,184 @@
+#!/usr/bin/env bash
+# Pin preflight gate — run BEFORE pushing evidence/assembly-s5 or tagging a pin.
+#
+# Mirrors the locally-reproducible MobilityDB CI workflows so a pin is never
+# published red. Each block maps to a .github/workflows/*.yml job. Exit code is
+# the count of failed gates; 0 = green-to-publish.
+#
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded by #1233, which consolidates the pin derivation toolkit and `pin_preflight.sh` into a single reviewable PR (`tools/pin` + `tools/pin_preflight.sh`), rebased onto current master.

PR #1216, opened 2026-06-16
Add RASTER=ON optional family for PostGIS raster band sampling along trajectories

estebanzimanyi:
Introduces \`-DRASTER=ON\` as a new optional build family (parallel to CBUFFER, POINTCLOUD, POSE, RGEO) that adds \`raster_value(raster, tgeompoint, band integer DEFAULT 1) → tfloat\` sampling a PostGIS raster band at every instant of a tgeompoint trajectory. Instants outside the raster extent or on nodata pixels are silently dropped; NULL is returned when no instants survive.

\`postgis_raster\` ships with the standard \`postgresql-N-postgis-3\` apt package; no additional dependency is required. When RASTER=ON the generated \`mobilitydb.control\` declares \`requires = 'postgis, postgis_raster'\` so a single \`CREATE EXTENSION mobilitydb CASCADE\` loads the full stack.

Includes the CMake gate (option/add_definitions/MOBILITYDB_REQUIRES), the SQL function, pg_regress tests with a synthetic 3×3 raster (4 cases: all inside, one outside, all outside, single instant), CI coverage flag (\`-DRASTER=ON\` alongside CBUFFER/POSE/RGEO), and a DocBook XML chapter documenting the operator and production roadmap.


estebanzimanyi:
Consolidated into #1218, which now carries the full RASTER family (build gate, MEOS kernel, and operators) as clean topical commits rebased on current master.

PR #1217, opened 2026-06-16
Add MEOS kernel for Raquet raster chip sampling via QUADBIN (RASTER=ON)

estebanzimanyi:
Adds `raster_tile_value_quadbin()` and `trajectory_quadbins()` to the MEOS public surface and the MobilityDB SQL extension, enabling cloud-native raster sampling along `tgeompoint` trajectories against Raquet tiles without requiring the PostGIS raster type at the kernel layer.

The QUADBIN Morton encode/decode follows the canonical CARTO quadbin-js algorithm (`spread_bits` / `compact_bits` with MAX_ZOOM=26, shifts 16→8→4→2→1), self-contained in `meos/src/raster/raster_quadbin.c` with no dependency on the separate tquadbin PR. `trajectory_quadbins(traj, zoom)` returns the distinct QUADBIN cells covered by a trajectory — the WHERE-clause join key for a Raquet table. `raster_tile_value_quadbin(pixels, width, height, quadbin, pixtype, nodata, has_nodata, traj)` samples a row-major pixel array (UINT8/INT16/INT32/FLOAT32/FLOAT64) at each trajectory instant using the Mercator inverse formula for accurate row mapping, silently dropping out-of-tile or nodata instants. `MeosPixType` is declared in `meos.h` under `#if RASTER`. pg_regress coverage exercises all three functions including the zoom-range error and a four-instant drop-one case.


estebanzimanyi:
Consolidated into #1218, which now carries the full RASTER family (build gate, MEOS kernel, and operators) as clean topical commits rebased on current master.

PR #1219, opened 2026-06-16
Harden tquadbin for the pin: basetype_byvalue merge-safety + tcellindex NULL guards

estebanzimanyi:
Stacked on #1210 (feat/tquadbin, open).

`basetype_byvalue` in #1210 re-wrapped the base-type terms (line break after `T_INT8`), while `master` and #1164 (clean/h3) break after `T_FLOAT8`. That whitespace divergence makes the `#if QUADBIN` and `#if H3` additions textually conflict when both features compose; in the ecosystem pin the conflict was mis-resolved into a **dead second `return`** that dropped QUADBIN from the reachable branch, so `basetype_byvalue(T_QUADBIN)` returned false and `meostype_length` raised `Unknown base type: quadbin` across the whole quadbin test family.

This re-wraps `basetype_byvalue` to match `master`, so each feature contributes only its own `#if` block and H3+QUADBIN combine with no conflict and no hand-resolution.


estebanzimanyi:
Subsumed into #1210 (feat/tquadbin, head cb796b015).

The two substantive changes here are already in #1210:
- **tcellindex NULL guards** (`d2e4ca7ea`): #1210's `meos/src/temporal/tcellindex.c` already has the 6 `if (! ops) return NULL;` guards at the same entry points (verified identical guard count).
- **basetype_byvalue merge-safety** (`d0e310d3e`): #1210's `meos_catalog.c` is already merge-safe (single reachable return, `T_QUADBIN` in a leading-`||` `#if QUADBIN` branch); the line-wrap alignment with master is folded in.

The remaining diffs in this branch (older `postgres_types.h`/`pg_hashint8`/`cstring2text` includes, internal `VALIDATE_TQUADBIN`) point the wrong way relative to #1210's modern `<pgtypes.h>` form and would regress it, so they are intentionally not carried. Closing to keep the pin fold clean.

estebanzimanyi:
Subsumed by #1210: the NULL DggsCellOps guards in tcellindex.c and the byvalue merge-safety are already present in the complete tquadbin PR. Closing as folded.

PR #1223, opened 2026-06-17
Reconcile h3/quadbin/pointcloud libmeos family link onto the #751 pgtypes foundation

estebanzimanyi:
**Stacked on #751** (`pg18_types_linked`, open) and the family PRs #1163 (tbigint) / #1210 (quadbin) / #1164 (h3) / #1165 (pointcloud) — their commits ride along and dedupe on their own merges, leaving this reconciliation delta.

### The defect (ecosystem-fatal, runtime-only)
The standalone `libmeos.so` that every binding loads (Spark/Flink/Kafka/Nebula/Duck/PyMEOS) **fails to load**: `undefined symbol: quadbin_cellops` (+ `h3index_in/out`, `th3index_*`, `tquadbin_*`, and the pointcloud family). Invisible to header/catalog/compile/per-PR CI — only **building + loading the all-families `.so`** exposes it.

### Root cause
#751 renamed the libmeos link list `PROJECT_OBJECTS` → `MEOS_OBJECTS` (and split SHARED/STATIC). The base families were updated; **h3/quadbin/pointcloud come from pre-#751 family PRs and still append to `PROJECT_OBJECTS`**, which the post-#751 lib never links. The mismatch exists only where #751 meets the families.

### The fix (this PR)
Each family's append becomes the sibling-mirror `set(MEOS_OBJECTS ${MEOS_OBJECTS} "$<TARGET_OBJECTS:fam>")`, plus the #751 pgtypes-port adaptations each family needs on the #751 base: `<pgtypes.h>` includes; `int8_out→int64_out` / `pg_hashint8→int64_hash` / `cstring2text→cstring_to_text` to #751's vendored names; `<utils/timestamp.h>` for `TimestampTzGetDatum`.

### Verified (this tree)
- libmeos `-DMEOS=ON` with all 7 families builds, and the `.so` **LOADS** (`meos_initialize OK`) — `quadbin_cellops`/`h3index_in`/`th3index_in`/`tpcpoint_in`/`tpcpatch_in` all defined, **zero undefined family symbols**.
- PG extension `-DMEOS=OFF` (all families) builds clean — **both targets**.


estebanzimanyi:
Superseded by the per-family rebases. The libmeos family-link reconciliation this PR carried is now materialized in each family's own PR:

- The `PROJECT_OBJECTS`→`MEOS_OBJECTS` libmeos link flip is applied per-family in #1163 (tbigint), #1164 (h3), #1210 (quadbin), and #1165 (pointcloud).
- The `basetype_byvalue`/catalog single-return fix is contributed per-family (each adds its own `#if FAMILY` term with the leading-`||` form) and combines correctly when the families compose.
- The `branch_ignore`→`branches-ignore` workflow fix is already on master (#1010).
- The win32 pgtypes build bits belong to the native-Windows track (#959 / #1105), not the family-link reconciliation.

This PR was also never a publishable superset on its own (it dropped the jsonb types and the extended binding externs). Closing in favor of the independent family PRs, which is cleaner for the automatic pin fold.

PR #1229, opened 2026-06-18
Export the pointcloud catalog type predicates instead of defining them inline

estebanzimanyi:
### Problem

`pointcloud_basetype`, `pointcloudset_type` and `tpointcloud_temptype` are
declared `extern` in the public `meos_catalog.h` but defined with the `inline`
specifier. Under C99 inline semantics a plain `inline` definition emits no
external symbol, so `libmeos.so` exports nothing for them and a consumer linking
against the shared library gets an undefined reference — and `pointcloudset_type`
/ `tpointcloud_temptype` are used in real (non-assertion) code paths
(`set.c`, `temporal_boxops.c`).

### Fix

Drop the `inline` specifier so each pointcloud predicate emits an exported
symbol matching its `extern` declaration, consistent with the exported
`geo_basetype` / `spatial_basetype` accessors. This makes the pointcloud
catalog-introspection surface linkable for downstream bindings, uniform with
every other type family — no type is excluded from the public surface.

Stacked on #1165 (which introduces these predicates).


estebanzimanyi:
Subsumed into #1165 (clean/pointcloud, head cef7edfbd).

The substantive fix here — dropping the `inline` specifier on the three pointcloud catalog predicates (`pointcloud_basetype`, `pointcloudset_type`, `tpointcloud_temptype`) so each emits an exported symbol matching its `extern` declaration in the public `meos_catalog.h` — is now folded directly into #1165 as its own commit, with the same C99-inline rationale.

This consolidation lets the pin fold #1165 without a separate satellite fix PR. Closing.

estebanzimanyi:
Subsumed by #1165: the pointcloud catalog type predicates (pointcloud_basetype/pointcloudset_type/tpointcloud_temptype) are already exported (declared extern in meos/include/temporal/meos_catalog.h) in the complete pointcloud PR. Closing as folded.

PR #1242, opened 2026-06-20
Fix memory-safety bugs in the skiplist splice path

estebanzimanyi:
Carved from #992 (the MEOS bug-audit batch) as a single-topic memory-safety fix.

Three related bugs in the skiplist splice path:

1. **`keyval_skiplist_common`** — `memset(update, 0, sizeof(&update))` zeroes only a pointer's worth (8 bytes) instead of the full `update` array (`SKIPLIST_MAXLEVEL` ints), leaving most levels uninitialised. Fixed to `sizeof(int) * SKIPLIST_MAXLEVEL`.
2. **`skiplist_splice`** — the per-element key is `palloc`'d and `memcpy`'d but never assigned to `newelem->key`, leaking the copy and leaving the element key unset. Fixed by assigning it.
3. **`skiplist_splice`** — the merge-helper pointer-array shells are never freed; released after the insertion loop. They are distinct allocations from `tofree`, so this does not double-free the merged elements.

All three verified still-present on master; strict build green (3 targets, cppcheck-clean, warning-clean). Single file.

### DIFF
--- a/meos/src/temporal/skiplist.c
+++ b/meos/src/temporal/skiplist.c
@@ -441,7 +441,7 @@ keyval_skiplist_common(SkipList *list, void **keys, void **values, int count,
   void *max_value = values[count - 1];
 
   /* Find the list values that are strictly before the span of new values */
-  memset(update, 0, sizeof(&update));
+  memset(update, 0, sizeof(int) * SKIPLIST_MAXLEVEL);
   int height = list->elems[0].height;
   SkipListElem *elem = &list->elems[0];
   int cur = 0;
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Closed: this introduces a **deterministic Linux crash** — `geo_test` aborts (SIGABRT, exit 134) on all four Linux builds while all macOS builds pass (the glibc-vs-macOS double-free / heap-corruption signature). The only delta vs the green master is `skiplist.c`, and these changes touch the keyval skiplist path documented as unvalidated with known correctness bugs, so the shell-`pfree` / `newelem->key` ownership assumptions are likely wrong. The `memset` under-initialisation part is a genuine fix and will be re-submitted alone; the keyval-ownership changes will be revisited only with valgrind/ASan + the test suite, not a build-only check.

PR #1252, opened 2026-06-21
Align the tcbuffer NAD and distance failure sentinel with DBL_MAX

estebanzimanyi:
The temporal circular buffer nearest-approach-distance functions (`nad_cbuffer_stbox`, `nad_tcbuffer_geo`, `nad_tcbuffer_stbox`, `nad_tcbuffer_cbuffer`, `nad_tcbuffer_tcbuffer`) returned `-1.0` on invalid/empty/no-overlap inputs, whereas the canonical failure sentinel across the distance surface is `DBL_MAX` (the sibling `nad_tgeo_*` functions all return `DBL_MAX`).

A `-1.0` sentinel sorts smaller than any real distance, so a KNN search using the `|=|` operator over a GiST/SP-GiST index returns the can't-compute entries first and reports `-1` instead of the true nearest distance; `DBL_MAX` sorts them last so the genuinely nearest rows are returned. The PostgreSQL wrappers already test `if (result == DBL_MAX) PG_RETURN_NULL();`, so the MEOS layer returning `-1.0` left that conversion as dead code.

Return `DBL_MAX` from these paths, matching the temporal geometry family. The base `distance_cbuffer_*` functions are unchanged.

Verified: full cbuffer regression suite green (incl. `166_tcbuffer_indexes_tbl` and `210_tcbuffer_distance`).

### DIFF
--- a/meos/src/cbuffer/tcbuffer_distance.c
+++ b/meos/src/cbuffer/tcbuffer_distance.c
@@ -32,6 +32,8 @@
  * @brief Temporal distance for temporal circular buffers
  */
 
+/* C */
+#include <float.h>
 /* MEOS */
 #include <meos.h>
 #include <meos_internal.h>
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded: the tcbuffer NAD/distance `DBL_MAX` sentinel is already materialized in #1157 (clean/cbuffer) — its `tcbuffer_distance.c` returns `DBL_MAX` at all `nad_*` sites (0 remaining `-1.0`). With the family PRs now using family-only coverage (h3/pointcloud no longer run cbuffer tests), this standalone extraction is unnecessary and only duplicates content for the pin fold. Closing in favor of the fix living in cbuffer's own PR.

PR #1270, opened 2026-06-29
Use SQL lookup for MF-JSON SRS names

Komzpa:
## Summary

- replace the direct `getSRSbySRID` call in `Temporal_as_mfjson` with a MobilityDB-owned SPI lookup against PostGIS `spatial_ref_sys`
- cache the last SRID/CRS-mode result in the `asMFJSON` function `fn_extra` context
- bind the SRID lookup as a typed SPI argument instead of interpolating it into the query text
- remove the now-unused `getSRSbySRID` declaration from the PostGIS shim header

## Rationale

`asMFJSON` only needs the CRS name text for MF-JSON output. Looking it up through SQL keeps that behavior local to MobilityDB and avoids depending on a private PostGIS C helper symbol for SRID-to-SRS conversion.

## Validation

- `cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug`
- `cmake --build build --target pg_temporal -j32`
- `git diff --check`
- `git diff HEAD^..HEAD --check`
- `rg -n "getSRSbySRID" mobilitydb meos -g "*.[ch]"` returned no matches

Full local `cmake --build build --target MobilityDB-1.4 -j32` still fails before linking MobilityDB in the existing `pgtypes` target: `pgtypes/libpq/pqformat.h` calls `elog` without a visible declaration under this local PostgreSQL 18/GCC 16 setup. The touched `pg_temporal` target builds successfully.

### DIFF
--- a/mobilitydb/src/temporal/type_out.c
+++ b/mobilitydb/src/temporal/type_out.c
@@ -35,9 +35,13 @@
 
 /* PostgreSQL */
 #include <postgres.h>
+#include <catalog/pg_type.h>
+#include <executor/spi.h>
 #include <pgtypes.h>
+#include <utils/memutils.h>
 /* PostGIS */
 #include <liblwgeom_internal.h>
+#include "lwgeom_pg.h"
 /* MEOS */
 #include <meos.h>
 #include <meos_internal.h>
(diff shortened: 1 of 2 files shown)


mschoema:
Looks good, but I have 2 comments:

- This function should call `postgis_initialize_cache()`, otherwise it could try to read an uninitialized cache in some scenarios.
- Maybe we should split the function into 2, just like postgis does with `GetSRSCacheBySRID` and `getSRSbySRID` (splitting the cache handling and SPI calls essentially).

mobilitydb/src/temporal/type_out.c
```diff
@@ -114,6 +118,104 @@ Temporalarr_as_text(PG_FUNCTION_ARGS)
...
+typedef struct
+{
+  int32_t srid;
+  bool short_crs;
+  char *srs;
+} MFSrsCache;
+
+/**
+ * @brief Return the SRS name associated to a SRID
+ */
```
mschoema, line 130: I would add the following note, as its important to know in case other functions want to use this as well:

```
/**
 * @brief Return the SRS name associated to a SRID
 * @note The result is allocated in the function memory context
 */
```

mobilitydb/src/temporal/type_out.c
```diff
@@ -114,6 +118,104 @@ Temporalarr_as_text(PG_FUNCTION_ARGS)
...
+
+  srs = SPI_getvalue(SPI_tuptable->vals[0], SPI_tuptable->tupdesc, 1);
+  if (! srs)
+  {
+    SPI_finish();
+    return NULL;
+  }
+
+  oldcontext = MemoryContextSwitchTo(fcinfo->flinfo->fn_mcxt);
+  srscopy = pstrdup(srs);
```
mschoema, line 178: I would also add a comment here:

```
/* Copy the result out of the SPI context before SPI_finish() frees it */
```

mschoema (commented):
I think we should add a couple comments to make sure we don't mess up the allocations later if we want to work on these functions. Other than that, everything looks.

PR #1315, opened 2026-07-06
Use the native geometry clip for tIntersects and tDisjoint on temporal points

estebanzimanyi:
The temporal spatial relationships `tIntersects` and `tDisjoint` between a temporal geometry point and a geometry delegate the intersection to GEOS via `ST_Intersection`. This computes the common case — a temporal geometry point with linear interpolation over a non-curved geometry — with the native temporal-point geometry clip instead, avoiding the GEOS call.

- The `tpoint_geom_clip.c` module is added to the build and adapted to the current `MeosArray` and R-tree APIs. `temporal_rtree.c` is moved to the shared source list so that `rtree_search`, on which the clip relies, is available to both `libmeos` and the PostgreSQL extension.
- `tIntersects` and `tDisjoint` of a temporal geometry point with linear interpolation over a non-curved geometry are computed by `tpoint_linear_inter_geom`. Curved geometries and instantaneous values fall through to the existing GEOS path unchanged.

The local regression suite passes with no output differences against master.

### DIFF
--- a/meos/include/meos_internal_geo.h
+++ b/meos/include/meos_internal_geo.h
@@ -224,6 +224,11 @@ extern Temporal *tgeoseq_restrict_stbox(const TSequence *seq, const STBox *box, 
 extern TSequenceSet *tgeoseqset_restrict_geom(const TSequenceSet *ss, const GSERIALIZED *gs, bool atfunc);
 extern TSequenceSet *tgeoseqset_restrict_stbox(const TSequenceSet *ss, const STBox *box, bool border_inc, bool atfunc);
 
+/* Native temporal-point geometry-clip engine */
+
+extern Temporal *tpoint_linear_inter_geom(const Temporal *temp, const GSERIALIZED *gs, bool clip);
+extern Temporal *tpoint_linear_restrict_geom(const Temporal *temp, const GSERIALIZED *gs, bool atfunc);
+
 /*****************************************************************************/
 
 /* Traditional comparison functions */
--- a/meos/src/geo/CMakeLists.txt
+++ b/meos/src/geo/CMakeLists.txt
@@ -14,6 +14,7 @@ set(GEO_SOURCES
   tgeo_tile.c
   geo_poly_clip.c
   tpoint_datagen.c
+  tpoint_geom_clip.c
   tpoint_spatialfuncs.c
   tspatial.c
   tspatial_parser.c
--- a/meos/src/geo/tpoint_geom_clip.c
+++ b/meos/src/geo/tpoint_geom_clip.c
@@ -40,6 +40,7 @@
 /* PostgreSQL */
 #include "postgres.h"
 #include <utils/float.h>
+#include <utils/timestamp.h>
 /* PostGIS */
 #include "liblwgeom.h"
 #include "liblwgeom_internal.h"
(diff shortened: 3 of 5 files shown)


estebanzimanyi:
### Why this is scoped to `tIntersects`/`tDisjoint`, and the open question for `atGeometry`

This PR routes `tIntersects`/`tDisjoint` of a linear temporal geometry point over a non-curved geometry through the native `tpoint_geom_clip` engine, with a GEOS fallback, and is output-identical to the GEOS path on the regression corpus (0 diffs).

Extending the same treatment to the restriction functions was tried and left out, because — unlike the relationship predicates, which return a `tbool` — they emit the exact restricted geometry and the boundary-crossing timestamps, so the native analytic solver and GEOS `ST_Intersection` become distinguishable. Note that `minusGeometry` carries no geometry code of its own: it is `atGeometry` followed by a temporal complement (`temporal_restrict_tstzspanset` of the original value against the `atGeometry` result's time spanset). So the whole question reduces to `atGeometry`.

**The substantive one — boundary-crossing timestamp, floating-point precision.** `306_tnpoint_spatialfuncs`:

```
round(atGeometry(tnpoint '[Npoint(1,0.2)@2000-01-01, Npoint(1,0.4)@2000-01-02, Npoint(1,0.5)@2000-01-03]',
                 geometry 'SRID=5676;Polygon((50 50,50 100,100 100,100 50,50 50))'), 6)

GEOS   : { ... NPoint(1,0.481784)@2000-01-02 19:37:41.052095 }
native : { ... NPoint(1,0.481784)@2000-01-02 19:37:41.052165 }
```

The exit position is identical (`0.481784`); the exit time differs by ~70 µs. `round(...,6)` rounds the position but not the derived timestamp, so a sub-1e-7 difference in the computed crossing point surfaces in the timestamp. The geometry is straight (a Polygon) — no linearization is involved, this is purely two floating-point paths (analytic solve vs `ST_Intersection`), and neither value is "more correct". `minusGeometry` inherits exactly these crossing times.

**Two closable details.**

Container subtype — `056_tpoint_spatialfuncs`:

```
minusGeometry(tgeompoint '[Point(0 0)@2000-01-01, Point(1 1)@2000-01-02)', geometry 'Linestring(0 1,2 1)')

GEOS   : {[POINT(0 0)@... , POINT(1 1)@...)}   -- tgeompointSeqSet
native : [POINT(0 0)@... , POINT(1 1)@...)     -- tgeompointSeq
```

Same values, different subtype. This is the no-intersection shortcut in the minus path returning the sequence directly instead of routing through `temporal_restrict_tstzspanset` (which yields the sequence set the GEOS path produces). A normalization detail, not geometry.

Curve fallback — `070_tpoint_spatialrels`:

```
aTouches(tgeompoint '[Point(0 0)@2000-01-01, Point(3 3)@2000-01-04]', geometry 'CurvePolygon((1 1,2 2,3 1,2 0,1 1))')

GEOS   : f
native : ERROR:  Unsupported geometry type
```

`tgeo_restrict_geom` guards curved geometry with `lwgeom_has_arc` and falls back to GEOS, but `aTouches` reaches the clip through a path that bypasses that guard; the fallback must cover every entry point before the restriction graft is enabled.

**Summary.** The container and curve items are closable either way. The one policy call is `atGeometry`'s crossing-time: enabling the native restriction means accepting the native (equally valid) crossing values and regenerating the affected expected outputs. That is left out of this PR, which stays limited to the 0-delta `tIntersects`/`tDisjoint` surface.

PR #1318, opened 2026-07-06
Add circular strings to the temporal-point spatial relationship tests

estebanzimanyi:
The table-driven temporal spatial relationship tests cover point, line and polygon geometries, but not circular strings, so the native circular-arc path used by `tIntersects` and `tDisjoint` is exercised only by the individual literal cases.

This adds a `random_geom_circularstring` generator (a random walk of `2n+1` points assembled into a circular string) and a `tbl_geom_circularstring` test table, and runs `tIntersects` and `tDisjoint` over it in both argument orders.

The table is kept separate from `tbl_geom` on purpose: only `tIntersects` and `tDisjoint` compute circular arcs natively, so circular strings are exercised by those tests only, not by the spatial functions, spatial relationships and distance tests that also read `tbl_geom`.

The circular strings are added to the frozen `load_point.sql.xz` fixture; the pre-existing tables are unchanged.


estebanzimanyi:
Closing in favor of making circular strings a first-class member of the general geometry test tables (tbl_geom / tbl_geometry) rather than a separate table, so every table-driven test exercises them uniformly. That approach supersedes this one.

PR #1364, opened 2026-07-10
Strip patch word from GEOS version.

sebastic:
The Debian package failed to build with GEOS 3.15.0beta1:
```
-- GEOS_VERSION: 3.15.0beta1
-- GEOS_VERSION_MAJOR: 3beta1
-- GEOS_VERSION_MINOR: 15beta1
-- GEOS_VERSION_MICRO: 0beta1
-- POSTGIS_GEOS_VERSION: ERROR
CMake Error at CMakeLists.txt:258 (math):
  math cannot parse the expression: "3beta1 * 10000 + 15beta1 * 100 +
  0beta1": syntax error, unexpected exp_NUMBER, expecting end of file (7).
```


estebanzimanyi (approved):
LGTM. `geos-config --version` reports `3.15.0beta1` for the current GEOS pre-releases, and the old `([0-9]+)\.([0-9]+)\.([0-9]+)(dev)?` pattern only anchors an optional `dev`, so `REGEX REPLACE` rewrites the matched `3.15.0` and leaves `beta1` dangling — hence `GEOS_VERSION_MAJOR=3beta1` and the `math(EXPR ...)` failure. Extending the suffix to `(dev|(alpha|beta|rc)[0-9]*)?` consumes the pre-release tag so the three components come out clean, matching PostGIS's own FindGEOS idiom. It patches all three MODULE-path sites (framework, geos-config, header fallback); the CONFIG path in `CMakeLists.txt` already handles suffixes via its trailing `.*`. Build-time version detection only, no runtime impact. Thanks for the fix.

PR #1421, opened 2026-07-16
Add an optional strict argument to valueAtTimestamp

estebanzimanyi:
The `valueAtTimestamp` functions are thin wrappers over the MEOS
`<type>_value_at_timestamptz(temp, t, bool strict, result)` primitives. The SQL
surface exposes only the first two arguments and the wrappers bind `strict` to
the literal `true`, so a non-strict lookup — one that returns the value even when
the timestamp falls on an exclusive bound of the temporal value — is not
reachable from SQL.

This change exposes `strict` as an optional third argument defaulting to `TRUE`,
leaving existing two-argument calls unchanged. It follows the pattern of the
sibling `beforeTimestamp`/`afterTimestamp` restrictions (same
`*_timestamptz(temp, t, bool strict, ...)` shape) and of `segmentMinDuration`.

The `Temporal_value_at_timestamptz` and `Trgeometry_value_at_timestamptz`
wrappers read the third argument via `PG_GETARG_BOOL(2)`. The change spans every
temporal family (temporal, tgeo/tpoint, tcbuffer, tnpoint, trgeo, tjsonb, tpose,
th3index, tquadbin, tpcpoint, tpcpatch), the English and Spanish documentation,
and a regression test that exercises the non-strict path at an exclusive bound.

### DIFF
--- a/doc/es/temporal_circular_buffers.xml
+++ b/doc/es/temporal_circular_buffers.xml
@@ -537,7 +537,7 @@ SELECT radius(tcbuffer 'SRID=5676;{Cbuffer(Point(3 3), 0.3)@2001-01-01,
 			<listitem xml:id="tcbuffer_valueAtTimestamp">
 				<indexterm significance="normal"><primary><varname>valueAtTimestamp</varname></primary></indexterm>
 				<para>Devuelve el valor en una marca de tiempo</para>
-				<para><varname>valueAtTimestamp(tcbuffer,timestamptz) → cbuffer</varname></para>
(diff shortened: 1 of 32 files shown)


estebanzimanyi:
The `strict` flag is an internal MEOS mechanism; exposing it across every family's public `valueAtTimestamp` signature bloats the SQL surface for a niche behaviour, and the value-at-timestamp semantics are well-defined without it (NULL at an exclusive bound is correct). The binding-generation gap this targets belongs in the catalog default handling, not the SQL signature. Closing in favour of that approach.

PR #1528, opened 2026-07-28
Document and test the native curved-geometry path of the tcbuffer relationships

estebanzimanyi:
The tcbuffer relationship kernels build their geometry context with the arc-aware edge decomposition, `tcbuffer_geo_ctx_make` calling `geodist_geom_edges` with arcs allowed, and the root searches dispatch on `is_arc` through `tcbuffersegm_arc_within_roots` and `geodist_segm_arc_mindist`. Circular strings, compound curves, curve polygons and the multi types grouping them are therefore handled natively and arc-exactly, without linearizing anything through GEOS.

The comments still described curved input as unsupported and sent to the traversed-area fallback, which stopped being true when the kernels started consuming arc edges. Only a geometry that has no edge decomposition, that is, a TIN or a polyhedral surface, still falls back, which is what the existing `eTouches` case on a polyhedral surface in `212_tcbuffer_spatialrels` already exercises under the label of an unsupported geometry type.

Correct the comments accordingly across the circular buffer distance, relationship and temporal relationship kernels and the shared edge decomposition.

## Tests

Add curved input to the ever, always and temporal relationship tests: circular string, compound curve, curve polygon, multi curve and multi surface operands for the intersects, disjoint, contains, covers, touches and within-distance families. Among them are the two cases that show the decomposition is exact rather than approximate:

- angular gating, where a disc reaches the circle supporting an arc but not the arc itself, so it neither intersects it nor comes within a distance of it;
- an internally tangent disc, which the ring covers but does not contain;
- and a disc grazing a bare circular string, which touches it at both the external and the internal tangency, where against a curve polygon only the external one is a touch.

### DIFF
--- a/meos/src/cbuffer/tcbuffer_distance.c
+++ b/meos/src/cbuffer/tcbuffer_distance.c
@@ -548,7 +548,7 @@ nai_tcbuffer_geo(const Temporal *temp, const GSERIALIZED *gs)
     return tinstant_make_free(value, temp->temptype, t);
   }
 
-  /* Curved or otherwise unsupported geometry: fall back to the centreline */
+  /* A geometry that has no edge decomposition: fall back to the centreline */
   Temporal *tpoint = tcbuffer_to_tgeompoint(temp);
   TInstant *resultgeom = nai_tgeo_geo(tpoint, gs);
   Datum value;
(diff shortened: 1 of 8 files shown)


estebanzimanyi:
The comment corrections and the curved test cases are on master as of 7800b0d1c, which carries this branch's commit. Master holds zero of the superseded phrases across the circular buffer distance, relationship and spatiotemporal relationship kernels and the shared edge decomposition, and holds all 26 curved cases in 212 and all 24 in 214, queries and expected output alike.

What remains here is older text: the unsplit root search, its two call sites, and the previous ever within-distance comment. Merging it reverts them, so this closes as superseded.

PR #1582, opened 2026-07-30
Generate the temporal circular buffer temporal spatial relationships

estebanzimanyi:
Bring `214_tcbuffer_tempspatialrels` onto the inherited-behaviour generator.

Unlike the cast-delegation families (pose, cell indices, rigid geometry) whose temporal spatial relationships are SQL wrappers casting to `tgeometry`, a temporal circular buffer implements them as **native C functions** declared with `AS 'MODULE_PATHNAME','<Symbol>'`. This adds a native template (`tempspatialrels_native.sql.tmpl`) and renderer that emit the whole file: per predicate a banner and one `CREATE FUNCTION` per direction, backed by the C symbol `T<rel>_<left>_<right>` (where `geometry` tokenizes to `geo`). A family declares only its ordered directions and predicate set in the manifest.

- **Behaviour-neutral** — the thirty relationships keep identical signatures and backing C symbols (the signature set and the symbol set are both diff-empty against the previous file).
- **Corrects two hand-written slips** — the `circurlar` typo in the file brief, and the `tCovers` section whose banner reads `tContains`.

`generate.py --validate` self-regenerates the file byte for byte; every other family regenerates unchanged.


estebanzimanyi:
Superseded: this commit is contained in the merged base type Input/Output generator branch, so its content (the native tempspatialrels file, the tempspatialrels_native template, and the manifest axis) is already on master. Closing to avoid a duplicate.

PR #1610, opened 2026-07-31
Prune the circular buffer containment of a temporal circular buffer with the nearest approach

estebanzimanyi:
The ever and always contains and covers of a temporal circular buffer and a circular buffer reject a pair whose radius-aware bounding boxes are disjoint, or whose exact disk-to-disk nearest-approach distance is strictly positive, before the per-instant geometry containment predicate: a positive nearest approach means the disks are disjoint, so neither contains nor covers the other under either quantifier.

### DIFF
--- a/meos/src/cbuffer/tcbuffer_spatialrels.c
+++ b/meos/src/cbuffer/tcbuffer_spatialrels.c
@@ -693,6 +693,17 @@ int
 ea_contains_tcbuffer_cbuffer(const Temporal *temp, const Cbuffer *cb,
   bool ever)
 {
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Retracting: the ever-contains/covers use a DE-9IM pattern on the polygon approximation of the disc (cbuffer_to_geom), which diverges from the exact disc-disc nearest-approach the prune uses — the prune rejects pairs the polygon interiors still intersect, making the result inconsistent (exact for far pairs, polygon-approx for near). A correct fix needs a fully native exact disc-contains, not a prune on the approximate path.

PR #1643, opened 2026-07-31
End the interleaved set comparison regions before the hash tail

estebanzimanyi:
The h3indexset/quadbinset and pcpointset/pcpatchset `comparison_families` entries slice their region through the hash surface: the h3/quadbin entries end at the unnest section, carrying the hash opclass, and the pointcloud entries end at the Set-operations section, carrying the hash functions and opclass — in both cases because there is no `/*` divider for the standard end-anchor rule to land on.

This PR adds an `end_exact` flag to the comparison axis: the extracted (and, for a future non-reference family, bootstrapped) region ends exactly at the `end:` anchor instead of at the `/*` opening the section that anchor names. The four entries are re-anchored on their hash tail — the hash opclass for h3indexset/quadbinset, the first hash function for the pointcloud sets — with their `lit` blocks truncated to match, so the comparison regions of all four families end after the B-tree opclass and the hash surface can be owned by its own axis.

One irregularity stays encoded rather than moved: h3indexset/quadbinset interleave their `hash()` function between `cmp()` and the operators, so that interior function remains inside the comparison region's compact-form `lit` (a contiguous region cannot exclude its middle). Their hash tail proper — the opclass — sits outside the region.

Generator-only change: no `.in.sql` is touched and `generate.py --validate` stays byte-for-byte green (106 OK, 0 DIFF); a full emit rewrites nothing. The branch is based on the uniform set-selectivity branch, so that commit appears in the commit list.


estebanzimanyi:

PR #1644, opened 2026-07-31
Add the set hash surface to the inherited SQL generator

estebanzimanyi:
Adds a `hash_families` axis to the inherited SQL generator, bringing the hash-indexing blocks of all sixteen set types under `--validate` governance: `hash()` -> integer (`Set_hash`) and `hashExtended(, bigint)` -> bigint (`Set_hash_extended`) — the canonical bare names across the tree — plus the `<set>_hash_ops ... USING hash` opclasses.

The CREATE FUNCTIONs render from the same four-line skeleton the comparison surface uses (`templates/comparisons.sql.tmpl`) with a `_HASH_OPS` table supplying the signatures; the opclasses render from a five-line `_HASH_OPCLASS` skeleton; banners, dividers and blank lines stay verbatim `lit` blocks. A multi-type file (`001_set`, `050_geoset`) groups the functions op-outer / type-inner and packs consecutive opclasses, exactly like the comparison surface.

Anchoring reuses the comparison-axis conventions plus two additions: `begin_exact` (the pointcloud hash blocks follow the B-tree opclass with no divider) and `whole_file` (001_set's hash section is the file tail). Two lit-only irregularities are encoded, not normalized: h3indexset and quadbinset ship only a hash opclass in this region — their `hash()` function sits inside the comparison region's compact-form `lit`, they have no `hashExtended`, and their opclass uses a narrower spelling with no FUNCTION 2. Completing their surface to the canonical shape is a real SQL change for its own patch.

Every entry is `reference: true` (validate-only): no `.in.sql` is touched, `generate.py --validate` is byte-for-byte green (116 OK, 0 DIFF) and a full emit rewrites nothing. The branch is based on the comparison-region re-slice branch, whose commits appear in the commit list.


estebanzimanyi:

PR #1645, opened 2026-07-31
Add the set operation surface to the inherited SQL generator

estebanzimanyi:
Adds a `setop_families` axis to the inherited SQL generator, bringing the `setUnion` / `setMinus` / `setIntersection` blocks of all sixteen set types under `--validate` governance.

Each operation carries three argument directions — (value, set), (set, value), (set, set) — rendered from the shared four-line skeleton over the family's declared `(value, set)` type pairs, followed by the matching operators from a `_SETOP_OPERATOR` skeleton: `+` for setUnion and `*` for setIntersection with a COMMUTATOR clause, `-` for setMinus without one. Functions and operators group op-outer / pair-inner (a pair's three blocks packed, consecutive pairs separated by one blank line), which reproduces both the single-type files and the multi-type `002_set_ops` (six alpha pairs) and `050_geoset` (geometry/geography) byte-for-byte. Anchoring reuses the axis conventions (`begin`/`end` with the `/*` rule, `begin_exact` for the pointcloud seams, `whole_file` for the files whose set operations are the file tail; quadbinset alone ends with no trailing divider, kept as-is).

One genuine irregularity is surfaced and encoded rather than silently normalized. The RETURNS column of the direction table is the alpha-set spelling, which matches the C wrappers: every direction dispatches through `Setop_base_set` / `Setop_set_base` / `Setop_set_set` (`mobilitydb/src/temporal/set_ops.c`), all of which `PG_RETURN_SET_P`. A family with `value_returns: true` instead declares the VALUE type on `setMinus(value, set)` and `setIntersection(value, set)` / `(set, value)` — geoset, cbufferset, jsonbset, npointset, the pointcloud sets and poseset all do, 24 declarations in total — mislabeling the set datum those wrappers return (the same class of type-label defect as the geodetic constructor RETURNS fix). The override reproduces the committed declarations verbatim; correcting them is an API-surface change left to its own patch.

Every entry is `reference: true` (validate-only): no `.in.sql` is touched, `generate.py --validate` is byte-for-byte green (126 OK, 0 DIFF) and a full emit rewrites nothing. The branch is based on the hash-surface branch, whose commits appear in the commit list.


estebanzimanyi:

PR #1646, opened 2026-07-31
Add the set topological operator surface to the inherited SQL generator

estebanzimanyi:
Adds a `topop_families` axis to the inherited SQL generator, bringing the `contains` / `contained` / `overlaps` blocks of all sixteen set types under `--validate` governance.

The boolean CREATE FUNCTIONs render from the shared four-line skeleton — contains carries the (set, value) and (set, set) directions, contained (value, set) and (set, set), overlaps (set, set) only — followed by the `@>`, `<@` and `&&` operators from a `_TOPOP_OPERATOR` skeleton with their COMMUTATOR cross-references. Unlike the set operations, functions and operators pack across all pairs with no blank line between pairs, which reproduces the single-type files, the six-pair `002_set_ops` and the two-pair `050_geoset` byte-for-byte.

The operators' selectivity clause is where the families (and, inside `002_set_ops`, the individual pairs) genuinely disagree, so it is encoded as data rather than normalized: `selectivity:` is one token or a per-pair list — `span` (active `RESTRICT = span_sel, JOIN = span_joinsel`: the alpha int/bigint/float/tstz pairs, h3indexset, quadbinset), `none` (no clause: the alpha text/date pairs, jsonbset, the pointcloud sets), `span_comment` or `stbox_comment` (the clause present but commented out, including the missing comma before it: cbufferset/npointset/poseset and geoset). Uniformizing those clauses would be a real SQL change in the spirit of the set-comparison selectivity uniformization and belongs to its own patch.

Every entry is `reference: true` (validate-only): no `.in.sql` is touched, `generate.py --validate` is byte-for-byte green (136 OK, 0 DIFF) and a full emit rewrites nothing. The branch is based on the set-operation-surface branch, whose commits appear in the commit list.


estebanzimanyi:

PR #1831, opened 2026-08-13
Improve the generation of the mobilitydb extension SQL file

mschoema:
Assemble mobilitydb--VERSION.sql from a header carrying the licence, the version and the psql guard once, followed by each source under a marker naming it. The script previously had none of those and repeated the licence once per source, 152 times with every available family enabled.

Add the .in.sql sources as configure dependencies, so that editing one re-assembles the script instead of leaving make install to ship the previous SQL. The contributing guide no longer bills the manual cmake re-run as a gotcha.

Process each source in one step into ${CMAKE_BINARY_DIR}/mobilitydb/sql, the directory cmake already creates for it; the configure_file that followed process_file only mapped @extschema@ onto itself, and wrote to a MobilityDB/sql that differs from that directory only where the filesystem is case sensitive. Reduce process_file to the version thresholds above the PostgreSQL 16 minimum, and drop both the tail of sql/CMakeLists.txt that substituted into a copy of the script and wrote the result nowhere and all_sql_files, a duplicate of mobilitydb_sql.

Restore the header and the per-source licence strip of the datagen script, lost in #664 and #794, and share the assembly between the two. Seeding the file from the header is what keeps a re-configure from appending a second copy of every rule.

The assembled SQL is unchanged, verified against a clean worktree of the parent commit with every available family enabled.


estebanzimanyi:
The failing `coverage ratchet` check needs one line.

```
[NEW ] ungoverned and unlisted: mobilitydb/sql/header.in.sql
coverage: 66/194 governed, 127 still excepted
coverage ratchet FAILED
```

The ratchet fails any `.in.sql` under `mobilitydb/sql` that is neither named by a manifest entry nor listed as an exception, so that a new hand-written surface cannot land unnoticed. The assembled extension header is neither: it is a preamble for the assembled file rather than a surface any family axis emits, so it belongs in the exception list.

`tools/codegen/inherited/coverage_exceptions.txt` is sorted, so the line goes between the h3 and json blocks, after `286_th3index_metrics.in.sql`:

```
mobilitydb/sql/header.in.sql
```

With that line, a run on the branch head `b9a451337b` prints `coverage ratchet OK` at 67/195 governed and 127 excepted, and `--validate` and `--classes` print no diffs.

mschoema:
This is the first time I see this part of the code and this test, so I'm trying to understand it first. I will push the fix soon.

estebanzimanyi (commented):
The assembly holds up under a configure of both sides with every family enabled (`-DALL=ON`), comparing the generated `mobilitydb--1.4.0.sql` against one from a clean worktree of the parent commit `4df8aaf56d`:

- **The executable SQL is identical except one line.** Stripping comments and blank lines from both scripts leaves 56664 lines against 56665, and the single added line is the psql guard the header introduces:
  ```
  +\echo Use "CREATE EXTENSION mobilitydb" to load this file. \quit
  ```
  That guard is the standard idiom — 721 of the extension scripts installed under `/usr/share/postgresql/*/extension` carry the same line, so `CREATE EXTENSION` handles it.
- **Nothing is dropped.** The tree carries 193 numbered `.in.sql` sources and the assembled script carries 193 `-- Begin contents of` markers, one per source.
- **The licence collapses as intended**: 193 blocks down to 1, which is the whole 276 KB size difference.

The load order the sort relies on is sound: every one of those 193 sources carries a three-digit prefix, so a lexicographic sort of the paths is a numeric one, and all modules stage into the same `${CMAKE_BINARY_DIR}/mobilitydb/sql`, so the directory part is common to all of them. `header.in.sql` is the only source without a numeric prefix — it sorts after `999_oid_cache` — and it reaches the assembly as `HEADER` rather than through `SOURCES`, so it stays out of the sorted list.

Anchoring the licence regex on the invariant `This MobilityDB code is provided under The PostgreSQL License.` line is what makes the strip safe: the `/**** ... ****/` section separators inside the file bodies do not carry that line, and the `( \*[^\n]*\n)*` group only matches ` *`-prefixed lines, so a match cannot run past the licence into SQL.

One change required, the single failing check:

```
[NEW ] ungoverned and unlisted: mobilitydb/sql/header.in.sql
coverage ratchet FAILED
```

The ratchet fails any `.in.sql` under `mobilitydb/sql` that is neither named by a manifest entry nor listed as an exception, so that a new hand-written surface cannot land unnoticed. The extension header is a preamble for the assembled file rather than a surface any family axis emits, so it belongs in the exception list. `tools/codegen/inherited/coverage_exceptions.txt` is sorted, so the line goes after `286_th3index_metrics.in.sql`:

```
mobilitydb/sql/header.in.sql
```

That yields `coverage ratchet OK` at 67/195 governed, with `--validate` and `--classes` clean.

One observation, not a blocker: `SQL_VERSION_THRESHOLDS` is dormant — no `.in.sql` in the tree carries a `#if POSTGRESQL_VERSION_NUMBER` marker at this head, so nothing exercises `process_file`'s version gating. The comment above the list already states the rule for extending it, which is the part that matters given the supported range reaches PostgreSQL 19.

estebanzimanyi:
On the dormant `SQL_VERSION_THRESHOLDS`: it is worth turning into a derived value rather than a maintained list, because a threshold missing from the list is not inert. Appending this probe to a source on `18956ec5f7` and configuring with `-DALL=ON` on PostgreSQL 18.1:

```
#if POSTGRESQL_VERSION_NUMBER >= 190000
SELECT 'pg19-only-probe';
#endif //POSTGRESQL_VERSION_NUMBER >= 190000
```

puts the markers in `mobilitydb--1.4.0.sql` verbatim:

```
#if POSTGRESQL_VERSION_NUMBER >= 190000
SELECT 'pg19-only-probe';
#endif //POSTGRESQL_VERSION_NUMBER >= 190000
```

`#if` is not SQL, so `CREATE EXTENSION` fails on the assembled script. PostgreSQL 19 sits inside the supported range, so `190000` is the next threshold a source is likely to gate on, and it is the one the list omits.

Reading the thresholds out of the source that gates on them removes the coupling — the same way `FindGEOS.cmake` and `FindPOSTGIS.cmake` read a version out of a file they are handed:

```cmake
# The thresholds come from the source that gates on them, in the manner
# FindGEOS and FindPOSTGIS read a version out of a file they are handed. A list
# kept here instead has to be edited whenever a source starts gating on a
# version it does not already name, and a threshold missing from that list is
# not inert: the markers reach the assembled script verbatim, where `#if` is not
# SQL and CREATE EXTENSION fails on it.
macro(process_file f srcdir bindir)
  file(READ ${srcdir}/${f}.in.sql CURR_CONTENTS)

  string(REGEX MATCHALL "#if POSTGRESQL_VERSION_NUMBER [<>=]+ [0-9]+"
    SQL_VERSION_GATES "${CURR_CONTENTS}")
  set(SQL_VERSION_THRESHOLDS)
  foreach(g ${SQL_VERSION_GATES})
    string(REGEX REPLACE "^.* ([0-9]+)$" "\\1" v "${g}")
    list(APPEND SQL_VERSION_THRESHOLDS ${v})
  endforeach()
  list(REMOVE_DUPLICATES SQL_VERSION_THRESHOLDS)

  foreach(v ${SQL_VERSION_THRESHOLDS})
```

with the rest of the macro as it stands and the `set(SQL_VERSION_THRESHOLDS 170000 180000)` line above it dropped.

The same probe then lands commented out, on a threshold no list names:

```
/* -- if POSTGRESQL_VERSION_NUMBER >= 190000
SELECT 'pg19-only-probe';
-- endif POSTGRESQL_VERSION_NUMBER >= 190000 */
```

and a `>= 170000` probe stays active with its markers turned into `--` comments. A source that gates on nothing yields an empty threshold list, so the macro copies it through and the loop runs zero times, which is what every source in the tree does today.

estebanzimanyi:
One correction to the access point in my first comment: maintainer edits are enabled on this PR and I am admin on the base repo, so pushing to `sql-extension-assembly` is available to me. A patch in a comment does not need hand-application.

Whichever you prefer for the threshold change:

- I push it as one commit on top of your head, or
- you apply it yourself.

The branch stays yours either way — if a push of mine ever collides with one of yours, yours wins and I rebase.

estebanzimanyi (approved):

PR #2692, opened 2026-09-13
Give each temporal type its typed text input and output

estebanzimanyi:
State the wrapper every text input and output backs

The rule. A MEOS function names in its @csqlfn tag the PostgreSQL
wrappers that call it, and the MEOS-API catalog gives the function the
SQL signatures of those wrappers, which every binding projects. The text
inputs and outputs state this unevenly. The typed inputs of the sets,
spans and span sets name Set_in, Span_in and Spanset_in, and ts2cell_in
and tquadbin_in name Temporal_in, while the typed inputs of tbool, tint,
tbigint, tfloat, ttext, tjsonb and th3index name nothing, nor do those of
the temporal points and temporal geometries, whose wrappers are Tpoint_in
and Tgeo_in. The typed outputs of the sets, spans and span sets name their
output wrapper, floatspan_out with its tag line indented one space more
than its siblings; the typed outputs of the temporal types name none. And
no function names the wrappers asText calls: Temporal_as_text calls
temporal_out, Set_as_text set_out, Span_as_text span_out, Spanset_as_text
spanset_out, Stbox_as_text stbox_out, Temporalarr_as_text temparr_out and
Spatialarr_as_text spatialarr_wkt_out. tbox_out alone names both wrappers
it backs, Tbox_out and Tbox_as_text. The catalog therefore carries asText
over no temporal, set, span or span set type and over no stbox, so no
binding can generate it.

The change. Each typed temporal input names the wrapper MobilityDB binds
its type's input to: Temporal_in, Tpoint_in for the temporal points and
Tgeo_in for the temporal geometries. Each typed temporal output names
Temporal_out and Temporal_as_text; tnpoint_out names Temporal_out, its
asText being the spatiotemporal one. temporal_out, set_out, span_out,
spanset_out and stbox_out, and the typed set, span and span set outputs
whose asText MobilityDB binds to Set_as_text, Span_as_text or
Spanset_as_text, add that wrapper after their output wrapper, in the order
tbox_out writes them. temparr_out names Temporalarr_as_text, and
spatialarr_wkt_out names Spatialarr_as_text and Spatialarr_as_ewkt.
floatspan_out's tag line takes the indentation of its siblings. Only
doxygen comments change.

Measured on master 909a2ef060 with MEOS-API 41b04caaf6: the catalog holds
11044 SQL signatures over 6025 functions, and 11159 with this change; 43
functions gain 115 signatures and none loses one, each typed function
receiving only the signatures of its own type. check_csqlfn.py reports 0
auto-resolvable gaps and check_meos_only_files.py is clean. The
MobilityDuck generator emits the same 6434 registrations from either
catalog, since a binding reads a function under its first wrapper's SQL
name and every function tagged here is either named by a type input or
output wrapper or internal.

Let the value-surface generator reproduce the files it writes

The rule. A file a generator claims as its output holds what the
generator writes, and the generator's own check says so.
tools/codegen/temporal_basetype/generate.py claims tpcpoint.c and
tpcpatch.c, yet on master its output differs from them in 44 and 45
lines. The committed files include varatt.h plainly where the generator
writes a version guard, keep a blank line after the includes that the
generator drops, and carry by hand the wrapping a longer token forces:
"pgpointcloud point" replacing "JSONB" pushes comments and one call past
79 columns. The two files do not agree with each other either: tpcpoint.c
keeps that call at 81 columns where tpcpatch.c breaks it, and joins two
validators that tpcpatch.c keeps apart. --validate checks that the
reference reproduces itself and that no json token leaks into a target,
never that a target equals its committed file, so it passes on both.

The change. The generator writes the plain varatt.h include every MEOS
source uses and the blank line after the includes. Where a substituted
token pushes a comment or a code line past 79 columns, it re-fills the
comment and breaks the code line after its last fitting comma, continuing
two columns deeper; a line whose template carries no token stays as the
reference writes it. --validate also compares each target rendering with
its committed file. The reference tjsonb.c keeps within 79 columns as
well, wrapping the three comments that did not, and writes its
tstzspanset validators on one line as its three sibling constructors do.
tpcpoint.c and tpcpatch.c are regenerated: the file brief, one call and
three comments are wrapped, and the restriction briefs take back the
reference's words "a specific".

Measured on master 909a2ef060: the plain varatt.h include appears in 38
MEOS sources and the guarded form in none. With this change --validate
reports both targets reproduced, no line of the reference or of the
targets runs past 79 columns, and the regenerated files differ from
master in formatting and in those two words only.

Give each temporal cell and point cloud type its typed text output

The rule. A temporal type states its text form through a typed public
function that tests its argument, as tint_out does for a temporal integer
and tjsonb_out for a temporal JSONB, and that function names the wrappers
MobilityDB binds to it, which is how the catalog reaches asText. The
generic temporal_out stays internal, since no single test covers every
temporal type; a public function opens with the one validator its type
has. The temporal H3, QUADBIN and S2 cell indexes and the temporal
pgpointcloud points and patches have no such output, although MobilityDB
binds asText and the text output of each to Temporal_as_text and
Temporal_out. The pgpointcloud temporals lack a typed text input too,
where tjsonb_in and ts2cell_in name Temporal_in.

The change. th3index_out, tquadbin_out and ts2cell_out follow their
family's text inputs, clone tint_out with VALIDATE_TH3INDEX,
VALIDATE_TQUADBIN and VALIDATE_TS2CELL, and name Temporal_out and
Temporal_as_text. The pgpointcloud temporals take theirs from the
value-surface generator, which renders the text input and output section
of the tjsonb reference for them as it renders their constructors,
accessors and restrictions: tpcpoint_in and tpcpatch_in, their instant,
sequence and sequence set inputs, and tpcpoint_out and tpcpatch_out with
VALIDATE_TPCPOINT and VALIDATE_TPCPATCH. A banner closes the reference's
text section before its MF-JSON input, which is not a value bridge, and a
token carries the internal group. The headers declare each new function
beside its family's input.

The witness. meos/test/temporal_text_io_test.c reads a value of each of
the five types, writes it, reads the output back and writes it again, and
checks that the two outputs agree with no error left behind; each output
reports a null value with MEOS_ERR_INVALID_ARG and a temporal integer with
MEOS_ERR_INVALID_ARG_TYPE, and the two pgpointcloud inputs report a null
string with MEOS_ERR_INVALID_ARG. The program runs in the MEOS workflow
beside cellcover_test. The pgpointcloud outputs answer only with the
zero-decimal-digits fix beneath this change, since their hex output
refuses a count of zero without it.

Measured on master 909a2ef060 with MEOS-API 41b04caaf6: the catalog
gains the 13 new functions and 12 SQL signatures, 11159 to 11171, each
typed output carrying only its own type's temporal_out and asText
signatures and each input its own input signature; the MobilityDuck
generator at main 767186f43b emits the same 6434 registrations from
either catalog. The witness writes each of the five types and reads it
back with no error, reports each null value with MEOS_ERR_INVALID_ARG and
each temporal integer with MEOS_ERR_INVALID_ARG_TYPE. The release build
with every family compiles without warnings, and --validate reports both
pgpointcloud targets reproduced.


estebanzimanyi:
This branch holds the zero-decimal-digits fix of branch meos/zero-decimal-digits-is-a-valid-count as its first commit, since the pgpointcloud text outputs it adds answer only with that fix; its own change is the three commits that follow.

